PR #30, opened 2018-09-13
Renaming these functions in a more intuitive way.

AndyZe:
I think the logic is backward. We're adding a collision pair that is not allowed


AndyZe:
I can see how the logic makes sense, now. Closing it.

PR #32, opened 2018-09-20
Add unit test for clone method and fix mesh to mesh unit test names

Levi-Armstrong:


Levi-Armstrong:
@Jmeyer1292 You ok with me merging this?

PR #40, opened 2018-10-21
Refractor out ContactRequest type

Levi-Armstrong:
The type ContactRequest was left over from MoveIt and no longer make sense with the new implementation of contact managers.

### DIFF
--- a/tesseract_collision/test/collision_sphere_sphere_unit.cpp
+++ b/tesseract_collision/test/collision_sphere_sphere_unit.cpp
@@ -80,12 +80,8 @@ void runTest(tesseract::DiscreteContactManagerBase& checker)
   //////////////////////////////////////
   // Test when object is in collision
   //////////////////////////////////////
-  tesseract::ContactRequest req;
-  req.link_names.push_back("sphere_link");
-  req.link_names.push_back("sphere1_link");
-  req.contact_distance = 0.1;
-  req.type = tesseract::ContactRequestType::CLOSEST;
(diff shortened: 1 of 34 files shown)


tesseract_collision/test/collision_sphere_sphere_unit.cpp
```diff
@@ -80,12 +80,8 @@ void runTest(tesseract::DiscreteContactManagerBase& checker)
   //////////////////////////////////////
   // Test when object is in collision
   //////////////////////////////////////
-  tesseract::ContactRequest req;
-  req.link_names.push_back("sphere_link");
-  req.link_names.push_back("sphere1_link");
-  req.contact_distance = 0.1;
-  req.type = tesseract::ContactRequestType::CLOSEST;
-  checker.setContactRequest(req);
+  checker.setActiveCollisionObjects({"sphere_link", "sphere1_link"});
```
Jmeyer1292: So is `req.link_names` now named `setActiveCollisionObjects()`? All former uses now directly translate to the latter? I'm trying to keep the [hybrid examples](https://github.com/Jmeyer1292/hybrid_planning_experiments) from ROSCON updated and these changes just broke everything.

Levi-Armstrong:
That is correct. Sorry breaking things I will add it to my workspace to catch breaking changes.

Jmeyer1292:
Don’t hesitate to break. My rosinstall points toward the version I built with. Just be warned that I may ping you.

PR #42, opened 2018-10-31
Feature/add travis ci

Levi-Armstrong:


Levi-Armstrong:
@gavanderhoorn This is my first time working with industrial-ci/travis-ci do you mind taking a look at my .travis.yaml file?

.travis.yml
```diff
@@ -0,0 +1,29 @@
+# This config file for Travis CI
+sudo: required
+dist: trusty
+services:
+  - docker
+language: cpp
```
gavanderhoorn: not sure this is needed.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+cache: ccache
+compiler: gcc
```
gavanderhoorn: Do you have a specific reason for specifying this? The default should work afaik.

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+  email:
+    recipients:
+      - levi.armstrong@swri.org
+env:
+  global:
+    - ROS_DISTRO=kinetic
+    - UPSTREAM_WORKSPACE=file
+    - ROSINSTALL_FILENAME=dependencies.rosinstall
+    - TEST_BLACKLIST=bullet3_ros,fcl_ros,libccd_ros
+    - MAKEFLAGS="-j 4"
```
gavanderhoorn: Travis has sane defaults for this, and specifying something else might make this config brittle (ie: Travis may limit free accounts to a certain amount of resources and if we hardcode 4 jobs here, it could actually make things slower).

Did you set `4` here for a specific reason?

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+    - NOT_TEST_INSTALL=true
+  matrix:
+    - ROS_DISTRO=kinetic ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - ROS_DISTRO=kinetic ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=kinetic ROS_REPO=ros TEST=clang-format
```
gavanderhoorn: Why the mixed styles? IIRC, `ROS_REPO` and `ROS_REPOSITORY_PATH` do the same if you specify the default upstream repositories.

See [this config](https://github.com/ros-industrial/ros_canopen/blob/410984dc5ec8be0e41686a42618fdfa7c037e570/.travis.yml) for an example.

Levi-Armstrong: Ah, is one preferred over the other?

gavanderhoorn: I would say the "short" style is now the preferred one.

ipa-mdl: I have added the aliases because they are more readable (in config and on travis).

.travis.yml
ipa-mdl: the compiler is not used at all in `industrial_ci`.

.travis.yml
ipa-mdl: This enables the `cache: ccache` shorthand, but does not have any other benifits (for use with `industrial_ci`).

.travis.yml
ipa-mdl: `TEST_BLACKLIST` and `MAKEFLAGS` do not have any effect at all, because they are not used by `industrial_ci` and not passed to Docker.

.travis.yml
ipa-mdl: where did you find `TEST=clang-format`? The clang format PR is still pending and for now we do not support `TEST=`.

Levi-Armstrong: I believe it was from the moveit repository. I will look.

gavanderhoorn: `moveit` isn't using `industrial_ci`.

Levi-Armstrong: Ok, that make sense why clang never gets run.

tesseract_collision/CMakeLists.txt
```diff
@@ -18,14 +16,18 @@ find_package(Eigen3 REQUIRED)
 find_package(Boost COMPONENTS system thread REQUIRED)
 find_package(octomap REQUIRED)
 find_package(PkgConfig REQUIRED)
+find_package(ccd REQUIRED)
 
 pkg_check_modules(LIBFCL REQUIRED fcl)
 find_library(LIBFCL_LIBRARIES_FULL ${LIBFCL_LIBRARIES} ${LIBFCL_LIBRARY_DIRS})
 set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")
-list(APPEND LIBFCL_LIBRARIES ccd)
+list(APPEND LIBFCL_LIBRARIES ${CCD_LIBRARIES})
```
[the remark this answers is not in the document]
Levi-Armstrong: I will fix this. I was not aware that they were meant to be read-only.

.travis.yml
Levi-Armstrong: @ipa-mdl I have some external libraries that I would like not to run there tests. Is this possible with industrial_ci?

ipa-mdl: At the moment there is only a whitelist option: `PKGS_DOWNSTREAM`

tesseract_collision/CMakeLists.txt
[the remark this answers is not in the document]
Levi-Armstrong: Oh, I create the variable above `set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")` so I could add the `ccd` library.

.travis.yml
Levi-Armstrong: I removed it.

PR #46, opened 2018-11-12
Add EIGEN_MAKE_ALIGNED_OPERATOR_NEW macros

mpowelson:
This appears to not fix the alignment issues, but perhaps it is a step in that direction.

### DIFF
--- a/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
+++ b/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
@@ -92,6 +92,7 @@ inline btTransform convertEigenToBt(const Eigen::Isometry3d& t)
 class CollisionObjectWrapper : public btCollisionObject
 {
 public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
   CollisionObjectWrapper(const std::string& name,
                          const int& type_id,
                          const std::vector<shapes::ShapeConstPtr>& shapes,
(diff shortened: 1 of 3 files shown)


Levi-Armstrong:
I would compare to PR #45 to see if you found anything new.

tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
```diff
@@ -159,7 +159,7 @@ class CollisionObjectWrapper : public btCollisionObject
   {
     m_data.push_back(t);
   }
-
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
Levi-Armstrong: Does it matter where it goes in the class/struct?

tesseract_collision/include/tesseract_collision/fcl/fcl_utils.h
```diff
@@ -119,6 +119,7 @@ class CollisionObjectWrapper
     clone_cow->m_enabled = m_enabled;
     return clone_cow;
   }
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
Levi-Armstrong: Same as above?

tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
mpowelson: I don't think so. The documentation seems to imply that as long as it is public it should be fine. I can move it to the top if you prefer though.

PR #50, opened 2018-11-30
Being forgiving to SRDF joint <group> which contain fixed joints

arocchi:
At the moment, in `kdl_joint_kin`, loading a robots whose SRDF file has fixed joints in a given `<group>` will cause an assert to trigger.
On the other side, it looks like tools like the MoveIt! Setup Assistant don't punish users when adding fixed joints in groups, even though ti goes against the purpose of defining the group in the first place (i.e., defining a set of joints for planning).
Being this the case, we should be forgiving and just skip joints that are `fixed` upon loading, so that we don't take them into account when planning.
This small fix allows to do that.

What do you think @Levi-Armstrong ?

### DIFF
--- a/tesseract_ros/src/kdl/kdl_joint_kin.cpp
+++ b/tesseract_ros/src/kdl/kdl_joint_kin.cpp
@@ -305,11 +305,11 @@ bool KDLJointKin::init(urdf::ModelInterfaceConstSharedPtr model,
 
     if (jnt.getType() != KDL::Joint::None)
       joint_to_qnr_[jnt.getName()] = tree_element.second.q_nr;
+    else
+      continue;
 
     if (joint_it == joint_names.end())
       continue;
-
-    assert(jnt.getType() != KDL::Joint::None);
 
     // Add affected link names to list
     std::vector<std::string>::const_iterator link_it = std::find(link_list_.begin(), link_list_.end(), seg.getName());


gavanderhoorn:
I'm not Levi, but I would like to suggest to actually be stricter than the MSA and raise an error when `fixed` joints are present in a group.

Having such joints as part of a group makes little sense and users should not configure a group to contain any.

MoveIt used to contain all sorts of codepaths that ignored mistakes/small erorrs or incomplete configuration just so it could "still return a result". This led to it starting to use defaults or guesses for important parts of planning requests for instance. That may seem like a nice thing to do but in the end this never really led to satisfying results being returned, and in addition hid errors and problems from users.

Being explicit about errors and strict with pre-conditions avoids such problems.

I would therefore suggest to *not* (silently) ignore `fixed` joints (or any joints that cannot actually be planned for), but to actively check for them and print appropriate error messages when they are found to be part of a group.

If so desired, the entire group could be ignored, leading to it not being known to tesseract. That by itself should result in an error message printed later on, when other code tries to use the group.

arocchi:
@gavanderhoorn I am ok with error messages; retrocompatibility is an important issue though, and you can find many robots SRDF still being "dirty" out there, so some degree of grace would be useful.
The original code `assert`ed, I think we should either print a big red error message (but grace the user - in the end, the case for a fixed joint in a group is not so catastrophic), or throw a specific exception AND provide a flag to the constructor that disables it (but still prints an error message).

Tagging also @JaldertVicarious to have another voice in the discussion.

gavanderhoorn:
> retrocompatibility is an important issue though, and you can find many robots SRDF still being "dirty" out there, so some degree of grace would be useful.

Or we could say this is a good opportunity to force users to use properly configured move groups.

Tesseract is a new development. The people trying to / looking to use it are not regular users, but developers themselves. Those should be able to properly configure a move group.

Accepting user errors for the sake of bw-compatibility is a rather nasty practice and will not lead to those users improving their packages: it will just make them lazy (it works, so why bother).

arocchi:
Yes I could agree:don't forget that [MoveIt! is still actively pushing people to add joints](http://docs.ros.org/hydro/api/moveit_setup_assistant/html/doc/tutorial.html#step-4-add-planning-groups) in such a way, regardless whether they are fixed or not.

Moreover, the `<group>` element in SRDF [does not have a specific semantic](http://wiki.ros.org/srdf#Semantic_Robot_Description_Format_.28SRDF.29) defined for planning groups.
I would 100% agree with you that we should be non-forgiving if we were using a different tag altogether in SRDF e.g. `<planning_group>` or `<group type=planning>`.
Enforcing a particular meaning to an element that does not have one does not feel right IMHO.

JaldertVicarious:
You both make excellent points. 

I think it would be fine to enforce removing fixed joints from `<group>` if we are sure that there are no popular packages that do use fixed joints - otherwise users would need to maintain multiple SRDFs to be able to use Tesseract which seems like a hurdle. Since SRDF indeed does not define explicit semantics for `<group>` it is possible that some users do have a use for fixed joints inside a `<group>`.

If we do decide to make this a requirement, we should raise a specific error with concrete suggestion on how to address the issue, e.g.:

```The [NAME] group contains fixed joints [X,Y,Z] which can't be used in planning. Please edit the SRDF [SRDF PATH] to remove these joints from the group.```

Levi-Armstrong:
I agree with @gavanderhoorn and @JaldertVicarious that it is fine to enforce removing the fixed joints from the `<group>` and printing a error telling them what to fix. Allowing this could cause propagating issues through the code depending how the SRDF gets handled throughout.

arocchi:
Sounds good. I will update the PR by throwing an Exception and printing the specified message to screen

PR #55, opened 2018-12-14
Fix warnings in unit tests

Levi-Armstrong:
For some reason i was recieving several warnings from within gtest and the ignore warning was not solving the issue. In order to resolve the warnings was to using the `target_compiler_options` instead of `add_compiler_options`.


Levi-Armstrong:
@mpowelson Is this PR ok to merge?

mpowelson:
@Levi-Armstrong Looks good to me.

PR #56, opened 2018-12-17
Remove industrial ci and use travis ci directly

Levi-Armstrong:


gavanderhoorn:
What was the motivation here? Can you clarify?

Levi-Armstrong:
The main motivation was to have clang-format test and second I was interested in understanding how to use travis-ci directly for adding continuous integration non ROS repositories like the ros_qtc_plugin repository.

gavanderhoorn:
> The main motivation was to have clang-format test

Isn't this actually supported? The [docs](https://github.com/ros-industrial/industrial_ci/blame/master/doc/index.rst#L180) seem to suggest that `CLANG_FORMAT_CHECK` can be used for this (we're using this in `ur_modern_driver` ([here](https://github.com/ros-industrial/ur_modern_driver/blob/e2d81cf2b88237701931e02dbf05780d135c81c4/.travis.yml#L17)).

Levi-Armstrong:
That is good to know the last I checked was a few weeks ago and it did not appear to be supported. Is there a reason to use industrial-ci over using travis-ci directly? Using travis-ci directly seem simple enough.

gavanderhoorn:
> Is there a reason to use industrial-ci over using travis-ci directly? Using travis-ci directly seem simple enough.

I'm not sure I follow. In general or just for clang format check?

It's probably personal, but I like that setting up a file similar to [this one](https://github.com/ros-industrial/ur_modern_driver/blob/e2d81cf2b88237701931e02dbf05780d135c81c4/.travis.yml) gets me cached builds, clang checks, builds across multiple OS+ROS versions, automated setup of the ROS build environment and much more in very little time.

If you're not building ROS packages then there is obviously less value in that, but personally I'm rather happy that `industrial_ci` exists as it takes care of a lot of things for me I don't want to spend time on.

---

IIRC, it took you quite some time to get things to work before you submitted this PR (I remember quite a few fixup commits to your `removeIndustrialCI` branch). Again, personal, but I'd rather spend my time on doing other things.

Levi-Armstrong:
I will do some testing on the clang support. I need the ability to disable clang for some directories and briefly looking at the industrial_ci code it does not appear to be support but I will test to confirm. If it is not I will create an issue on industrial_ci and work on a PR.

If everything works on industrial_ci I will most likely just use this for the ros_qtc_plugin repository.

gavanderhoorn:
You're obviously completely free to use whatever you feel more comfortable with :)

I was just curious what the rationale was.

Levi-Armstrong:
I tested clang support with industrial_ci and everything works. I will need to work with industrial_ci to have the ability to blacklist/whitelist packages from clang-format test. Thank you for pointing this out.

It was a good exercise and now should be pretty easy to add ci support for ros_qtc_plugin repository.

PR #58, opened 2018-12-18
WIP: Add tesseract scene graph

Levi-Armstrong:
This is a implementation of a scene graph utilizing boost graph. Instead of creating class I decided to embed all necessary data within the graph itself and provide helper functions for operation on the graph. I have also provided an example in the test directory on how this can be used. There is also methods for getting chains from the graph and checking if the graph is acyclic. 

tesseract_scene_graph_1

tesseract_scene_graph_2


Levi-Armstrong:
@Jmeyer1292 and @gavanderhoorn  here is a start to scene graph using boost graph would be interested in your feedback if you have time.

Levi-Armstrong:
@arocchi I also wanted to get your input. I am working to further decouple things and the first part is to create a way to represent the objects and there connectivity that can handle trees, chains and graphs. This was another limitation of MoveIt that I wanted to develop solution to be able to handle cyclic connectivity.

Levi-Armstrong:
Integrated into master branch

PR #64, opened 2019-01-08
Add active link unit test

Levi-Armstrong:
The first commit introduces the test which results in a test failure. The second commit fixes the issue and the test now passes.


Levi-Armstrong:
The setActiveCollisionObjects indicates these objects can move and are not static in the environment. Some collision libraries handle static objects differently for computational reason because it never moves. The enable/disable of a collision object indicates if a object should be considered when performing contact tests.

Levi-Armstrong:
@mpowelson Friendly Ping!

mpowelson:
@Levi-Armstrong Sorry. Looks fine to me. It was just that one concern, but what you said makes sense.

Issue #36, opened 2018-10-04
KDLEnv can cause warnings in pluginlib


arocchi:
When using `KDLEnv` it often happens to get warning messages like the following
```
[ WARN] [1538691530.287577408]: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
```

PR #67, opened 2019-02-08
WIP: Reverse order of PluginLoader and ContactManager

mpowelson:
I'm not really sure why this fixes it, but this seems to fix the error below in the TrajOpt examples.  (https://github.com/ros-industrial-consortium/trajopt_ros/issues/56) 

```
terminate called after throwing an instance of 'class_loader::LibraryUnloadException'
  what():  Attempt to unload library that class_loader is unaware of.
```

### DIFF
--- a/tesseract_ros/include/tesseract_ros/kdl/kdl_env.h
+++ b/tesseract_ros/include/tesseract_ros/kdl/kdl_env.h
@@ -178,10 +178,11 @@ private:
       allowed_collision_matrix_; /**< The allowed collision matrix used during collision checking */
   IsContactAllowedFn
       is_contact_allowed_fn_; /**< The function used to determine if two objects are allowed in collision */
-  DiscreteContactManagerBasePluginLoaderPtr discrete_manager_loader_;     /**< The discrete contact manager loader */
-  ContinuousContactManagerBasePluginLoaderPtr continuous_manager_loader_; /**< The continuous contact manager loader */
+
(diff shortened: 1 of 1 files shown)


arocchi:
I remember that I did the exact opposite in here https://github.com/arocchi/tesseract/commit/9b3fee08c7e45ff05e73b966447c71f40422223e for another warning I was receiveing, but I can't remember (maybe that was not tracked in an issue) which waning exactly got solved with this - but for the specific error that you are mentioning, the reason of the error is explained by the deallocation order of those smart pointers.
When you delete a `KDLEnv` object, you will call destructors for all members, in reverse order to which they are declared. Apparently the loader needs himself to unload the objects it previously loaded?

mpowelson:
Ok. So #36 and #37  are the issue and PR you are talking about. I think you're right. It needs to stay the way it is now. 

It looks like in kdlenv we are calling createUniqueInstance and then immediately converting it to a shared_ptr. We should perhaps instead be either calling CreateInstance (Which returns a shared_ptr) or createUnmanagedInstance and adding unloadLibraryForClass() to the kdlenv destructor.

mpowelson:
@Levi-Armstrong So on the glass upright example, if I comment out the static KDLenvPtr defined at 58 and instead define it locally inside main, I do not get the error. I'm not really sure why.

Levi-Armstrong:
@mpowelson should this be closed?

mpowelson:
If the example tests pass now then I guess so. This was supposed to fix the error when it finished, but I don't think we ever got to the bottom of it.

PR #69, opened 2019-02-11
Add test to benchmark ACM isCollisionAllowed

mpowelson:
This is the benchmarking test I made to compare the new ACM method to the old one.


arocchi:
Would be very nice to report also numeric results of timing with the two implementations (`pair` vs string concatenation)!

mpowelson:
My testing was pretty informal, but on my PC it was taking ~2.7e-7s for both

mpowelson:
@Levi-Armstrong I added a regular assert that the average was less than 1e-4s. That way it won't run on debug builds.

Levi-Armstrong:
@mpowelson I believe assert only get included when building in debug and are not included in release.

mpowelson:
That was the idea. I was concerned that debug releases might be slower and artificially fail the test.

mpowelson:
Should be fixed.

PR #72, opened 2019-02-13
Add built-in Collision Shapes and Isolate the namespace in tesseract_collision

Levi-Armstrong:


tesseract_rviz/src/render_tools/env/robot_link.cpp
```diff
@@ -1328,6 +1330,501 @@ bool RobotLink::createEntityForGeometryElement(const shapes::Shape& geom,
...
+          }
+        }
+        break;
+      }
+      case tesseract::CollisionShapeType::CONVEX_MESH:
+      {
+        bool use_resource = false;
+        if (use_resource)
+        {
+          //      const urdf::Mesh& mesh = static_cast<const urdf::Mesh&>(geom);
```
mpowelson: Should be commented?

tesseract_collision/include/tesseract_collision/core/common.h
```diff
@@ -34,7 +34,9 @@ TESSERACT_COLLISION_IGNORE_WARNINGS_PUSH
 #include <Eigen/Geometry>
 #include <fstream>
 #include <iostream>
+#include <iomanip>
```
mpowelson: Is this needed?

mpowelson:
@Levi-Armstrong It looked good to me. I marked a few things for you to look at, but I doubt any of them will lead to required changes.

tesseract_collision/include/tesseract_collision/core/collision_shapes.h
```diff
@@ -0,0 +1,283 @@
...
+    OCTREE
+  };
+
+  class CollisionShape;
+  typedef std::shared_ptr<CollisionShape> CollisionShapePtr;
+  typedef std::shared_ptr<const CollisionShape> CollisionShapeConstPtr;
+  typedef std::vector<CollisionShapePtr> CollisionShapes;
+  typedef std::vector<CollisionShapeConstPtr> CollisionShapesConst;
+
+  class CollisionShape
```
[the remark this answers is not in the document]
Levi-Armstrong: Yep. I will add documentation.

tesseract_collision/src/create_convex_hull.cpp
```diff
@@ -23,13 +23,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include <tesseract_core/macros.h>
-TESSERACT_IGNORE_WARNINGS_PUSH
+#include <tesseract_collision/core/macros.h>
```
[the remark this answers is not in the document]
Levi-Armstrong: I created the macro to make it easier to add different compiler warnings without having to modify every header or source file. Now you just have to modify the macro and all files get updated. I have been basically wrapping all externally includes inside this macro so we don't get warnings we don't have control of.

My thinking was was to keep the macros separate, mainly because every file most likely will include the macro but may not need the utils. Other than that I don't have a strong opinion either way.

>   We really should figure out something for the documentation. I would have never known this tool was here.

I agree, The tesseract_collision package for the most part after this change should not change a lot going forward.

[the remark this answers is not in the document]
mpowelson: Oh, I meant the whole create_convex_hull executable

tesseract_collision/include/tesseract_collision/core/macros.h
```diff
@@ -0,0 +1,41 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_COLLISION_MACROS_H
+#define TESSERACT_COLLISION_MACROS_H
```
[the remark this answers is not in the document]
Levi-Armstrong: I created on specific to tesseract_collision just so it is completely standalone. 

>If so, is there a better alternative?
Supposedly when you add a included directory in CMake you can indicate it with SYSTEM and it should be ignored, but I could not get it to fully work. Now that I am thinking of it I might not have been adding it `catkin_INCLUDE_DIRS` but I cant remember. Its worth looking at again.

tesseract_rviz/src/render_tools/env/robot_link.cpp
Levi-Armstrong: Yea, it was originally commented out. It needs to be added to the TODO list.

tesseract_collision/include/tesseract_collision/core/common.h
Levi-Armstrong: I don't think so. I will remove it and see if things compile.

tesseract_msgs/msg/Mesh.msg
```diff
@@ -0,0 +1,9 @@
+# General definition of a mesh
+
+# the actual vertices that make up the mesh
+geometry_msgs/Point[] vertices
+
+# the indicies that make up the faces in the mesh
+# the first number indicates the number of vertices that make up the face,
+# followed by the indices of the vertices.
+uint32[] faces
```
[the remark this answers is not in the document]
Levi-Armstrong: This is how fcl does this and I do believe how bullet does also.

tesseract_collision/src/create_convex_hull.cpp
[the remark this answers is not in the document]
Levi-Armstrong: I was just looking at the line. The code that creates the convex hull is in the `core/common.h` but this is just a wrapper around it to allow for it to be used from the command line.

tesseract_monitoring/demos/demo_scene.cpp
```diff
@@ -55,20 +56,18 @@ void sendSphere()
...
-  sphere->radius = 0.15;
-
   sphere_pose.setIdentity();
   sphere_pose.translation() = Eigen::Vector3d(0.5, 0, 0.55);
 
   obj.name = "sphere_attached";
   obj.visual.shapes.push_back(sphere);
   obj.visual.shape_poses.push_back(sphere_pose);
-  obj.collision.shapes.push_back(sphere);
+  obj.collision.shapes.push_back(c_sphere);
```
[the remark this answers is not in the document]
Levi-Armstrong: Not at the moment.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_collision_object.h
```diff
@@ -0,0 +1,121 @@
+#ifndef TESSERACT_COLLISION_BULLET_COLLISION_OBJECT_H
+#define TESSERACT_COLLISION_BULLET_COLLISION_OBJECT_H
+
+#include <tesseract_collision/core/collision_object_base.h>
+#include <tesseract_core/macros.h>
```
mpowelson: @Levi-Armstrong I believe this needs to be changed to tesseract_collision/core/macros.h. Same for types

tesseract/tesseract_environment/include/tesseract_environment/core/types.h
```diff
@@ -0,0 +1,243 @@
...
+                                               std::hash<Key>,
+                                               std::equal_to<Key>,
+                                               Eigen::aligned_allocator<std::pair<const Key, Value>>>;
+
+using VectorIsometry3d = AlignedVector<Eigen::Isometry3d>;
+using VectorVector4d = AlignedVector<Eigen::Vector4d>;
+using VectorVector3d = std::vector<Eigen::Vector3d>;
+using TransformMap = AlignedMap<std::string, Eigen::Isometry3d>;
+
+typedef Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> TrajArray;
```
mpowelson: This also seems a bit odd that it is in environment.

Levi-Armstrong: I think it may be good to break these common type and the macro stuff and put it into a tesseract_common package that all use. What do you think?

mpowelson: That would make sense to me.

tesseract/tesseract_planning/include/tesseract_planning/core/utils.h
```diff
@@ -0,0 +1,105 @@
...
+#include <tesseract_planning/core/macros.h>
+TESSERACT_PLANNING_IGNORE_WARNINGS_PUSH
+#include <Eigen/Geometry>
+#include <memory>
+TESSERACT_PLANNING_IGNORE_WARNINGS_POP
+
+#include <tesseract_environment/core/environment.h>
+#include <tesseract_environment/core/types.h>
+#include <tesseract_kinematics/core/forward_kinematics.h>
+#include <tesseract_planning/waypoint_definitions.h>
```
mpowelson: Should be tesseract_planning/core/waypoint.h

mpowelson:
It also seems that anything that uses tesseract_rosutils also must include tesseract_collision or else you get an error 
`LinearMath/btConvexHullComputer.h: No such file or directory`

gavanderhoorn:
@Levi-Armstrong: was this PR supposed to get so big? Or are you committing to the wrong branch? :)

PR #78, opened 2019-03-18
Fix waypoint orientation return order and change joint waypoint to eigen

Levi-Armstrong:


Levi-Armstrong:
@mpowelson I made some changes to so the config only require base class and changed joint waypoint to use an eigen vector instead of std::vector.

tesseract_planning/src/trajopt/trajopt_freespace_planner.cpp
```diff
@@ -115,7 +117,7 @@ bool TrajOptFreespacePlanner::solve(PlannerResponse& response, TrajOptFreespaceP
       // Add initial joint position constraint
       std::shared_ptr<JointPosTermInfo> jv = std::shared_ptr<JointPosTermInfo>(new JointPosTermInfo);
       jv->coeffs = std::vector<double>(pci.kin->numJoints(), 1.0);
-      jv->targets = end_position->joint_positions_;
+      jv->targets = std::vector<double>(end_position->joint_positions_.data(), end_position->joint_positions_.data() + end_position->joint_positions_.size());
       jv->first_step = pci.basic_info.n_steps - 1;
       jv->last_step = pci.basic_info.n_steps - 1;
       jv->name = "target_joint_position";
```
mpowelson: I realize this is something I left out, but shouldn't there be a break here?

mpowelson: I looks like I'm missing it in array planner as well

Levi-Armstrong:
I will add it to this PR

Levi-Armstrong:
@mpowelson Added the remaining changes

Levi-Armstrong:
@mpowelson Clang passed

PR #79, opened 2019-03-26
Add ability to add tcp to array and freespace planners

Levi-Armstrong:


mpowelson:
@Levi-Armstrong Looks fine to me. I don't think we really use this in anything right now, so nothing should get broken.

PR #82, opened 2019-03-29
Trajopt planner unit tests

mpowelson:


mpowelson:
@Levi-Armstrong I don't really know what else I could add to these. We don't return the problem to check that the costs were actually added (perhaps we should). I'm just checking that the results have the correct number of rows and that it doesn't crash.

What would you think about adding a generateProblem method to these TrajOpt planners? They would take a config and return the problem. solve() would then start by calling that function. I don't know that it would really have much use other than testing.

Levi-Armstrong:
One option is to move the code that creates the problem into a protected method of the TrajOptArrayPlanner Class. Then in your test create a TrajOptArrayPlannerTest which inherits from TrajOptArrayPlanner and add a method to TrajOptArrayPlannerTest  which would return the problem  or runs the actual tests, because it has access to the protected methods. This would prevent you form having to add new things to the main class.

mpowelson:
> One option is to move the code that creates the problem into a protected method of the TrajOptArrayPlanner Class. Then in your test create a TrajOptArrayPlannerTest which inherits from TrajOptArrayPlanner and add a method to TrajOptArrayPlannerTest which would return the problem or runs the actual tests, because it has access to the protected methods. This would prevent you form having to add new things to the main class.

You have done this before. I like it.

Levi-Armstrong:
It looks like new test is failing and the clang formatting failed.

Levi-Armstrong:
@mpowelson Lets also switch this to the be against the master branch.

mpowelson:
@Levi-Armstrong See #100

PR #83, opened 2019-04-07
Fix trajopt planner to return correct status

Levi-Armstrong:
The tesseract planner interface expects all error codes to be less than one. Trajopt returns all positive status codes so this fixes the tesseract to return errors from TrajOpt with the correct tesseract planner interface error codes.


Levi-Armstrong:
@jrgnicho Could you please review?

PR #85, opened 2019-04-24
Add Enum for collision object types to CollisionGeometry.msg

mpowelson:


tesseract_msgs/msg/CollisionGeometry.msg
```diff
@@ -21,3 +21,9 @@ octomap_msgs/Octomap[] octomaps
 geometry_msgs/Pose[] octomap_poses
 std_msgs/Int32[] octomap_collision_object_types
 std_msgs/ColorRGBA[] octomap_colors # The color is optional
+
+# Types for collision objects
+byte USE_SHAPE_TYPE = 0          # Infer the type from the type specified in the shapes::Shape class
+byte CONVEX_HULL = 1             # Use the mesh in shapes::Shape but make it a convex hulls collision object. (if not convex it will be converted)
+byte MULTI_SPHERE = 2            # Use the mesh and represent it by multiple spheres collision object
+byte SDF = 3                     # Use the mesh and rpresent it by a signed distance fields collision object
```
gavanderhoorn: I'm not 100% sure, but I believe it's customary for enumerations (constants really) to be defined at the top of the message file.

Levi-Armstrong:
@mpowelson Can you switch this to be against master branch and the flags should go at the top of the file as @gavanderhoorn suggested.

mpowelson:
> @mpowelson Can you switch this to be against master branch and the flags should go at the top of the file as @gavanderhoorn suggested.

I don't think this PR is relevant in master. In master CollisionGeometry just has a geometry msg, and geometry msg has the types defined at the top already.

mpowelson:
I moved them to the top though if you still want to merge it into kinetic-devel. If not, that's fine. We can just close this.

Levi-Armstrong:
> I don't think this PR is relevant in master. In master CollisionGeometry just has a geometry msg, and geometry msg has the types defined at the top already.

Your correct closing.

PR #86, opened 2019-06-14
Robot comm monitor

AdmiralWall:
Added fanuc communication error monitoring (robot_communication_monitor.cpp node) under tesseract_monitoring.


mpowelson:
@AdmiralWall A couple of quick comments. 

- It looks like you may have forgotten an --amend. There are two commits with the same name.
- You probably didn't mean to commit those auto save files
- It looks like you may have built your changes on top a kinetic-devel based branch and then are merging into master. Master and kinetic-devel are pretty different right now. You'll want to make sure you submit the PR against the branch you are based on. For what it's worth, I believe we will need this functionality in master as well.

AdmiralWall:
Thanks for the feedback! I think I changed the branch to be based on master and got rid of those extra files.

gavanderhoorn:
@AdmiralWall: can you describe what you are monitoring with this node? Seeing as the name suggests it's Fanuc-specific, perhaps I can add something.

AdmiralWall:
Chris asked me to monitor and alert if Ethernet connection was cut out. 

Basically, the idea for this node is that when subscribing to joint_states, a count is created. Every 3 seconds, my node sets off a timer and checks the count. If we are successfully subscribing to joint_states (i.e., if we have a good connection), then the count will be incremented, and the error is flagged as false. However, if we are not getting the subscription to joint_states, then the count will not increment (count=0), and the communication error is true.

gavanderhoorn:
So I'm guessing the disconnection problems with `fanuc_driver_exp` (https://github.com/gavanderhoorn/fanuc_driver_exp/issues/17#issuecomment-487593665) haven't been completely disappeared then?

AdmiralWall:
@mpowelson  If I'm understanding Travis's build error correctly, this is failing due to Clang. I have followed instructions at https://ros-qtc-plugin.readthedocs.io/en/latest/_source/Setup-Qt-Creator-for-ROS.html and https://blog.qt.io/blog/2019/04/17/clangformat-plugin-qt-creator-4-9/ with no apparent change.

mpowelson:
> @mpowelson If I'm understanding Travis's b uild error correctly, this is failing due to Clang. I have followed instructions at https://ros-qtc-plugin.readthedocs.io/en/latest/_source/Setup-Qt-Creator-for-ROS.html and https://blog.qt.io/blog/2019/04/17/clangformat-plugin-qt-creator-4-9/ with no apparent change.

@AdmiralWall So you ran clang using tools->Beautifier->clangFormat but it formatted it in such a way that it fails CI? A couple of things

- If you go to Travis, click on the job, and expand the log you can see what exactly Clang didn't like. It looks like in your case it's the spacing on some comments. You can just manually correct the issues if you can't get Clang running (not recommended)

- Tesseract also has its own clang file in the root of the repo. I would imagine it is the same as the one on the wiki (standard ROS style), but it is possible they are different. Just copy the text of the one in the repo into the QT settings like you did before.

- I have had issues with descrepencies between clang versions on my local machine and the one in CI. I think CI uses 6 and the Xenial debian is 5 (or something like that). I don't think I've ever gotten that particular error though. It's always an issue with line spacing in classes.

- If you set up a Travis account, you can turn CI on in your own fork. That way Travis will run when you push commits and you don't have to wait until you do the PR to find out that it fails. I know that can be frustrating.

mpowelson:
Also, your git commit history still seems to be messed up somehow. You might consider just creating a new branch from the current master (or should she use Levi-Armstrong/Contact_shapes? @Levi-Armstrong ), and then cherry picking over your commits. We'll probably want to squash all of your commits as well.

I know I plugged this before, but you might consider a git GUI if things get too tangled. [This one](https://www.gitkraken.com/) is free for open source use, but there are others.

AdmiralWall:
Thanks, I appreciate the feedback. I'll look into the manual clang format, as well as Tesseract's specific file later today. And I did try to make a new branch previously, but I'll do it again and see if I missed something. Thanks!

AdmiralWall:
I created a new, clean robot_comm_monitor_2 branch and initiated a pull request after checking it through Travis. I am closing this pull request now in deference to the new one.

PR #89, opened 2019-07-08
Robot comm monitor 2

AdmiralWall:
Created new branch to clean up previous commits/pull requests with Clang errors for robot_comm_monitor_node.cpp. This version runs and builds and passes through Travis. 

Added fanuc communication error monitoring (robot_communication_monitor.cpp node) under tesseract_monitoring.

### DIFF
--- a/tesseract_monitoring/src/robot_comm_monitor_node.cpp
+++ b/tesseract_monitoring/src/robot_comm_monitor_node.cpp
@@ -1,0 +1,99 @@
+/*
+ * robot_communication_monitor.cpp
+ *
+ * Checks for fanuc communication errors
+ *
+ * Created on June 14, 2019
+ *    Author: rwall
+ *
+ *
+ */
+
+#include "boost/bind.hpp"
+#include "ros/ros.h"
+#include "std_msgs/Bool.h"
+#include "std_msgs/String.h"
+#include <sensor_msgs/JointState.h>
+#include <tesseract_monitoring/current_state_monitor.h>
+
+// Test run commands:
+// rosrun tesseract_monitoring robot_communication_monitor_node
+// rostopic pub -r 1 /joint_states sensor_msgs/JointState
+// rostopic echo /fanuc_com_error
+
+class FanucCommError
+{
+public:
+  FanucCommError(ros::NodeHandle& nh)
+  {
+    fanuc_comm_error.data = false;
+    chatter_pub = nh.advertise<std_msgs::Bool>("fanuc_com_error", 20);
+
+    chatter_sub = nh.subscribe("joint_states", 20, &FanucCommError::subscriberCallback, this);
+  }
+  void startTimer()
+  {
+    timer = FanucCommError::nh.createTimer(ros::Duration(error_timer), &FanucCommError::timerCallback, this);
+  }
+
+  void subscriberCallback(const sensor_msgs::JointStateConstPtr& joint_state)
+  {
+    count = count + 1;  // increase count of joint states subscription
+
+    // ROS_INFO("\n subscriber called. Count is %d \n", count); //debugging
+    // statement
+  }
+
+  void timerCallback(const ros::TimerEvent&)
+  {
+    // ROS_INFO("\n Timer called. Count is %d",count); //debugging statement
+
+    if (count < 1)
+    {
+      fanuc_comm_error.data = true;
+    }
+
+    else
+    {
+      fanuc_comm_error.data = false;
+    }
+
+    count = 0;                              // re-initialize error check
+    chatter_pub.publish(fanuc_comm_error);  // publish error status
+
+    // ROS_INFO("FANUC COM ERROR= %s \n", fanuc_comm_error.data ? "true" :
+    // "false");//debugging statement
+  }
+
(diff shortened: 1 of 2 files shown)


tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+// rostopic pub -r 1 /joint_states sensor_msgs/JointState
+// rostopic echo /fanuc_com_error
+
+class FanucCommError
+{
+public:
+  FanucCommError(ros::NodeHandle& nh)
+  {
+    fanuc_comm_error.data = false;
+    chatter_pub = nh.advertise<std_msgs::Bool>("fanuc_com_error", 20);
```
mpowelson: I would make this pull the topic name from a rosparam (or at least rename it). Also a lot of times hardcoded defaults like this will be defined at the top of the file so they are easy to find. e.g. at the top: `const std::string COM_ERROR_PUB_TOPIC_DEFAULT` or similar.

Then here `nh_.param<std::string>("error_topic", topic_name, COM_ERROR_PUB_TOPIC_DEFAULT)`

tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "std_msgs/String.h"
+#include <sensor_msgs/JointState.h>
+#include <tesseract_monitoring/current_state_monitor.h>
+
+// Test run commands:
+// rosrun tesseract_monitoring robot_communication_monitor_node
+// rostopic pub -r 1 /joint_states sensor_msgs/JointState
+// rostopic echo /fanuc_com_error
+
+class FanucCommError
```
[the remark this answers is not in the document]
mpowelson: I'd also throw your test commands into some doxygen comments since there is no header here.

For example from a Tesseract header
```
/** @class CurrentStateMonitor
    @brief Monitors the joint_states topic and tf to maintain the current state of the robot. */
```

tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+  {
+    count = count + 1;  // increase count of joint states subscription
+
+    // ROS_INFO("\n subscriber called. Count is %d \n", count); //debugging
+    // statement
+  }
+
+  void timerCallback(const ros::TimerEvent&)
+  {
+    // ROS_INFO("\n Timer called. Count is %d",count); //debugging statement
```
mpowelson: I'd say probably just get rid of these.

tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    count = 0;                              // re-initialize error check
+    chatter_pub.publish(fanuc_comm_error);  // publish error status
+
+    // ROS_INFO("FANUC COM ERROR= %s \n", fanuc_comm_error.data ? "true" :
+    // "false");//debugging statement
+  }
+
+private:
+  ros::Timer timer;
```
mpowelson: I know it's a pain, but ROS style is to have trailing underscores on class member variables like `ros::Timer timer_`. See [HERE](http://wiki.ros.org/CppStyleGuide#Classes_.2BAC8_Types)

tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
...
+  FanucCommError FCE(nh);
+
+  ros::Rate loop_rate(10);  // run at 10 Hz
+
+  while (ros::ok())
+  {
+    FCE.startTimer();
+    ros::spin();
+
+    loop_rate.sleep();
```
mpowelson: You don't actually need this. ros::spin() is blocking so you'll never get here. If you really needed to run in a loop you could call spinOnce(), but this is fine here. It'll just keep your node alive until shutdown.

tesseract_monitoring/src/robot_comm_monitor_node.cpp
```diff
@@ -0,0 +1,97 @@
+/*
+ * robot_communication_monitor.cpp
+ *
+ * Checks for fanuc communication errors
```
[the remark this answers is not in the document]
gavanderhoorn: There does not seem to be anything Fanuc specific here, no.

The node subscribes to `joint_states` and then keeps resetting a timeout upon reception of msgs. No msgs -> publish a `Bool` on a specific topic.

This would work for any ROS driver that publishes `JointState`s.

tesseract_monitoring/src/robot_comm_monitor_node.cpp
gavanderhoorn: I know this is unsollicited advice, so take it or leave it, but: don't configure topic names through parameters.

Hard-code some sensible name and then, if a consumer requires it, `remap` it to something else.

AdmiralWall:
This feature is being integrated into another in-progress branch as part of current_state_monitor

PR #107, opened 2019-07-29
Enable FCL tests since PR #338 was merged

Levi-Armstrong:
This enables FCL tests, but one test still fails. I have confirmed in blender that Bullet produces the correct results and FCL does not. Need to create an issue on FCL.

### DIFF
--- a/tesseract/tesseract_collision/test/collision_sphere_sphere_unit.cpp
+++ b/tesseract/tesseract_collision/test/collision_sphere_sphere_unit.cpp
@@ -194,17 +194,21 @@ void runConvexTest(DiscreteContactManager& checker)
   ContactResultVector result_vector;
   flattenResults(std::move(result), result_vector);
 
+  // Bullet: -0.270548 {0.232874,0,-0.025368} {-0.032874,0,0.025368}
+  // FCL:    -0.270548 {0.232874,0,-0.025368} {-0.032874,0,0.025368}
   EXPECT_TRUE(!result_vector.empty());
-  EXPECT_NEAR(result_vector[0].distance, -0.27552, 0.03);
+  EXPECT_NEAR(result_vector[0].distance, -0.270548, 0.001);
 
   std::vector<int> idx = { 0, 1, 1 };
   if (result_vector[0].link_names[0] != "sphere_link")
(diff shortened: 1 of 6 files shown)


tesseract/tesseract_collision/test/collision_sphere_sphere_unit.cpp
```diff
@@ -244,67 +253,43 @@ void runConvexTest(DiscreteContactManager& checker)
...
-  EXPECT_GT((idx[2] * result_vector[0].normal).dot(Eigen::Vector3d(0, 1, 0)), 0.0);
+  EXPECT_NEAR(std::abs(result_vector[0].nearest_points[idx[0]][0]), 0.0425563, 0.001);
+  EXPECT_NEAR(result_vector[0].nearest_points[idx[0]][1], 0.2308753, 0.001);
+  EXPECT_NEAR(std::abs(result_vector[0].nearest_points[idx[0]][2]), 0.0263040, 0.001);
+  EXPECT_NEAR(std::abs(result_vector[0].nearest_points[idx[1]][0]), 0.0425563, 0.001);
+  EXPECT_NEAR(result_vector[0].nearest_points[idx[1]][1], -0.0308753, 0.001);
+  EXPECT_NEAR(std::abs(result_vector[0].nearest_points[idx[1]][2]), 0.0263040, 0.001);
+  EXPECT_NEAR(std::abs(idx[2] * result_vector[0].normal[0]), 0.3037316, 0.001);
+  EXPECT_NEAR(idx[2] * result_vector[0].normal[1], 0.9340783, 0.001);
+  EXPECT_NEAR(std::abs(idx[2] * result_vector[0].normal[2]), 0.1877358, 0.001);
```
mpowelson: My only comment is that these seem like magic values. I assume these are correct.

Levi-Armstrong: Yea I agree.

Levi-Armstrong: Right now on some of the tests, if both bullet and fcl produces the same results I add the values into the test. In this case it is symmetric which is why I added the std::abs.

Issue #109, opened 2019-08-19
Return vector of succeeded/failed waypoints in PlannerResponse


mpowelson:
This is important for descartes especially where we want to remove failed vertices and try again.

mpowelson:
It looks like this was addressed in #113 but didn't get closed

PR #113, opened 2019-08-21
Add succeeded waypoints and failed waypoints to PlannerResponse

mpowelson:
Addresses issue #109

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
@@ -29,6 +29,7 @@
 #include <tesseract/tesseract.h>
 #include <tesseract_common/status_code.h>
 #include <tesseract_common/types.h>
+#include <tesseract_motion_planners/core/waypoint.h>
 
 namespace tesseract_motion_planners
 {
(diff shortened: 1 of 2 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
```diff
@@ -45,6 +46,8 @@ struct PlannerResponse
 {
   tesseract_common::JointTrajectory joint_trajectory; /**< @brief The joint trajectory */
   tesseract_common::StatusCode status;                /**< @brief The status information */
+  std::vector<Waypoint::Ptr> succeeded_waypoints;     /**< @brief Waypoints for which the planner succeeded */
+  std::vector<Waypoint::Ptr> failed_waypoints;        /**< @brief Waypoints for which the planner failed */
```
Levi-Armstrong: Should these be ConstPtr?

mpowelson: I don't know. I could imagine a case where you might want to modify the returned waypoints and it would be convenient for them not to be. Perhaps you want to adjust the coefficient values between Descartes and TrajOpt. Or maybe Descartes fails for one waypoint and you want to replan for the succeeded waypoints but adjust some of them to smooth something out.

I'll defer to your judgement.

Levi-Armstrong: We should a comment that these point to the waypoints provided in the input. So changing them here changes the source. Another alternative is return the index instead of the waypoint itself. 

Also should we update the other planners to fill out the failed and succeed waypoints as part of this PR?

mpowelson: Probably. I think I already did descartes, but we could at least return the failed ones from the collision check in trajopt.

Issue #114, opened 2019-08-21
Collision geometry flattened when loading multibody .DAE from URDF


schornakj:
I have a URDF that contains a .DAE file with multiple geometry elements. When I open the file in a text editor it has a structure like this (trimmed down significantly, it's quite bulky):
```
<?xml version="1.0" encoding="utf-8"?>
<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
  <asset>
...
  </asset>
  <library_geometries>
    <geometry id="body1" name="body1">
      <mesh>
        ...
      </mesh>
    </geometry>
    <geometry id="body2" name="body2">
      <mesh>
        ...
      </mesh>
    </geometry>
  </library_geometries>
...
</COLLADA>
```

I would like this file to be parsed such that every geometry element is loaded as a separate collision body. Based on conversations with @Levi-Armstrong this should be possible. However, I noticed that during collision checking Tesseract was treating the geometry as if it were a convex hull, and when I stepped into the URDF parser in my debugger I saw that the link associated with this object was assigned just one mesh that contained the vertices and triangles for all the bodies in the .DAE.

I think the culprit is line 208 in `tesseract_scene_graph::convert`:

https://github.com/ros-industrial-consortium/tesseract/blob/6598b801a4c3ac8d113f4aa716bda2bd023af73a/tesseract/tesseract_scene_graph/include/tesseract_scene_graph/parser/urdf_parser.h#L198-L213

This passes True to the `flatten` parameter of [`tesseract_scene_graph::createMeshFromPath`](https://github.com/ros-industrial-consortium/tesseract/blob/6598b801a4c3ac8d113f4aa716bda2bd023af73a/tesseract/tesseract_scene_graph/include/tesseract_scene_graph/parser/mesh_parser.h#L159-L162), which ultimately causes `Assimp::Importer` to be called with the `aiProcess_OptimizeMeshes` flag, which flattens all the bodies in the .DAE into a single mesh.

When I switched it to False the URDF parser assigned the link the correct number of collision meshes, and my collision checker works in the expected way (i.e. only reporting collisions when actually intersecting a mesh, instead of when intersecting the link's convex hull). The documentation for that function says that the `flatten` parameter shouldn't be used for collision meshes, so I think `urdf_parser.h` may need to be fixed.

Levi-Armstrong:
Addressed by #115

PR #115, opened 2019-08-21
Avoid flattening collision meshes when parsing URDFs

schornakj:
Fix for #114

Does changing this flag break anything else?


Levi-Armstrong:
Could you write a test that catches this in the future?

PR #118, opened 2019-08-29
Planner Verbosity

marip8:
This PR adds functionality for printing collision information about a trajectory to the console. It also adds an additional optional boolean argument to the `solve` method of the motion planner base class about whether or not to print important motion planning information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,11 @@ matrix:
         submodules: false
     - env: ROS_DISTRO=kinetic
     - env: ROS_DISTRO=kinetic ROS_REPO=ros-shadow-fixed
-    - env: ROS_DISTRO=melodic ROS_PARALLEL_JOBS=-j4  # Travis was running out of resources and hanging for melodic builds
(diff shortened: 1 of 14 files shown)


Levi-Armstrong:
The PR looks good except it looks like the wrong version of clang was ran. Below is the command line command to run for tesseract.

```
find . -type d \( -name bullet3_ros -o -name fcl_ros -o -name libccd_ros -o -name gtest_ros -o -name tesseract_ext \) -prune -o -type f -regex '.*\.\(cpp\|hpp\|cc\|cxx\|h\|hxx\)' -exec clang-format-8 -style=file -i {} \;
```

Levi-Armstrong:
@marip8 If you delete the clang format commit, I will create a PR against yours using clang-8.

marip8:
@Levi-Armstrong I just removed the clang formatting commit as requested

Levi-Armstrong:
@marip8 I just pushed clang format. If CI passes I will merge.

PR #123, opened 2019-09-12
Modifying Flags in GetEnvironmentInformation.srv

DavidMerzJr:
The GetEnvironmentInformation service uses the bitwise AND operator to separate user requests and send only certain information.  However, there are two things that seem incorrect.
- the 'flag' for ALLOWED_COLLISION_MATRIX was 254.  254 converts to 1111 1110.  Requesting this matrix was identical to requesting everything from LINK_LIST through JOINT_TRANSFORM.  Therefore, all of those would return if the ACM was requested, and the ACM would return if any of those were requested.
- the 'flag' for COMMAND_HISTORY was 0.  0 converts to 0000 0000.  Since bitwise AND between any number and 0 is 0 (which C++ translates as false), COMMAND_HISTORY was never returned.

In order to fix this:
- Changed from 8-bit ints to 16-bit ints (needed the room, since there are 10 items to 'flag')
- Incremented all 'flag' numbers (0->1, 1->2, 2->4, etc.)
- Added a comment to the srv file explaining how these values are used.

### DIFF
--- a/tesseract_ros/tesseract_monitoring/src/environment_monitor.cpp
+++ b/tesseract_ros/tesseract_monitoring/src/environment_monitor.cpp
@@ -859,6 +859,29 @@ bool EnvironmentMonitor::getEnvironmentInformationCallback(tesseract_msgs::GetEn
     }
   }
 
+  if (req.flags & tesseract_msgs::GetEnvironmentInformationRequest::JOINT_TRANSFORMS)
+  {
+    tesseract_environment::EnvState::ConstPtr current_state = tesseract_->getEnvironmentConst()->getCurrentState();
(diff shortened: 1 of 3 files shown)


DavidMerzJr:
@Levi-Armstrong 
The CI box says "build errored", but it looks like the build just timed out.

Levi-Armstrong:
I testing out one thing now to see if it fixes the issue with travic ci melodic builds.

Levi-Armstrong:
@DavidMerzJr Can you rebase on the latest. I changed some travis ci parameters that seem to help the melodic builds?

Levi-Armstrong:
@mpowelson You were correct some flags were not being handled. I have added to this PR to handle the remaining flags. Can you please review?

PR #129, opened 2019-09-20
Make tesseract environment thread safe

Levi-Armstrong:
@schornakj I was able to track down the multi threaded issue thanks to the valgrind tool helgrind which quickly pointed out the issue. Could you test this?

### DIFF
--- a/tesseract/tesseract_environment/src/kdl/kdl_state_solver.cpp
+++ b/tesseract/tesseract_environment/src/kdl/kdl_state_solver.cpp
@@ -36,6 +36,14 @@ namespace tesseract_environment
 using Eigen::MatrixXd;
 using Eigen::VectorXd;
 
+KDLStateSolver::KDLStateSolver(const KDLStateSolver &solver)
+  : scene_graph_(solver.scene_graph_)
+  , current_state_(std::make_shared<EnvState>(*(solver.current_state_)))
+  , kdl_tree_(std::make_shared<KDL::Tree>(*(solver.kdl_tree_)))
+  , joint_to_qnr_(solver.joint_to_qnr_)
+  , kdl_jnt_array_(solver.kdl_jnt_array_)
+{}
+
 bool KDLStateSolver::init(tesseract_scene_graph::SceneGraph::ConstPtr scene_graph)
 {
   scene_graph_ = std::move(scene_graph);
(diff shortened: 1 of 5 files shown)


Levi-Armstrong:
> do we need to do the kdl_state_solver as well?

I am adding to a clone method to the state solver and exposing a function to a clone of the state solver. We will need to update trajopt to store a cloned version of the state solver and use it to avoid the overhead of the mutex locks in the environment.

schornakj:
I get correct results now, both in my little standalone test case and in my bigger application, so I'd say that this is good to go. Thank you very much for looking into this, @Levi-Armstrong !

Levi-Armstrong:
@mpowelson I just added one commit that add the ability to clone the state solver and expose a method in the environment to get a cloned state solver. Can you take a look?

tesseract/tesseract_environment/src/kdl/kdl_state_solver.cpp
```diff
@@ -141,6 +141,13 @@ EnvState::Ptr KDLStateSolver::getState(const std::vector<std::string>& joint_nam
   return state;
 }
 
+StateSolver::Ptr KDLStateSolver::clone() const
+{
+  KDLStateSolver::Ptr state_solver = std::make_shared<KDLStateSolver>();
+  state_solver->init(scene_graph_);
```
mpowelson: Is this all that really has to be done to clone it? It doesn't need any of the other private members like current_state?

mpowelson: Also, should we put a mutex around this as well? Could something else be changing one of these members while you are cloning?

Levi-Armstrong:
> Is this all that really has to be done to clone it? It doesn't need any of the other private members like current_state?

Good catch, I need to set the state solver to the current state of the solver being cloned. 

> Also, should we put a mutex around this as well? Could something else be changing one of these members while you are cloning?

You are correct. Instead of adding mutex's, I restructured it such that it is thread safe. The only function that was not thread safe was the onEnvironmentChanged which is only intended to be used by the environment. I made this protected and the Environment a friend class in the StateSolver. Since the environment already has a mutex this should only ever be accessed by one thread.

Let me know if you see any issues with these changes.

mpowelson:
Ok. That seems fine. One question, why is kdl_tree_ a shared pointer? I was initially concerned that it might be passed in, but I see know that that is not the case. However, if it is created internally from a const scene graph and only used internally, why make it a pointer? Just curious.

schornakj:
Could we merge in the original Tesseract environment fix and split the KDL state solver fixes into a separate PR?

Levi-Armstrong:
> One question, why is kdl_tree_ a shared pointer?

I have not logical reason. I will change it.

mpowelson:
@Levi-Armstrong Looks good to me. Is it ready to merge?

PR #130, opened 2019-09-24
Allow tesseract_collision plugins to be found by ROS2 packages

schornakj:
This is part of a workaround for an issue in ROS2 pluginlib (https://github.com/ros/pluginlib/issues/161). Briefly, the ROS2 implementation of ClassLoader uses `ament_index_cpp::get_package_prefix()` when loading plugins, which makes it difficult to load plugins from non-ROS libraries since they aren't added to the `AMENT_PREFIX_PATH` environment variable and don't have some ROS2-specific config files in their package `share` directory.

This change solves the second of these problems by writing an empty file to `ros2_ws/install/tesseract_collision/share/ament_index/resource_index/packages/tesseract_collision` when the package is built.

I think that the first problem will need to be handled separately. So far I've been adding the path to the installed location of `tesseract_collision` to `AMENT_PREFIX_PATH` in my launch files.

### DIFF
--- a/tesseract/tesseract_collision/CMakeLists.txt
+++ b/tesseract/tesseract_collision/CMakeLists.txt
@@ -114,6 +114,8 @@ install(DIRECTORY include/${PROJECT_NAME}
 
 install(FILES package.xml DESTINATION share/${PROJECT_NAME})
 
+file(WRITE ${CMAKE_INSTALL_PREFIX}/share/ament_index/resource_index/packages/${PROJECT_NAME} "")
+
 # Create cmake config files
 include(CMakePackageConfigHelpers)
 configure_package_config_file(${CMAKE_CURRENT_LIST_DIR}/cmake/${PROJECT_NAME}-config.cmake.in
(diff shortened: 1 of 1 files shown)


tesseract/tesseract_collision/CMakeLists.txt
```diff
@@ -114,6 +114,8 @@ install(DIRECTORY include/${PROJECT_NAME}
 
 install(FILES package.xml DESTINATION share/${PROJECT_NAME})
 
+file(WRITE ${CMAKE_INSTALL_PREFIX}/share/ament_index/resource_index/packages/${PROJECT_NAME} "")
```
Levi-Armstrong: I recommend adding a comment explaining why this is required similar to what you wrote in the description on the PR.

schornakj: I added an explanatory comment with some info about the workaround and the other steps required to find the plugins from ROS2.

PR #149, opened 2019-10-14
Use GTest named targets instead of lib and include

Levi-Armstrong:
${GTEST_BOTH_LIBRARIES} becomes GTest::GTest and GTest::Main
GTEST_INCLUDE_DIRS is no longer needed


Levi-Armstrong:
@haudren I tried to fix the conflicts through the github interface but did not like it so I pulled your code and rebased on the latest. It would not let me force push to your branch so I pushed the updated branch to fork and created a PR. I will merge once CI passes.

PR #151, opened 2019-10-14
WIP: Add windows support

Levi-Armstrong:


Levi-Armstrong:
@johnwason Here is my start to incorperate your changes. I have created cmake macros for tesseract which simplify the cmake files for each of the projects. I have also created a visibility_control.h file. Now do I just need for every class is to change so it is `class TESSERACT_EXPORT` and then it will work on windows for creating shard libraries?

johnwason:
Exporting DLL functions is a little complicated. Examples can be found in the ROS 2 library:

http://wiki.ros.org/win_ros/Contributing/Dll%20Exports

https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/visibility_control.hpp

There are also helpers in CMake

https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

There is also a helper macro to create the export definitions:

https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

Levi-Armstrong:
Is there any reason not to use the option below? It seems simple to make this change for now.

> There are also helpers in CMake
https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

johnwason:
Hard to say if that option will work. It looks like you still need to export the classes anyway. 

There are a lot of functions in tesseract defined in header files that are not specified as inline, static, or a template. These need to be moved into a source file to be exported correctly. I have run into errors when I have multiple source files including tesseract headers.

johnwason:
I patched the refactored tesseract code to build on Windows. There are a few packages that wouldn't build and have been CATKIN_IGNORED.

https://github.com/johnwason/tesseract/tree/windows2

Levi-Armstrong:
Addressed in #197

PR #158, opened 2019-10-17
Add plotWaypoint

mpowelson:
This adds the ability to plot cartesian waypoints to the ROS plotter. Eventually we could add plotters for joint waypoints (perhaps as a kinematic chain).

### DIFF
--- a/tesseract_ros/tesseract_rosutils/include/tesseract_rosutils/plotting.h
+++ b/tesseract_ros/tesseract_rosutils/include/tesseract_rosutils/plotting.h
@@ -34,6 +34,7 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <ros/publisher.h>
 #include <tesseract_msgs/Trajectory.h>
 #include <tesseract_msgs/TesseractState.h>
+#include <tesseract_motion_planners/core/waypoint.h>
 #include <Eigen/Geometry>
 #include <ros/ros.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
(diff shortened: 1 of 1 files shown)


mpowelson:
@Levi-Armstrong I did not add it to the base class because then Tesseract_visualization would depend on tesseract_motion_planning which seemed like a bad idea.

tesseract_ros/tesseract_rosutils/include/tesseract_rosutils/plotting.h
```diff
@@ -72,6 +73,25 @@ class ROSPlotting : public tesseract_visualization::Visualization
...
+   */
+  void plotWaypoints(const std::vector<tesseract_motion_planners::Waypoint::Ptr>& waypoints)
+  {
+    for (auto& waypoint : waypoints)
+    {
+      if (waypoint->getType() == tesseract_motion_planners::WaypointType::CARTESIAN_WAYPOINT)
+      {
+        auto cart_wp = std::static_pointer_cast<const tesseract_motion_planners::CartesianWaypoint>(waypoint);
+        plotAxis(cart_wp->getTransform(), 0.05);
+        if (cart_wp->getParentLinkName() != "world")
```
Levi-Armstrong: This should be `if (!cart_wp->getParentLinkName().empty() && cart_wp->getParentLinkName() != env_->getSceneGraph()->getRoot())` . Also you could get the current state in this condition and update the published transform so it is in the correct frame.

mpowelson:
@Levi-Armstrong  Alright. See if you like this better.

PR #160, opened 2019-10-24
OMPL Planner Simplification

marip8:
This PR updates the OMPL planner to simplify configuration; it also refactors the unit test to be able to test all types of OMPL planners. I wasn't able to get the RRT* planner to converge in the unit test scenario, but all of the other planners work. The refactor of the OMPL planner part of an effort to create a hybrid OMPL planner that is still in progress

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/impl/ompl_freespace_planner.hpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/impl/ompl_freespace_planner.hpp
@@ -94,8 +94,7 @@ bool OMPLFreespacePlanner<PlannerType>::terminate()
 }
 
 template <typename PlannerType>
-tesseract_common::StatusCode OMPLFreespacePlanner<PlannerType>::solve(PlannerResponse& response,
-                                                                      const bool verbose)
+tesseract_common::StatusCode OMPLFreespacePlanner<PlannerType>::solve(PlannerResponse& response, const bool verbose)
 {
   tesseract_common::StatusCode config_status = isConfigured();
   if (!config_status)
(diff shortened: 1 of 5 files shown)


Levi-Armstrong:
@marip8 I like the changes, but the unit tests are failing to build.

marip8:
@Levi-Armstrong I'm having issues with the unit test failing due to several planners failing to find solutions (namely RRT, RRT* and SPARS). Is there a way to visualize or debug what's going wrong? I'm not sure if the planner parameters are just not good or if we're setting up a problem that somehow can't be solved by these planners

Levi-Armstrong:
Other than stepping through the code with a debugger. I did find this [tool](https://github.com/davetcoleman/ompl_visual_tools) but not sure it still builds.

Levi-Armstrong:
You may also try increasing the range

Levi-Armstrong:
@marip8 Based on what we learned over the past week at ROSCon should we disable these planners test until we have a chance to include the constraint/cost to ompl?

marip8:
@Levi-Armstrong let me try disabling the continuous collision checking or interpolating the trajectory to see if that's what is causing the test failures. Otherwise maybe we disable for the time being. I can work on a separate unit to test the constraint planning side of things

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/impl/ompl_freespace_planner.hpp
```diff
@@ -303,21 +305,27 @@ bool OMPLFreespacePlanner<PlannerType, PlannerSettingsType>::setConfiguration(
...
-    simple_setup_->getSpaceInformation()->setMotionValidator(std::move(mv));
+    simple_setup_->getSpaceInformation()->setMotionValidator(config_->mv);
   }
-  else if (config_->mv != nullptr)
+  else
   {
-    simple_setup_->getSpaceInformation()->setMotionValidator(config_->mv);
+    if (config_->collision_check)
+    {
+      ompl::base::MotionValidatorPtr mv;
```
Levi-Armstrong: I originally exposed the motion validator in the config so it is possible to set something else if desired. Is this still possible with these changes?

marip8: Yes, the behavior of the motion validator in the config the same as before. I just rearranged the logic here because the check of `config_->mv == nullptr` was used in several if statements in a row

PR #161, opened 2019-10-25
Exposes joint weighting

DavidMerzJr:
@Levi-Armstrong 

The changes made earlier today.


Levi-Armstrong:
@johnwason Is it preferred to use std::vector<double> or Eigen::VectorXd for SWIG C++ to python?

johnwason:
Either way can work, but Eigen::VectorXd is probably more of a canonical representation.

Levi-Armstrong:
@johnwason Thank you. I update it to use Eigen.

Levi-Armstrong:
See #165

PR #166, opened 2019-10-30
Trajopt Planner: Switch setConfiguration to pass shared_ptr by value

mpowelson:
When passed by reference, calling clear on the planner also clears the config that was passed in. If it is by value, you will just be setting the planners config to nullptr not the original.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/trajopt_motion_planner.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/trajopt_motion_planner.h
@@ -54,7 +54,7 @@ public:
    * @param config The planners configuration
    * @return True if successful otherwise false
    */
-  bool setConfiguration(const TrajOptPlannerConfig::Ptr& config);
+  bool setConfiguration(const TrajOptPlannerConfig::Ptr config);
 
   /**
    * @brief Sets up the opimizer and solves a SQP problem read from json with no callbacks and dafault parameterss
--- a/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/trajopt_motion_planner.cpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/trajopt_motion_planner.cpp
@@ -191,7 +191,7 @@ tesseract_common::StatusCode TrajOptMotionPlanner::isConfigured() const
     return tesseract_common::StatusCode(TrajOptMotionPlannerStatusCategory::IsNotConfigured, status_category_);
 }
 
-bool TrajOptMotionPlanner::setConfiguration(const TrajOptPlannerConfig::Ptr& config)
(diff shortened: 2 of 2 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/trajopt_motion_planner.h
```diff
@@ -54,7 +54,7 @@ class TrajOptMotionPlanner : public MotionPlanner
    * @param config The planners configuration
    * @return True if successful otherwise false
    */
-  bool setConfiguration(const TrajOptPlannerConfig::Ptr& config);
+  bool setConfiguration(const TrajOptPlannerConfig::Ptr config);
```
Levi-Armstrong: It may be good to remove the const so it can be moved into config_.

mpowelson: Ok. All of the discussion about RVO before got me curious. I put this into godbolt as well. All 4 of these seem to do the same in GCC 9.2. Maybe I'm doing something wrong
```
    void setBarMember1(const std::shared_ptr<Bar> input_bar) {bar_member = input_bar;}
    void setBarMember2(const std::shared_ptr<Bar> input_bar) {bar_member = std::move(input_bar);}
    void setBarMember3(std::shared_ptr<Bar> input_bar) {bar_member = input_bar;}
    void setBarMember4(std::shared_ptr<Bar> input_bar) {bar_member = std::move(input_bar);}
```
[move_vs_assign_in_setMember.txt](https://github.com/ros-industrial-consortium/tesseract/files/3794158/move_vs_assign_in_setMember.txt)

Levi-Armstrong: I think that is because in gcc 9.2 uses c++17 with guaranteed NRVO (named return value optimization) which is probably why you get the same results. If you add the compiler warnings below you most likely would get a warning that the std::move is redundant. 

-Wpessimizing-move -Wredundant-move

Levi-Armstrong: Though prior to c++17 this was not the case.

mpowelson: I added those options, but I don't see where compiler warnings are printed. In the compiler output it just says "compiler returned 0". I was under the impression that most major compilers already did NRVO but C++17 just added it to the spec. It looks like GCC 9 is available in Ubuntu 19, not 16, or 18. 

Regardless, I'll follow your recommendation. What would you like me to do?

Levi-Armstrong: Actually I think what you have is good.

PR #167, opened 2019-10-30
Process Planner: Generating add from_start to generateProcessDefinition

mpowelson:


tesseract/tesseract_planning/tesseract_process_planners/src/tesseract_process_planners/process_definition.cpp
```diff
@@ -56,6 +56,25 @@ ProcessDefinition generateProcessDefinition(const ProcessDefinitionConfig& proce
           transition_def.transition_from_end = process_config.transition_generator[i]->generate(
               process_definition.segments[i].departure.back(), process_definition.segments[i + 1].approach.front());
 
+        // Connect the appropriate waypoints for the transitions from start
```
Levi-Armstrong: It may be good to also update/add the same comments to the function below.

mpowelson: I didn't notice we had 2 copies. Done.

Issue #148, opened 2019-10-14
Return file contents instead of filename for tesseract_scene_graph::ResourceLocatorFn


johnwason:
Currently tesseract_scene_graph::ResourceLocatorFn returns the filename to a resource instead of the resource contents. This is inflexible since the resource may be stored in a zip file or retrieved from a URL. The locator function should return the file contents as a byte array. Assimp can load in the raw bytes using ReadFileFromMemory so there shouldn't be a compatibility issue.

Levi-Armstrong:
@johnwason I don't have an issue with change. @mpowelson, @schornakj  Do you see any issues with this change?

johnwason:
Can we use std::vector<uint8_t> as the return type?

mpowelson:
It would seem to me that it might be a bit of a misnomer if we changed it to return the data instead of the location. Also, currently you could use it to get the filepath of a directory to write data to in place of something like rospack. I imagine we are using it that way in a few unit tests.

Can you overload a std::function definition? Then we could have one that returns a byte array and one that returns a string. Or we could make the things using the ResourceLocatorFn be smarter and know how to parse URLs and extract zip files. Perhaps some convenient utilities are in order.

johnwason:
I plan to use a custom resource loader, so having that logic within the tesseract functions would not be great. Could be define a ResourceLoader class with "IsFile", "GetFilePath", "GetResourceData" functions?

Levi-Armstrong:
I took a look at ros resource retriever and this is what they do to address this issue. It looks like they use `std::vector<uint8_t>` when initially retrieving the data then they write it `boost::shared_array<uint8_t> data`. I assume they use a shared array to prevent from coping data. Should we return a shared pointer?

https://github.com/ros/resource_retriever/blob/kinetic-devel/include/resource_retriever/retriever.h
https://github.com/ros/resource_retriever/blob/kinetic-devel/src/retriever.cpp

johnwason:
`boost::shared_array<uint8_t>` does not contain length information. It would have to be `std::shared_ptr<std::vector<uint8_t> >` or similar.

Levi-Armstrong:
> I plan to use a custom resource loader, so having that logic within the tesseract functions would not be great. Could be define a ResourceLoader class with "IsFile", "GetFilePath", "GetResourceData" functions?

This change will work.

johnwason:
I will implement this option and issue a pull request.

johnwason:
[...]

PR #172, opened 2019-11-07
Use ResourceLocator class instead of ResourceLocatorFn

johnwason:
This patch introduces `ResourceLocator` and `LocatedResource` classes to replace `ResourceLocatorFn`. This allows for more general resource loading, such as over the network instead of expecting a file.

Because `ResourceLocatorFn` is used in `trajopt_ros`, a patch must be applied there as well. I will issue a pull request. Due to the dependency on updates in `trajopt_ros` the CI tests will fail until both have been accepted.


Levi-Armstrong:
@johnwason This is great! I will review it and provide feedback if any. Also I will pull both tesseract and trajopt PR to confirm unit test pass and post results.

tesseract/tesseract_common/include/tesseract_common/resource.h
```diff
@@ -0,0 +1,31 @@
...
+namespace tesseract_common
+{
+// Represents resource data available from a file or url
+class Resource
+{
+public:
+  using Ptr = std::shared_ptr<Resource>;
+  using ConstPtr = std::shared_ptr<const Resource>;
+
+  virtual bool IsFile() = 0;
```
Levi-Armstrong: I recommend renaming the functions to match the ROS Cpp style guide. This function would be `isFile()`.

tesseract/tesseract/include/tesseract/tesseract.h
```diff
@@ -60,14 +61,14 @@ class Tesseract
 
   bool init(tesseract_scene_graph::SceneGraph::Ptr scene_graph);
   bool init(tesseract_scene_graph::SceneGraph::Ptr scene_graph, tesseract_scene_graph::SRDFModel::ConstPtr srdf_model);
-  bool init(const std::string& urdf_string, tesseract_scene_graph::ResourceLocatorFn locator);
+  bool init(const std::string& urdf_string, tesseract_scene_graph::ResourceLocator::Ptr locator);
```
Levi-Armstrong: Is there any reason these should not be ConstPtr?

tesseract/tesseract_geometry/include/tesseract_geometry/impl/convex_mesh.h
```diff
@@ -92,7 +93,7 @@ class ConvexMesh : public Geometry
    *
    * @return Absolute path to the mesh file
    */
-  const std::string& getFilePath() const { return file_path_; }
+  const tesseract_common::Resource::Ptr getResource() const { return resource_; }
```
Levi-Armstrong: Should this be a ConstPtr also? The function is marked const.

tesseract/tesseract_common/include/tesseract_common/resource.h
Levi-Armstrong: Also should all these functions be const?

tesseract/tesseract_geometry/include/tesseract_geometry/impl/mesh.h
```diff
@@ -104,15 +105,15 @@ class Mesh : public Geometry
...
+    return std::make_shared<Mesh>(vertices_, triangles_, triangle_count_, resource_, scale_);
   }
 
 private:
   std::shared_ptr<const tesseract_common::VectorVector3d> vertices_;
   std::shared_ptr<const Eigen::VectorXi> triangles_;
   int vertice_count_;
   int triangle_count_;
-  std::string file_path_;
+  tesseract_common::Resource::Ptr resource_;
```
Levi-Armstrong: ConstPtr?

tesseract/tesseract_scene_graph/src/resource_locator.cpp
```diff
@@ -0,0 +1,87 @@
+/**
+ * @file graph.cpp
```
Levi-Armstrong: resource_locator.cpp

gavanderhoorn:
This is not a :-1: or anything like that, but for future readers coming here (and for me in the present): this seems to load files into memory now instead of returning a filename. What would be the main difference with [ros/resource_retriever](https://github.com/ros/resource_retriever)?

From its manifest:

> This package retrieves data from url-format files such as http://, ftp://, package:// file://, etc., and loads the data into memory. The package:// url for ros packages is translated into a local file:// url.  The resourse retriever was initially designed to load mesh files into memory, but it can be used for any type of data. The resource retriever is based on the the libcurl library.

tesseract/tesseract_common/include/tesseract_common/resource.h
johnwason: I will update to use the ROS Cpp style.

No, I am not using const because it is likely that the located resources will be modifying cache data during operations.

johnwason:
@gavanderhoorn the need for these changes is discussed in https://github.com/ros-industrial-consortium/tesseract/issues/148. RPI and my company are working on a project that uses a custom resource system and needs more flexibility. By making `ResourceLocator` an abstract is class is makes the implementation more flexible. `SimpleResourceLocator` and `ROSResourceLocator` implement the existing tesseract resource loading behavior.

johnwason:
@Levi-Armstrong The `ResourceLocator` and `Resource` should not be const because they will be caching data during operations.

tesseract/tesseract_common/include/tesseract_common/resource.h
johnwason: The style should be updated now

tesseract/tesseract_scene_graph/src/resource_locator.cpp
johnwason: Fixed

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -0,0 +1,65 @@
...
+  using Ptr = std::shared_ptr<SimpleLocatedResource>;
+  using ConstPtr = std::shared_ptr<const SimpleLocatedResource>;
+
+  SimpleLocatedResource(const std::string& url, const std::string& filename);
+
+  virtual bool isFile() override;
+
+  virtual std::string getUrl() override;
+
+  virtual std::string getFilePath() override;
```
Levi-Armstrong: I just have concern that this may need to be used within a const function which would not be allowed since these are not const. 

I would expect that the three function isFile(), getUrl() and getFilePath() should be able to be const?

I can see where getResourceContents and getResourceContentStream but I believe in the case of caching data it is permitted to make these function const and use mutable on the cached object along with mutex. What are your thoughts?

gavanderhoorn:
> The need for these changes is discussed in #148.

Is it though? The first comment in that issue seems to state an already chosen approach. There doesn't seem to be a discussion of why this is needed.

The last comment (also by you) also seems to only describe a chosen approach.

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
johnwason: Resolving the resource may also result in cached value like the real filename or real url. This would be stored in the `ResourceLocator`. The actual data would be cached in the `Resource`, and the `ResourceLocator` may cache the retrieved data in case it is requested again.

johnwason:
@gavanderhoorn our application will be using directories, zip files, or Robot Raconteur [com.robotraconteur.resource.ResourceStorage](https://github.com/johnwason/robotraconteur_standard_robdef/blob/261cbb857d53a892937c4dca21eba87836c8959c/group1/com.robotraconteur.resource.robdef#L51) services. This project is not using ROS. Using an abstract resource locator class makes it flexible enough for our use case. SwRI is part of the Robot Raconteur project so @Levi-Armstrong is working with us to adapt Tesseract for our use case.

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
Levi-Armstrong: @gavanderhoorn and @johnwason Do either you see an issue with making these const and leverage mutable when caching?

Levi-Armstrong:
@johnwason I wanted to test the caching component and wanted to see how best to approach it. I wanted create a URDF with which uses the same collision object for for multiple links. When the resource is first requested cache the results to be reused. Should the caching happen in the resource or resource locator? After looking at the two classes, I believe the answer is you would cache data in the resource (binary, full path, etc.) and then in the resource locator you would maintain a map of url to resource so when locateResource is called you check the map and if present immediately return and if not create the resource and store it in the map. Is this the correct approach?

johnwason:
@Levi-Armstrong for resources on the filesystem it makes sense to keep a map of the file location since there isn't much of a penalty to retrieve the data. For retrieving data over the network it would make sense to keep a copy in memory of the actual data.

Would you prefer using a singleton cache so `Resource` and `ResourceLocator` remain const? This design would solve the const problem but breaks the spirit of const to certain extent.

Levi-Armstrong:
@johnwason I think it is good as is. I checked out how rviz leverages the resource_retriever which is similar to this. The resource_retriever methods are not const and in rviz they make the retriever object mutable which I will take this approach and if challenges are found we can address them at that time.

tesseract/tesseract_common/include/tesseract_common/resource.h
```diff
@@ -0,0 +1,31 @@
+#ifndef TESSERACT_COMMON_RESOURCE_H
```
Levi-Armstrong: I would add the license.

tesseract/tesseract_common/include/tesseract_common/resource.h
```diff
@@ -0,0 +1,31 @@
...
+namespace tesseract_common
+{
+// Represents resource data available from a file or url
+class Resource
+{
+public:
+  using Ptr = std::shared_ptr<Resource>;
+  using ConstPtr = std::shared_ptr<const Resource>;
+
+  virtual bool isFile() = 0;
```
Levi-Armstrong: I recommend adding doxygen to each of the class methods.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -203,7 +206,96 @@ inline std::vector<std::shared_ptr<T>> createMeshFromPath(const std::string& pat
...
+ * @brief Create a mesh using assimp from resource
+ * @param resource The located resource
+ * @param scale Perform an axis scaling
+ * @param trianglulate If true the mesh will be trianglulated. This should be done for visual meshes.
+ *        In the case of collision meshes do not triangulate convex hull meshes.
+ * @param flatten If true all meshes will be condensed into a single mesh. This should only be used for visual meshes,
+ * do not flatten collision meshes.
+ * @return
+ */
+
```
Levi-Armstrong: not sure if the extra space will cause an issue when generating doxygen documentation.

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -0,0 +1,65 @@
+#ifndef TESSERACT_SCENE_GRAPH_RESOURCE_LOCATOR_H
```
Levi-Armstrong: I recommend adding the license.

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -0,0 +1,65 @@
...
+#define TESSERACT_SCENE_GRAPH_RESOURCE_LOCATOR_H
+
+#include <tesseract_common/resource.h>
+#include <tesseract_scene_graph/allowed_collision_matrix.h>
+#include <tesseract_scene_graph/graph.h>
+#include <tesseract_scene_graph/parser/srdf_parser.h>
+#include <functional>
```
Levi-Armstrong: I would wrap this in the macro to prevent warnings generated form system dependencies.

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace tesseract_scene_graph
+{
+// ResourceLocator abstract class
```
Levi-Armstrong: Recommend using doxygen

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -0,0 +1,65 @@
...
+class ResourceLocator
+{
+public:
+  using Ptr = std::shared_ptr<ResourceLocator>;
+  using ConstPtr = std::shared_ptr<const ResourceLocator>;
+
+  virtual tesseract_common::Resource::Ptr locateResource(const std::string& url) = 0;
```
Levi-Armstrong: I recommend adding doxygen.

Levi-Armstrong:
@johnwason I made a few request and then it should be ready to merge. Thank you for working on this.

johnwason:
@Levi-Armstrong I have made the requested changes.

Levi-Armstrong:
@johnwason I have pulled down and all of the unit tests pass for both tesseract and trajopt. The only test that does not pass is the clang. I will see if I can create a PR.

Levi-Armstrong:
@johnwason I created a PR against your branch.

johnwason:
@Levi-Armstrong I merged your request

PR #180, opened 2019-11-19
Fix quaternion error and add test

Levi-Armstrong:


tesseract/tesseract_urdf/test/tesseract_urdf_origin_unit.cpp
```diff
@@ -7,6 +7,23 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+  the = pitch / 2.0;
+  psi = yaw / 2.0;
+
+  double x, y, z, w;
+  x = std::sin(phi) * std::cos(the) * std::cos(psi) - std::cos(phi) * std::sin(the) * std::sin(psi);
+  y = std::cos(phi) * std::sin(the) * std::cos(psi) + std::sin(phi) * std::cos(the) * std::sin(psi);
+  z = std::cos(phi) * std::cos(the) * std::sin(psi) - std::sin(phi) * std::sin(the) * std::cos(psi);
+  w = std::cos(phi) * std::cos(the) * std::cos(psi) + std::sin(phi) * std::sin(the) * std::sin(psi);
+
+  return Eigen::Quaterniond(w, x, y, z);
```
marip8: Not that it makes much difference (this being a test), but it's probably way easier and less prone to mistakes to let Eigen do the work for you here:
```c++
return = Eigen::AngleAxisd(roll, Eigen::Vector3d::UnitX()) * Eigen::AngleAxisd(pitch, Eigen::Vector3d::UnitY()) * Eigen::AngleAxisd(yaw, Eigen::Vector3d::UnitZ()); 
```

According to the "Transformation Types" section of the [Eigen documentation](https://eigen.tuxfamily.org/dox/group__TutorialGeometry.html) this should work. Just something to consider

Levi-Armstrong: I copied this calculation for urdfdom just for the test to confirm that the origin class when parsing rpy is producing the same rotation matrix that would be generated from urdfdom.

Levi-Armstrong: @mpowelson Do you want me to add a comment to function explaining its purpose?

mpowelson: I imagine that comment was intended for @marip8 . But I'm going to say if there's ever a question on whether there should be a comment, my answer is yes.

Levi-Armstrong: @marip8 I add a documentation to explain the function and its purpose within the unit test.

marip8: Sounds good. I just thought this was a helper function that could be simplified

PR #185, opened 2019-12-09
Replace raw lock / unlock by lock_guard

haudren:
This is to avoid potential deadlock when returning early
e.g. when deleting a non-existing link.

Currently, if any of the calls to `Environment` that's mutex-protected and can fail does fail, the mutex is left locked, and any subsequent calls will fail.

Also updated the test to check double-removing, maybe it should be done for other functions as well.

### DIFF
--- a/tesseract/tesseract_environment/src/core/environment.cpp
+++ b/tesseract/tesseract_environment/src/core/environment.cpp
@@ -32,52 +32,46 @@ namespace tesseract_environment
 {
 void Environment::setState(const std::unordered_map<std::string, double>& joints)
 {
-  mutex_.lock();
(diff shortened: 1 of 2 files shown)


haudren:
Not sure what up with Travis btw, seems to have failed on installing `tesseract_motion_planners` ?

Levi-Armstrong:
The tesseract_motion_planners unit test failed. I found an issue in the ompl that is in my latest pull request that should solve this issue. I will restart the test.

haudren:
Thanks for this, tests passed now :+1: If you don't have any issues, I think it would be good to merge quickly as those deadlocks can be pretty painful.

drchrislewis:
@Levi-Armstrong isn't this ready for merge?

PR #190, opened 2019-12-18
Expose the contact distance threshold for DescartesCollision

schornakj:
Add a parameter to the constructor of `tesseract_motion_planners::DescartesCollision` that sets the contact distance threshold of its underlying `DiscreteContactManager`. This is useful if you want Descartes to provide a seed trajectory that leaves a buffer vs. collision geometry near the process waypoints, since I think the distance threshold defaults to 0.0 m otherwise.


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/descartes_collision.h
```diff
@@ -98,6 +99,7 @@ class DescartesCollision : public descartes_light::CollisionInterface<FloatType>
   std::vector<std::string> active_link_names_;                       /**< @brief A vector of active link names */
   std::vector<std::string> joint_names_;                             /**< @brief A vector of joint names */
   tesseract_collision::DiscreteContactManager::Ptr contact_manager_; /**< @brief The discrete contact manager */
+  double contact_dist_threshold_;
```
Levi-Armstrong: I recommend adding doxygen.

schornakj: Added!

PR #194, opened 2019-12-23
Use state solver in descartes edge evaluator and ompl motion validator

Levi-Armstrong:
The environment is thread safe but the state solver and contact managers are not so they must be cloned for each thread that is being used.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/continuous_motion_validator.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/continuous_motion_validator.h
@@ -42,7 +42,7 @@ class ContinuousMotionValidator : public ompl::base::MotionValidator
 {
 public:
   ContinuousMotionValidator(const ompl::base::SpaceInformationPtr& space_info,
-                            tesseract_environment::Environment::ConstPtr env,
+                            const tesseract_environment::Environment::ConstPtr& env,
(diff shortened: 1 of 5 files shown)


gavanderhoorn:
Could you perhaps add some comments on the rationale for this change? The code doesn't add comments that explain why this is a needed/good change, and neither does the PR description.

Levi-Armstrong:
@gavanderhoorn I added a description to the PR explaining why the change is needed.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/continuous_motion_validator.h
```diff
@@ -71,6 +71,9 @@ class ContinuousMotionValidator : public ompl::base::MotionValidator
   /** @brief The Tesseract Environment */
   tesseract_environment::Environment::ConstPtr env_;
 
+  /**< @brief The tesseract state solver */
+  tesseract_environment::StateSolver::Ptr state_solver_;
```
marip8: Do you do anything with the `state_solver_` object besides `clone`? If not, you might consider making this a `ConstPtr` since `clone` is a `const` function

Levi-Armstrong: That is probably a good idea. I will make the change.

PR #195, opened 2019-12-23
TrajOpt Config Update: Avoid Singularity 

marip8:
This PR updates the TrajOpt Default Configuration class with a boolean and coefficient value for using the new TrajOpt avoid singularity cost


marip8:
The existing TrajOpt array planner test asserts that the planning problem does not contain any costs of type `TrajOptCostFromErrFunc`. This PR adds the avoid singularity cost (which is a `TrajOptCostFromErrFunc`) to the default planner config (used by the test, and "on" by default), so it is no longer true that the problem does not contain any costs of the `TrajOptCostFromErrFunc` variety

Levi-Armstrong:
I am not opposed to make it default, but I just want to check before merging since this will change the current behavior. Have you been using this on your current development without any issues?

marip8:
I have been using recently without any issues, but I can't say I've tested it extensively. I can change the default value of avoid singularity to false until we do more testing with it if you want

PR #203, opened 2020-01-09
Update trajopt planner handling of fixed start and end states for collision

Levi-Armstrong:
This add ability to handle fixed states throughout the provided trajectory for collision. It also simplifies the trajopt configs to reduce code duplication. 

Update:
Noticed that there is an intermittent bug in OMPL hybridization which will return a plan that starts at the end and finishes at the end. Need to create issue on OMPL. 

Updated tesseract to support bullet being built using double precision.

Update rviz link_widget to support tesseract octree sub shape type SPHERE_INSIDE and SPHERE_OUTSIDE visualization.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
@@ -141,7 +141,7 @@ protected:
   bool addBasicInfo(trajopt::ProblemConstructionInfo& pci) const;
   bool addInitTrajectory(trajopt::ProblemConstructionInfo& pci) const;
   void addWaypoints(trajopt::ProblemConstructionInfo& pci, std::vector<int>& fixed_steps) const;
-  void addConfiguration(trajopt::ProblemConstructionInfo& pci, const std::vector<int>& fixed_steps) const;
+  void addKinematicConfiguration(trajopt::ProblemConstructionInfo& pci, const std::vector<int>& fixed_steps) const;
   void addCollision(trajopt::ProblemConstructionInfo& pci, const std::vector<int>& fixed_steps) const;
   void addVelocitySmoothing(trajopt::ProblemConstructionInfo& pci, const std::vector<int>& fixed_steps) const;
   void addAccelerationSmoothing(trajopt::ProblemConstructionInfo& pci, const std::vector<int>& fixed_steps) const;
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
@@ -93,7 +93,7 @@ struct TrajOptPlannerDefaultConfig : public TrajOptPlannerConfig
   /** @brief If true, collision checking will be enabled. Default: true*/
   bool collision_check = true;
   /** @brief If true, use continuous collision checking */
-  bool collision_continuous = true;
+  trajopt::CollisionEvaluatorType collision_type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
   /** @brief Max distance over which collisions are checked */
   double collision_safety_margin = 0.025;
   /** @brief The collision coeff/weight */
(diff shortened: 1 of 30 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
```diff
@@ -135,6 +135,19 @@ struct TrajOptPlannerDefaultConfig : public TrajOptPlannerConfig
    */
   std::vector<std::tuple<sco::VectorOfVector::func, sco::MatrixOfVector::func, sco::ConstraintType, Eigen::VectorXd>>
       constraint_error_functions;
+
+protected:
+  bool checkUserInput() const;
+  bool addBasicInfo(trajopt::ProblemConstructionInfo& pci) const;
+  bool addInitTrajectory(trajopt::ProblemConstructionInfo& pci) const;
+  void addWaypoints(trajopt::ProblemConstructionInfo& pci, std::vector<int>& fixed_steps) const;
```
mpowelson: `void addWaypoints(trajopt::ProblemConstructionInfo& pci, std::vector<int>& fixed_steps = std::vector<int>())` ?

tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
```diff
@@ -142,143 +206,185 @@ std::shared_ptr<trajopt::ProblemConstructionInfo> TrajOptPlannerDefaultConfig::g
...
+    if (target_waypoints[ind]->getType() == WaypointType::JOINT_WAYPOINT)
+    {
+      fixed_steps.push_back(static_cast<int>(ind));
+    }
   }
-  if (target_waypoints.back()->getType() == WaypointType::JOINT_WAYPOINT ||
-      target_waypoints.back()->getType() == WaypointType::JOINT_TOLERANCED_WAYPOINT)
+}
+
+void TrajOptPlannerDefaultConfig::addConfiguration(trajopt::ProblemConstructionInfo& pci,
```
mpowelson: Perhaps we should make this addConfigurationCost or addKinematicConfiguration. addConfiguration sounds like you are adding a "config".

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
Levi-Armstrong: I assume you are referring the fixed_steps parameter. It should be removed.

tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
Levi-Armstrong: Sure.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
Levi-Armstrong: Sorry I was incorrect. This is required. The fixed_steps are defined by the type of waypoints. If provided a waypoint of type JointWaypoint if would be set as a fixed_step all other types at the moment are considered not fixed steps.

tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
Levi-Armstrong: done

Levi-Armstrong:
@johnwason Do you mind taking a look at the changes to the python in latest commit (Upate tesseract_python)?

Levi-Armstrong:
@johnwason I also found out why SWIG was not being found. The first issue was that FindSWIG looked for swig3.0 then swig2.0 then swig which is why it would find swig3.0 before the new version because it was named swig. I updated the install to rename swig executable to swig4.0 and added a new FindSWIG that looks for swig4.0 the others. With these changes it is no longer required to pass the compiler flag.

Levi-Armstrong:
@mpowelson Good to merge?

mpowelson:
@Levi-Armstrong Sure.

Levi-Armstrong:
I will remove the ci commit and merge.

johnwason:
@Levi-Armstrong the SWIG updates look fine

Levi-Armstrong:
@johnwason Thanks for taking a look.

hzyjerry:
I was able to build successfully on ubuntu 16.04. Thanks for the effort.

The key step was updating from swig3.0. To do that I downloaded latest swig4.0 from [this tutorial](https://github.com/swig/swig/wiki/Getting-Started). To install it, I also downloaded latest pcre based on [this issue](https://github.com/swig/swig/issues/1175)

PR #208, opened 2020-01-21
Add initial support for Groups and GroupStates

haudren:
This commit adds a new binding for srdf_parser.h
Current support is:
- Retrieving Groups and GroupStates
- Accessing their members
Note that due to SWIG's lack of support of nested classes, they end up
flattened, at the root of the module.

This PR still lacks a few things, namely tests but I'd like to get a few comments on this code first.

### DIFF
--- a/tesseract_python/tesseract_python/swig/tesseract/tesseract.i
+++ b/tesseract_python/tesseract_python/swig/tesseract/tesseract.i
@@ -54,7 +54,7 @@ public:
   //          const boost::filesystem::path& srdf_path,
   //          tesseract_scene_graph::ResourceLocator::Ptr locator);
 
-  //const tesseract_scene_graph::SRDFModel::ConstPtr& getSRDFModel() const;
+  const tesseract_scene_graph::SRDFModel::ConstPtr& getSRDFModel() const;
 
   const tesseract_environment::Environment::Ptr getEnvironment();
   const tesseract_environment::Environment::ConstPtr getEnvironmentConst() const;
--- a/tesseract_python/tesseract_python/swig/tesseract_python.i
+++ b/tesseract_python/tesseract_python/swig/tesseract_python.i
@@ -78,6 +78,7 @@
 %include "tesseract_geometry/geometry_loaders.i"
 %include "tesseract_scene_graph/graph.i"
 %include "tesseract_scene_graph/resource_locator.i"
+%include "tesseract_scene_graph/srdf_parser.i"
 %include "tesseract_kinematics/core/forward_kinematics.i"
 %include "tesseract_kinematics/core/forward_kinematics_factory.i"
 %include "tesseract_kinematics/core/inverse_kinematics.i"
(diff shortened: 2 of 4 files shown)


Levi-Armstrong:
@johnwason Do you mind taking a look at this PR?

Levi-Armstrong:
Would moving the structs outside the class help?

Levi-Armstrong:
@haudren  Just a side note. You can also get manipulator group names using the `getAvailableFwdKinematicsManipulators` in the forward and inverse kinematics managers.

johnwason:
Huh, this didn't work for me when I tried it. Can you read the values in Group and GroupState?

haudren:
Interesting, I managed to get the values of the GroupState for sure... I'll have another look.

johnwason:
Sometimes SWIG will return an opaque pointer if it doesn't understand the type. Can you read the fields in Group and GroupState?

haudren:
OK I just pushed another update, I was indeed missing a couple of template bindings, which led to opaque pointers appearing. I fixed those, and added some tests which show how you can recover the SRDF information from Python.

johnwason:
Most of the std templates are already defined in other *.i files. Can you remove the duplicates?

haudren:
I would, but how do I go about it? When I did not include them, I believe I got some opaque pointers... I'll give it another go.

Levi-Armstrong:
@johnwason What is the best way to address your comment?

johnwason:
My suggested corrections:

* Move the std::pair<std::string,std::string> templates to tesseract_python.i where vector_string is defined.
* std::vector<double> has a custom typedef that should be used instead of overriding the definition. See https://github.com/ros-industrial-consortium/tesseract/blob/772ef43cbe4c270deea223560c02d9f35309e4a2/tesseract_python/tesseract_python/swig/trajopt/problem_description.i#L25

The rest look fine.

haudren:
So I tried to address your comment @johnwason but I keep on hitting this error
```
error: ‘type_name’ is not a member of ‘swig::traits<double>’
     return traits<typename noconst_traits<Type >::noconst_type >::type_name();
```
Which goes away when I explicitly instantiate the `std::vector<double>` template. I tried to play around with the typemaps:
- Move them from `problem_description.i` to `tesseract_python.i`
- Create my own custom typemap similar to what you did but for `std::vector<double>` (not a pointer)
- Do the above and introduce a typedef to alias to a non-templated type
All to no avail

I think I could use some help here...

Levi-Armstrong:
@haudren It looks like CI is passing. Is the issue resolved or are you still waiting on a response from @johnwason?

johnwason:
I will take a look at this today.

johnwason:
Everything looks good to me. The only problem I found is that the `test_srdf_parser.py` doesn't actually call the test function. The following is required at the bottom of the file:

    if __name__ == "__main__":
        test_environment()

johnwason:
CMake sometimes won't detect that a swig .i file has changed if the root file isn't modified. Try touching the tesseract_python.i file and see if that helps.

    touch -c tesseract_python.i

Levi-Armstrong:
@johnwason Thank you for looking into this.

Levi-Armstrong:
@haudren Looks just one thing to add before this can be merged.

haudren:
Fixed it thanks :+1:

PR #210, opened 2020-01-21
Allow adding TrajOpt collision terms as both constraints and costs

schornakj:
Add separate collisions terms for constraint and cost and expose them in the TrajOpt planner configs. It's important for collisions to be represented as constraints to ensure that the resulting trajectory is collision-free. In certain cases it is desirable for collisions to also be modeled as costs to drive the solution further away from nearby collisions if possible.

- Add new functions to allow adding collision terms as either costs or constraints.
- ~~Added new parameters in the TrajOpt planner configs to set if collisions will be disabled, modeled as constraints, or modeled as both costs and constraints.~~
- ~~If collisions are enabled they will be modeled as constraints. If `collision_check = true`, a collision constraint term will be added with a safety margin of `collision_safety_margin`.~~
- ~~Add a new flag to the TrajOpt planner config to enable a separate collision cost term. If `use_collision_cost_term = true`, an additional collision cost term will be added with a safety margin of `collision_buffer_margin`.~~
- Add new config structs to represent settings for collision costs and constraints. This allows collision costs and collision constraints to be independently enabled or disabled.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
@@ -59,6 +59,6 @@ struct CollisionConstraintConfig
   /** @brief The collision coeff/weight */
   double coeff = 20;
 };
-}
+}  // namespace tesseract_motion_planners
 
-#endif // TESSERACT_MOTION_PLANNERS_TRAJOPT_CONFIG_TRAJOPT_COLLISION_CONFIG_H
+#endif  // TESSERACT_MOTION_PLANNERS_TRAJOPT_CONFIG_TRAJOPT_COLLISION_CONFIG_H
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
@@ -1,5 +1,5 @@
 /**
- * @file trajopt_collision_config_base.h
+ * @file trajopt_collision_config.h
  * @brief TrajOpt collision configuration settings
  *
  * @author Joseph Schornak
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
@@ -11,7 +11,7 @@ namespace tesseract_motion_planners
 struct CollisionCostConfig
 {
   /** @brief If true, a collision cost term will be added to the problem. Default: true*/
-  bool check = true;
+  bool enabled = true;
   /** @brief The evaluator type that will be used for collision checking. */
   trajopt::CollisionEvaluatorType type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
   /** @brief Max distance in which collision costs will be evaluated. */
@@ -26,7 +26,7 @@ struct CollisionCostConfig
 struct CollisionConstraintConfig
 {
   /** @brief If true, a collision cost term will be added to the problem. Default: true*/
-  bool check = true;
+  bool enabled = true;
   /** @brief The evaluator type that will be used for collision checking. */
   trajopt::CollisionEvaluatorType type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
   /** @brief Max distance in which collision constraints will be evaluated. */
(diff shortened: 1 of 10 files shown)


gavanderhoorn:
Hm. This sounds familiar @Levi-Armstrong ?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
```diff
@@ -90,12 +90,16 @@ struct TrajOptPlannerDefaultConfig : public TrajOptPlannerConfig
...
   bool collision_check = true;
+  /** @brief If true, a collision cost term will be added to the problem. Default: false*/
+  bool use_collision_cost_term = false;
   /** @brief If true, use continuous collision checking */
   trajopt::CollisionEvaluatorType collision_type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
-  /** @brief Max distance over which collisions are checked */
+  /** @brief Max distance in which collisions will be evaluated as constraints */
   double collision_safety_margin = 0.025;
+  /** @brief Max distance in which collisions will be evaluated as costs if use_collision_cost_term is true */
+  double collision_buffer_margin = 0.05;
```
Levi-Armstrong: I recommend have different parameters for each.

Constraint Config. If you want to put these in a struct that is also ok.
```
collision_constraint_check
collision_constraint_type
collision_constraint_safety_margin
collision_constraint_safety_margin_buffer
collision_constraint_coeff
```
or 
```
struct CollisionConstraintConfig
{
  bool check = true;
  trajopt::CollisionEvaluatorType type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
  double safety_margin = 0.01;
  double safety_margin_buffer = 0.05;
  double coeff = 20;
}
```

Cost Config. If you want to put these in a struct that is also ok.
```
collision_cost_check
collision_cost_type
collision_cost_buffer_margin
collision_cost_coeff
```
or 
```
struct CollisionCostConfig
{
  bool check = true;
  trajopt::CollisionEvaluatorType type = trajopt::CollisionEvaluatorType::CAST_CONTINUOUS;
  double buffer_margin = 0.025;
  double coeff = 20;
}
```

schornakj:
> Do you plan to add the code from my branch that would leverage the constraint safety margin buffer in this PR?

Which branch was that in? I thought that that change was only in trajopt_ros.

In any case I think it would be better to make a separate PR.

Levi-Armstrong:
Yea, the changes on the TrajOpt side shown [here](https://github.com/ros-industrial-consortium/trajopt_ros/commit/958567389952b081f8b297a844aa13dacbd9f905).

Levi-Armstrong:
> Hm. This sounds familiar @Levi-Armstrong ?

I assume your talking about the hierarchical solver that you had mentioned?

Levi-Armstrong:
> In any case I think it would be better to make a separate PR.

It may be good to do it in this PR because the function and struct will need to change again to support the change in trajopt.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_default_config.h
Levi-Armstrong: Maybe instead of check we change to enabled?

schornakj: I added some config structs to separately wrap settings for costs and constraints.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
```diff
@@ -0,0 +1,39 @@
+#ifndef TESSERACT_MOTION_PLANNERS_TRAJOPT_CONFIG_TRAJOPT_COLLISION_CONFIG_H
```
Levi-Armstrong: Add license

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace tesseract_motion_planners
+{
+/**
+ * @brief Config settings for collision cost terms.
+ */
+struct CollisionCostConfig
+{
+  /** @brief If true, a collision cost term will be added to the problem. Default: true*/
+  bool check = true;
```
Levi-Armstrong: I recommend switching to enabled.

schornakj:
> > In any case I think it would be better to make a separate PR.
> 
> It may be good to do it in this PR because the function and struct will need to change again to support the change in trajopt.

I think that's OK, since the concepts of enabling dual collision terms and having a config struct for collision settings are functionally distinct from the adjustments to the low-level mechanics of collision checking added in that Trajopt branch.

Also, I'm still working out how to expose the new Trajopt safety_buffer_margin variable in a clean way, so I don't think this PR needs to wait for those changes to be finalized.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_collision_config.h
```diff
@@ -1,3 +1,28 @@
+/**
+ * @file trajopt_collision_config_base.h
```
Levi-Armstrong: Should be trajopt_collision_config.h

Levi-Armstrong:
One last thing the python `trajopt_planner_default_config.i` needs to be updated.

Levi-Armstrong:
@schornakj Looks good. I will wait for CI to pass before merging.

Levi-Armstrong:
@schornakj I believe if you rebase on the master CI should pass.

Levi-Armstrong:
@schornakj It looks like some unit tests were not updated so Travis CI is failing.

PR #213, opened 2020-01-23
Add double precision flag to tesseract_collision target compile definitions

schornakj:
Supports [this `tesseract_ext` issue](https://github.com/ros-industrial-consortium/tesseract_ext/issues/10) and [this PR](https://github.com/ros-industrial-consortium/tesseract_ext/pull/11). Workaround for [a bullet3 issue](https://github.com/bulletphysics/bullet3/issues/2582) where it doesn't correctly export some compiler definitions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,13 @@ env:
     - ROSINSTALL_FILENAME=dependencies.rosinstall
     - ROS_REPO=ros
     - NOT_TEST_INSTALL=true
-    - CATKIN_CONFIG="--make-args CTEST_OUTPUT_ON_FAILURE=TRUE CTEST_PARALLEL_LEVEL=2"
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
@schornakj It looks like CI is running out of resources now with the switch to Double. For melodic can you reduce the jobs to 4 like what is done for melodic and then reduce melodic to 2?

schornakj:
@Levi-Armstrong Thanks for that quick fix

Levi-Armstrong:
@schornakj Can you verify that the tesseract_ros_examples run with this change?

schornakj:
@Levi-Armstrong The demos in tesseract_ros_examples all complete successfully. The car door example produces a path where the headrest of the car seat is slightly in collision with the top of the car's frame, but I don't know if that's a pre-existing issue. The pick-and-place example works completely as expected.

Issue #207, opened 2020-01-20
Move Tesseract_ros to its own package


mpowelson:
I know this has been discussed, and I believe there are some roadblocks to this. What are they and how do we plan on resolving them?

schornakj:
I think that the packages are fully separated now. When I build Tesseract for ROS2 I just put a `COLCON_IGNORE` file in `tesseract/tesseract_ros` and Colcon uses `tesseract_ros2` instead, which indicates that there aren't any dependencies between `tesseract_ros` and the rest of the core Tesseract packages.

Levi-Armstrong:
I came across this repository showing an example having ros1 and ros2 packages in the same repository. I created a [PR](https://github.com/romainreignier/share_ros1_ros2_lib_demo/pull/1) showing how to merge them into a single package so the package name does not change between versions of ros. I wanted to get your thoughts on using this approach to merge in tesseract_ros2 into this repository? @schornakj @mpowelson @marip8 @jrgnicho @jdlangs

jrgnicho:
@Levi-Armstrong  you proposed there looks fine to me, although just adding a COLCON_IGNORE file would be much simpler and it would still build under ROS1 as long as tesseract_ros2 remains in its own repo

gavanderhoorn:
Not using the ugly conditionals in package manifests would also be much less .. well .. *ugly* ;)

Levi-Armstrong:
Currently there are three options.

- Option 1:
  - Core Libraries in its own repository
  - ROS1 Libraries in its own repository
  - ROS2 Libraries in its own repository
- Option 2:
  - All in the same repository but ROS1 and ROS2 packages must have different names. Example tesseract_monitoring (ROS1) then tesseract_monitoring2 (ROS2). Also the CMakeLists.txt must check for Catkin and if not found do not build shown in the [original example](https://github.com/romainreignier/share_ros1_ros2_lib_demo).
- Option 3:
  - All in the same repository where ROS packages can have the same names between ros version. But it requires a conditional package.xml and different folder structure shown the [modified example](https://github.com/romainreignier/share_ros1_ros2_lib_demo/pull/1).

I would like to get everyone's opinion about which one they prefer. @gavanderhoorn  @schornakj @mpowelson @marip8 @jrgnicho @jdlangs

mpowelson:
I think my preference would be first Option 1. Then Option 3.

[...]

PR #218, opened 2020-01-28
Adds colcon ignore in order to ignore ROS1 packages from colcon build in ROS2

jrgnicho:


jrgnicho:
@Levi-Armstrong  it errors out whenever it finds any catkin related commands in the CMakeLists.txt.  This change would allow to build the pure cmake packages in a colcon ROS2 workspace.

Levi-Armstrong:
I believe the reason it errors out using colcon is because the package.xml has not been updated to format 2 and it needs the [<build_type>catkin</build_type>](https://www.ros.org/reps/rep-0149.html#build-type-multiple) added.

jrgnicho:
It would still fail in a ROS2 workspace because the ROS1 libraries aren't part of that environment

jrgnicho:
Furthermore, in a colcon ROS2 ws that also uses tesseract_ros2 there will be conflicting packages because they're named the same as those in tesseract/tesseract_ros

Levi-Armstrong:
Lets discuss further on issue #207.

mpowelson:
@Levi-Armstrong Until @207 is implemented, can we just COLCON_IGNORE tesseract_ros and tesseract_python? Having these in here is making CI problematic.

mpowelson:
@Levi-Armstrong ping

Levi-Armstrong:
@jrgnicho Can you rebase? I think it should fix CI to get this merged.

mpowelson:
I think we'll have to ignore tesseract_python as well since it uses tesseract_rosutils

jrgnicho:
@Levi-Armstrong I just rebased

Levi-Armstrong:
@jrgnicho Something is still not happy.

mpowelson:
I don't think this is going to work. I thought that catkin only obeyed CATKIN_IGNORE, but colcon ignored all. That does not appear to be the case. See [HERE](https://github.com/ros-infrastructure/catkin_pkg/blob/35137d5522fc4b4140bf24d4d35373a1ff74e4e9/src/catkin_pkg/packages.py#L47) linked from [THIS issue](https://answers.ros.org/question/306624/ignore-package-in-colcon-but-not-catkin/) which I see Jorge is a part of. Perhaps we should just add a check to the package.xml/cmake like we do in yak_ros? 

It looks like it doesn't like that we are trying to run tests on tesseract_ros_examples using PKGS_DOWNSTREAM.

mpowelson:
This has been replaced by #241

PR #219, opened 2020-01-30
Allow to merge a SceneGraph into another one

### LINKED ISSUE #142: How to add multiple "robots" to the same Environment?
I've been experimenting a bit more with Tesseract / TrajOpt and it seems to give good results :+1: 

However I would like to do the following:
- Create an Environment containing my robot via `Tesseract::init(urdf, srdf, locator)`
- Add collision shapes and other objects independently

So far I've been constructing them by hand (create Link, create Joint, then call `addLink`), but I would ideally load them from a URDF file. However a quick look at the documentation tells me that:
- `parseURDF` either returns a `scene_graph` or appends to a `scene_graph`
- `Environment` only has a `const` getter for its scene graph

I believe that I could get away with any of the following:
1. Merging my URDFs using xacro
2. Create a `scene_graph`, parse the various URDFs into it, and finally use `Tesseract::init`
3. By hand traverse each loaded model and add their Link / Joints to the scene
4. Each time a collision shape is to be added, update a stored `scene_graph` and call again `Environment::init`

Options 1. and 2. are suboptimal because they don't allow for online modification of the environment but 3. means writing quite a lot of code and 4. feels like a big hack.

Ideally I would like to reach something akin to `Environment::addRobot(parseURDF(urdf))` that would allow me to add "robots" (they are static for now) to the environment.

haudren:
This is a proposal to fix #142 

Basically introduce a new API to add a SceneGraph to another one, possibly with a prefix. This allows to `parseURDF` once, and add the same robot to the same scene multiple times with different names. Unfortunately, I needed to create prefixed copy operators for links and joints. As long as `name_` is a `const` member, I couldn't figure out a very good way to write them any other way.

The copying itself is done as breadth-first search over all links / joints, copying it into a prefixed version and adding to current Environment.

I also created a test that demonstrates usage and checks the following:
- Adding an empty scene graph always succeeds (nothing to do)
- Adding multiple times the same scene graph fails without prefix, succeeds with it

### DIFF
--- a/tesseract/tesseract_environment/src/core/environment.cpp
+++ b/tesseract/tesseract_environment/src/core/environment.cpp
@@ -28,6 +28,8 @@
 #include <tesseract_environment/core/utils.h>
 #include <tesseract_collision/core/common.h>
 
+#include <queue>
+
 namespace tesseract_environment
 {
 void Environment::setState(const std::unordered_map<std::string, double>& joints)
@@ -592,4 +594,62 @@ bool Environment::removeLinkHelper(const std::string& name)
   return true;
 }
 
+bool Environment::addSceneGraph(const tesseract_scene_graph::SceneGraph& scene_graph, const std::string& prefix)
+{
+  // Connect root of subgraph to graph
+  tesseract_scene_graph::Joint::Ptr root_joint =
+      std::make_shared<tesseract_scene_graph::Joint>(scene_graph.getName() + "_joint");
+  root_joint->type = tesseract_scene_graph::JointType::FIXED;
+  root_joint->parent_link_name = getRootLinkName();
+  root_joint->child_link_name = scene_graph.getRoot();
+  root_joint->parent_to_joint_origin_transform = Eigen::Isometry3d::Identity();
+
+  return addSceneGraph(scene_graph, root_joint, prefix);
+}
+
+bool Environment::addSceneGraph(const tesseract_scene_graph::SceneGraph& scene_graph,
+                                tesseract_scene_graph::Joint::ConstPtr root_joint,
+                                const std::string& prefix)
+{
+  auto link = scene_graph.getLink(scene_graph.getRoot());
+  if (!link)
+  {
+    return true;
+  }
+  auto new_link = std::make_shared<tesseract_scene_graph::Link>(link->prefix(prefix));
+
+  auto new_joint = std::make_shared<tesseract_scene_graph::Joint>(root_joint->prefix(prefix));
+  // Preserve reference to common ancestor
+  new_joint->parent_link_name = root_joint->parent_link_name;
+
+  bool res = addLink(*new_link, *new_joint);
+  if (!res)
+  {
+    CONSOLE_BRIDGE_logError("Could not add the root joint");
+    return false;
+  }
+
+  std::queue<std::string> work_links;
+  work_links.push(link->getName());
+  while (!work_links.empty())
+  {
+    auto joints = scene_graph.getOutboundJoints(work_links.front());
+    work_links.pop();
+    for (const auto& joint : joints)
+    {
+      auto new_joint = std::make_shared<tesseract_scene_graph::Joint>(joint->prefix(prefix));
+      link = scene_graph.getLink(joint->child_link_name);
+      auto new_link = std::make_shared<tesseract_scene_graph::Link>(link->prefix(prefix));
+      res = addLink(*new_link, *new_joint);
+      if (!res)
+      {
+        CONSOLE_BRIDGE_logError("Could not add link (%s)", new_link->getName().c_str());
+        return false;
+      }
+      work_links.push(link->getName());
+    }
+  }
+  environmentChanged();
+  return res;
+}
 }  // namespace tesseract_environment
--- a/tesseract/tesseract_environment/src/core/environment.cpp
+++ b/tesseract/tesseract_environment/src/core/environment.cpp
@@ -28,7 +28,9 @@
 #include <tesseract_environment/core/utils.h>
 #include <tesseract_collision/core/common.h>
 
+TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <queue>
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_environment
 {
--- a/tesseract/tesseract_environment/src/core/environment.cpp
+++ b/tesseract/tesseract_environment/src/core/environment.cpp
@@ -78,7 +78,7 @@ EnvState::Ptr Environment::getState(const std::vector<std::string>& joint_names,
   return state;
 }
 
-bool Environment::addLink(const tesseract_scene_graph::Link& link)
+bool Environment::addLink(tesseract_scene_graph::Link link)
 {
   std::string joint_name = "joint_" + link.getName();
   tesseract_scene_graph::Joint joint(joint_name);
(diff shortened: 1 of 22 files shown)


tesseract/tesseract_scene_graph/include/tesseract_scene_graph/joint.h
```diff
@@ -256,6 +256,24 @@ class Joint
     this->type = JointType::UNKNOWN;
   }
 
+  /* Create a shallow copy of current joint, with prefix prepended to its name, child link name and parent link name.
+   * Shallow means that underlying properties, such as dynamics, limits... are shared with the original object.*/
+  Joint prefix(const std::string& prefix) const
+  {
```
Levi-Armstrong: With this being a shallow means that when one changes they all change, is this the desired affect? Is there a down side with making this a deep copy? If this is needed there is some more code throughout that would need to be tested and updated like rviz plugins,etc. because I am not sure if you update one it would update them all.

haudren: Yes that's the idea, every prefixed version of the Joint will share (via `shared_ptr`) the same attributes. I implemented it this way because:
- It was easier
- I did not want to copy Link geometry, and applied the same logic to Joint
But I understand your point... I can make it a deep copy easily, by copying the underlying objects instead of the pointers.

Levi-Armstrong: I agree with your point. The geometry has a clone method which does an efficient copy. The geometry data are stored in shared pointer where it copies the shared pointer on clone. This is ok because there is no way to change the geometry data from the classes interface.

It would be good to add a clone method the objects in link.h and joint.h. Also would be good to delete the copy and assignment constructor.

haudren: Please let me know what you would prefer, I'll update here.

Levi-Armstrong: I think it would be good to implement the clone method and delete the copy and assignment constructor for the types in link.h and joint.h. I think the only class that does not need a clone is the material.

Levi-Armstrong: Then your prefix function for link and joint would become the clone function that takes a name.

haudren:
@Levi-Armstrong : Sorry for the delay, I could only work on this today. I pushed some changes that delete the copy operators, and unfortunately the consequences are pretty far-ranging. I'll let you have a look and tell me if this is worth pursuing before I move on to fixing the ROS and Python side.

Levi-Armstrong:
I will take a look.

Levi-Armstrong:
@haudren I like it. It also shows that the scene graph takes ownership of the links and that they can only be changed through the scene graph interface. @mpowelson @johnwason Do you see any issues with this change?

haudren:
Ok so I've pushed another commit to that branch that fixes the remaining issues in the ROS side of things. Now the SWIG bindings seem to complain quite a bit, I guess we need some sort of typemap / wrapper for the environment functions:
```
addLinkWrapper(const Link& link)
{
    auto new_link = link.clone(link.getName());
    addLink(std::move(new_link));
}
```
This obviously introduces a copy but I don't think there is a better way from Python code. @johnwason do you have any idea how to introduce that functionality in the bindings?

haudren:
Ok I found a way using SWIG's `%extend` directive. All python tests are still passing (on my machine, waiting for CI) :tada:

Levi-Armstrong:
Thank you. I should be able to review this by Friday.

tesseract_python/tesseract_python/swig/tesseract_environment/core/environment.i
```diff
@@ -150,5 +145,28 @@ public:
   registerContinuousContactManager(const std::string& name,
                                    tesseract_collision::ContinuousContactManagerFactory::CreateMethod create_function);
 
+  %extend
```
Levi-Armstrong: I am not very familiar with SWIG and writting python wrappers but does python not support the ability to move objects?

Levi-Armstrong: If not, is there a reason not to use `const Joint& joint` for these function, it would be more flexible to the python developer so they do not have to create a shared pointer?

johnwason: `Joint` is always a shared pointer in Python so it shouldn't make much difference. The only wrinkle is that shared pointers can be null so that needs to be checked.

tesseract_python/tesseract_python/swig/tesseract_scene_graph/graph.i
```diff
@@ -59,7 +59,7 @@ bool setRoot(const std::string& name);
 
 const std::string& getRoot() const;
 
-bool addLink(Link link);
+bool addLink(Link::Ptr link);
```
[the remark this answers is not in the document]
haudren: Ok I'll protect it, but I might need to make the URDF parser a friend of the scene graph to use it... I'll see how it goes.

tesseract_python/tesseract_python/swig/tesseract_environment/core/environment.i
haudren: Indeed @Levi-Armstrong : in Python everything is a pointer anyways, and there is no such thing as "moving"... So I settled for this as a compromise.

tesseract/tesseract_environment/src/core/environment.cpp
```diff
@@ -28,6 +28,8 @@
 #include <tesseract_environment/core/utils.h>
 #include <tesseract_collision/core/common.h>
 
+#include <queue>
+
```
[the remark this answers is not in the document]
haudren: Do we really need it? On my system at least `queue` emits no warnings, and stdlib is included with `-isystem` anyways :thinking:

haudren:
I believe I have addressed all the comments, please let me know what you think @Levi-Armstrong @johnwason .

haudren:
I found a bug in one of the ROS tests: it was segfaulting because I was passing in an empty string to `getInboundJoints`. It seems to me that `moveLink` is missing a precondition check, and then calls `getInboundJoints`, which itself attempts to dereference a null pointer (that should normally point to a vertex).

tesseract/tesseract_environment/src/core/environment.cpp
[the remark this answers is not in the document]
Levi-Armstrong: I have ran into situations where this is needed. I believe this is needed when including header only components, so I have just been enclosing system level headers by default.

tesseract_python/tesseract_python/swig/tesseract_environment/core/environment.i
Levi-Armstrong: Ah, Can you tell I don't use python much :). Looks good.

Levi-Armstrong:
> I found a bug in one of the ROS tests: it was segfaulting because I was passing in an empty string to getInboundJoints. It seems to me that moveLink is missing a precondition check, and then calls getInboundJoints, which itself attempts to dereference a null pointer (that should normally point to a vertex).

Did you address this in this PR?

haudren:
@Levi-Armstrong : The bug itself is addressed, all tests are passing. However, we could, in a future PR change the behaviour of `getInboundJoints` to avoid segfaulting when passed a non-existent link name.

Levi-Armstrong:
@haudren Thank you for your contribution.

haudren:
@Levi-Armstrong : You're welcome, using this version of Tesseract and TrajOpt is much more convenient than the original OpenRAVE implementation :sweat_smile:

PR #222, opened 2020-02-01
Use BMPD solver for puzzle piece examples

Levi-Armstrong:
When the default solver was switched to OSQP the test takes longer to run causing the test to fail. I set the two puzzle piece test to use the BMPD solver which solve the issue.


Levi-Armstrong:
@schornakj can you review?

arocchi:
This then makes the two puzzle piece test a good test to stress-test the new solvers. Would be curious to see `ipopt` on this.

mpowelson:
@Levi-Armstrong Should we open an issue for this? Obviously this PR just hides the fact that OSQP is not working well for this problem.

Levi-Armstrong:
Yea, Currently CI builds everything in Debug which may be why OSQP takes longer. It would be good to benchmark the puzzle piece when built in release. It is able to solve the problem just times out during the unit test run in CI. I will create an issue.

Levi-Armstrong:
@arocchi Do you know how much work it would be to integrate ipopt into trajopt solver interface?

Levi-Armstrong:
Opened issue #225

arocchi:
> @arocchi Do you know how much work it would be to integrate ipopt into trajopt solver interface?

@Levi-Armstrong I would guess a couple of days of work (~16 man hours)

PR #223, opened 2020-02-01
Modify OMPL Planner and Config to support multiple planner types

Levi-Armstrong:
@schornakj I modified the ompl planner so you could pass in multiple ompl planner type which get solved in parallel.


Levi-Armstrong:
@schornakj Can you review?

mpowelson:
@Levi-Armstrong This looks pretty nice. However, the threading seems clunky. What would you think about either creating another type or adding things like num_threads, optimize, max_solutions, simplify, etc to OMPLSettings? 

The reason I mention it is that it seems natural that the planning time, manipulator, etc are forced to be the same for all of the planners. However, it seems like some of the others might be beneficial to set on a planner by planner basis. You've already got a tuple of OMPLSettings, so you could add them to that or you could create another type to contain some of those parameters if you want to divide them up by what is passed to OMPL and what is configuring Tesseract

Levi-Armstrong:
So currently we leveraging OMPL Parallel Planner object which manages the planning time, max_solutions, and number of threads. It does not provide functionality at the individual planner. Although I think this would be beneficial functionality, but it would require writing our own parallel planner to manage the settings at an individual thread/planner. Lets add it to the action list on the todo issue. Also I image at some point we will create a parallel planner similar to the ompl parallel planner that takes tesseract planner.

In the case of threads. Should the num_threads be changed to be a vector where you set the number of threads desired per planner? This would allow for you define say RRTConnect and SBL and then define say 3 threads/planners for RRTConnect and 4 threads/planners for SBL.

mpowelson:
I'd say it's fine the way you have it. There's no reason why you can't add the same planner more than once is there?

Levi-Armstrong:
>I'd say it's fine the way you have it. There's no reason why you can't add the same planner more than once is there?

Nope. You will just have to configure each planners OMPLSetting which is not an issue.

marip8:
I think this might be a good time to revisit the design of this planning class. It seems a unnecessary to template this class because we do not actually need to know anything about the specific implementations of the OMPL planner or its configuration. We use the templates simply to overload the configuration of the planner. Instead I think we should remove the templates and only store a configurator class that can create an OMPL planner base class with the desired settings. 

As far as variadic templating goes, I don't have much experience with it, but it seems like a complicated solution to the issue of solving a motion planning problem with multiple planners. I would suggest creating a new OMPL planning class (or an overload to the main OMPL planning class) that holds a vector of OMPL planner configurators.

Here is my initial thought for an improved OMPL motion planner that could easily be extended to work with multiple planners

```
// Refactored OMPL configurator base class that is not templated and can create an instance of an OMPL planner base class
// This replaces the OMPLSettings
struct OMPLPlannerConfigurator
{
  using Ptr = std::shared_ptr<OMPLPlannerConfigurator>;
  virtual ~OMPLPlannerConfigurator() = default;
  virtual ompl::base::PlannerPtr create(ompl::base::SpaceInformationPtr si) const = 0;
};

// Example specific planner configurator
struct SBLConfigurator : OMPLPlannerConfigurator
{
  double range = 0.0;
  virtual ompl::base::PlannerPtr create(ompl::base::SpaceInformationPtr si) const override
  {
    auto planner = std::make_shared<ompl::geometric::SBL>(si);
    planner->setRange(range);
    return planner;
  }
};

// Refactored OMPL motion planning class without template
class OMPLMotionPlanner : public MotionPlanner
{
public:
  OMPLMotionPlanner(OMPLPlannerConfigurator::Ptr configurator);
  // OMPLMotionPlanner(std::vector<OMPLPlannerConfigurator::Ptr> configurators);
  
  solve() 
  {
    ...
    ompl::base::PlannerPtr planner = planner_configurator_.create(space_info);
    planner->solve(...);

    //for(const auto& pc : planner_configurators_)
    //{
      //ompl::base::PlannerPtr planner = pc.create(space_info);
      //parallel_planner_.addPlanner(*planner);
    //}
    ...
  }

private:
  OMPLPlannerConfigurator::Ptr planner_configurator_;
  // std::vector<OMPLPlannerConfigurator::Ptr> planner_configurators_;
}
};
```

Levi-Armstrong:
@marip8 Based on your recommendation I have restructured the ompl planner removing the use of templates. Do you mind reviewing?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/config/ompl_planner_config.h
```diff
@@ -32,27 +32,27 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <ompl/base/OptimizationObjective.h>
 #include <ompl/tools/multiplan/ParallelPlan.h>
 #include <ompl/base/objectives/PathLengthOptimizationObjective.h>
+#include <tuple>
```
marip8: Probably unneeded now

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/impl/config/ompl_planner_freespace_config.hpp
```diff
@@ -43,15 +43,15 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 namespace tesseract_motion_planners
 {
 /** @brief Construct a basic planner */
-template <typename PlannerType>
-OMPLPlannerFreespaceConfig<PlannerType>::OMPLPlannerFreespaceConfig(tesseract::Tesseract::ConstPtr tesseract_,
-                                                                    std::string manipulator_)
-  : OMPLPlannerConfig<PlannerType>(std::move(tesseract_), std::move(manipulator_))
+template <typename... PlannerType>
```
marip8: Is this a remnant of the previous effort? I don't think this class is templated any more, and this file should probably be moved to a .cpp source file

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/ompl_planner_configurator.h
```diff
@@ -0,0 +1,234 @@
...
+ * @brief Tesseract OMPL planner configurator.
+ *
+ * If a settings class does not exist for a planner available
+ * in ompl you may simply create your own that has an apply
+ * method that takes the specific planner you would like to use
+ * and construct the OMPLFreespacePlanner with the desired planner
+ * and newly created config class and everything should work.
+ *
+ * @author Levi Armstrong
+ * @date April 18, 2018
```
marip8: Update date

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_motion_planner.cpp
```diff
@@ -1,5 +1,5 @@
 /**
- * @file ompl_motion_planner.cpp
+ * @file ompl_motion_planner.hpp
```
marip8: No change necessary here

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_planner_configurator.cpp
```diff
@@ -0,0 +1,154 @@
+/**
+ * @file ompl_planner_configurator.cpp
+ * @brief Tesseract OMPL planner conversions
+ *
+ * @author Jonathan Meyer
```
marip8: Change name and date

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_planner_configurator.cpp
```diff
@@ -0,0 +1,154 @@
+/**
+ * @file ompl_planner_configurator.cpp
+ * @brief Tesseract OMPL planner conversions
```
marip8: conversions -> configurators

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp
```diff
@@ -116,32 +116,170 @@ static void addBox(tesseract_environment::Environment& env)
...
-                                         ompl::geometric::KPIECE1,
-                                         // ompl::geometric::RRT,
-                                         // ompl::geometric::RRTstar,
-                                         // ompl::geometric::SPARS,
-                                         // ompl::geometric::TRRT,
-                                         ompl::geometric::RRTConnect>;
-
-TYPED_TEST_CASE(OMPLTestFixture, Implementations);
-
-TYPED_TEST(OMPLTestFixture, OMPLFreespacePlannerUnit)
```
marip8: You could probably retain the template test structure that currently exists and replace the implementations with the new configurator classes. This would cut down on the amount of repeat code you need to create here

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_trajopt_planner_test.cpp
```diff
@@ -119,32 +118,25 @@ static void addBox(tesseract_environment::Environment& env)
...
-                                         ompl::geometric::KPIECE1,
-                                         // ompl::geometric::RRT,
-                                         // ompl::geometric::RRTstar,
-                                         // ompl::geometric::SPARS,
-                                         // ompl::geometric::TRRT,
-                                         ompl::geometric::RRTConnect>;
-
-TYPED_TEST_CASE(OMPLTrajOptTestFixture, Implementations);
-
-TYPED_TEST(OMPLTrajOptTestFixture, OMPLTrajOptFreespacePlannerUnit)  // NOLINT
```
marip8: You could probably retain the template test structure here as well and just replace the implementations with the new planner configurator classes

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/impl/config/ompl_planner_freespace_config.hpp
Levi-Armstrong: Yep Fixed

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/ompl/ompl_planner_configurator.h
Levi-Armstrong: Done

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_motion_planner.cpp
Levi-Armstrong: Fixed

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_planner_configurator.cpp
Levi-Armstrong: Fixed

tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_planner_configurator.cpp
Levi-Armstrong: fixed

Levi-Armstrong:
@marip8 I believe I have made all the requested changes.

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp
Levi-Armstrong: Switched.

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_trajopt_planner_test.cpp
Levi-Armstrong: Switched

schornakj:
@Levi-Armstrong I finally had a chance to try this out. I like the decision to make a separate configurator object with planner-specific settings.

I noticed that the `num_threads` field in the `OMPLPlannerConfig` class doesn't have the same effect on parallel planning as it did previously, and if I want N planners running in parallel I have to construct the planner config with a vector of N planner configurators. I think that to reduce confusion we should either a) deprecate the `num_threads` field, or b) if a single planner configurator is passed in check `num_threads` and replicate the configurator that many times to emulate the current behavior.

marip8:
@schornakj I think my vote would be to deprecate the `num_threads` variable in favor of making parallel plans for only the planners in the vector of configurators (and adding supporting documentation). I think the `num_threads` variable is ambiguous if more than one planner exists in the vector of configurators

Levi-Armstrong:
> @schornakj I think my vote would be to deprecate the `num_threads` variable in favor of making parallel plans for only the planners in the vector of configurators (and adding supporting documentation). I think the `num_threads` variable is ambiguous if more than one planner exists in the vector of configurators

This works for me. I will make the change.

Levi-Armstrong:
> Looks good to me. It looks like your unit tests create two planner configurators: one for the template type configurator and one SBL configurator. Is this the desired behavior?

@marip8 This was an error. I fixed it.

PR #224, opened 2020-02-01
Add OMPL Constraint Support

Levi-Armstrong:
This add  the ability to add a constraint to the ompl planner. Next to create a ompl constraint that takes a trajopt problem for projecting the sampled state on to the constraint manifold.

Update:
With constraints it solves but fails post collision check. Need to check if it passes post check prior to interpolation.


Levi-Armstrong:
replaced by #233

PR #227, opened 2020-02-06
Add eigen to package.xml and alphabetize the entries.

mpowelson:


tesseract/tesseract_collision/package.xml
```diff
@@ -9,7 +9,9 @@
 
   <depend>tesseract_geometry</depend>
   <depend>tesseract_common</depend>
+
   <depend>bullet3</depend>
+  <depend>eigen</depend>
```
gavanderhoorn: Seeing as Eigen is a header-only library, shouldn't this be a `build_depend` and `build_export_depend` instead?

mpowelson: I suppose. What does that practically get you? I can replace them with build_export_depend if you want, but I'd rather not have to go through all of the packages and see how it is being used. Currently I just added it if there was a find_package in the cmake.

gavanderhoorn: More fine grained dependency management.

Eigen is not a large library, but it's not very nice to install packages which are build dependencies only on systems that will never be compiling something.

`depend` == `union(build_depend, build_export_depend, exec_depend)`.

And on systems where Eigen is not actually distributed as a package (and those do exist), this saves users from having to `skip-key` yet another key a maintainer has just blanketly added to the `depend` set.

gavanderhoorn:
Thanks for changing it, but `eigen` needs to be both a `build_depend` *and* a `build_export_depend` (ie: it's needed for the current build, and it's needed by packages that use your packages, when they are built).

PR #228, opened 2020-02-07
Adjust Departure Generator

DavidMerzJr:
If a path has all poses in all raster strokes with x and y pointing in the same direction (not flipping back and forth each stroke), a departure with x/y components will 'backtrack' on some strokes and continue 'forward' on others.  This commit changes that so that 'positive' x/y values always move away from the penultimate pose in a stroke, and 'negative' x/y values always move back towards the penultimate pose in a stroke.

If the general method used is agreeable, I could add a second commit doing the same for the axial approach generator.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_process_planners/include/tesseract_process_planners/generators/extension_departure_generator.h
+++ b/tesseract/tesseract_planning/tesseract_process_planners/include/tesseract_process_planners/generators/extension_departure_generator.h
@@ -1,3 +1,29 @@
+/**
+ * @file extension_departure_generator.cpp
+ * @brief Tesseract process departure implementation
+ *
+ * @author David Merz, Jr.
+ * @date February 17, 2020
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2020, Southwest Research Institute
+ *
+ * @par License
+ * Software License Agreement (Apache License)
+ * @par
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


tesseract/tesseract_planning/tesseract_process_planners/include/tesseract_process_planners/generators/extension_departure_generator.h
```diff
@@ -0,0 +1,108 @@
+#ifndef TESSERACT_PLANNING_EXTENSION_DEPARTURE_GENERATOR_H
```
Levi-Armstrong: Missing license.

Levi-Armstrong:
@DavidMerzJr I missunderstood your question; I would us the license format you used in the extension generator in all of the other files.

tesseract/tesseract_planning/tesseract_process_planners/include/tesseract_process_planners/generators/extension_departure_generator.h
DavidMerzJr: Added.

DavidMerzJr:
@Levi-Armstrong I updated all the license statements I added to include the Doxygen macros and my best guess at a short description.  I listed you as the author on them.

tesseract/tesseract_planning/tesseract_process_planners/include/tesseract_process_planners/generators/extension_departure_generator.h
Levi-Armstrong: Thanks

PR #230, opened 2020-02-13
Planner fixes

mpowelson:
Allows the use of JOINT_INTERPOLATED in Trajopt_planner.

Only add waypoints to the fixed_waypoint list if they are critical. Currently it treats any joint position waypoint as fixed which may not be the case depending on the coefficient

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
@@ -215,7 +215,7 @@ void TrajOptPlannerDefaultConfig::addWaypoints(trajopt::ProblemConstructionInfo&
      * that are incapable of changing (i.e. joint positions). Therefore, the first and last indices of these
      * costs (which equal 0 and num_steps-1 by default) should be changed to exclude those states
      */
-    if (target_waypoints[ind]->getType() == WaypointType::JOINT_WAYPOINT)
+    if (target_waypoints[ind]->getType() == WaypointType::JOINT_WAYPOINT && target_waypoints[ind]->isCritical())
     {
       fixed_steps.push_back(static_cast<int>(ind));
     }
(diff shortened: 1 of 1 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/config/trajopt_planner_default_config.cpp
```diff
@@ -179,6 +179,16 @@ bool TrajOptPlannerDefaultConfig::addInitTrajectory(trajopt::ProblemConstruction
       }
     }
   }
+  if (init_type == trajopt::InitInfo::JOINT_INTERPOLATED)
```
Levi-Armstrong: I thought this is handled in trajopt which will interpolate from the start to the end location; is that not the case?

mpowelson: It interpolates from the current state to the joint position specified in pci.init_info.data. 

https://github.com/ros-industrial-consortium/trajopt_ros/blob/e37cba610deb32cbb3136cc32cf5e87ca2ee6c78/trajopt/src/problem_description.cpp#L324

Levi-Armstrong: Now I remember. Any thought on how to improve this?

- Case 1:  If the start and end state are Joint States then joint interpolate between and fill out the seed trajectory and change the init type passed to trajopt.
- Case 2: If the only the end state is a joint state then use the current robot state and the end state to fill out the seed trajectory and change the init type passed to trajopt.
  - Another option would be to solve IK for the start the start state.
- Case 3: If only the start state is a joint state, then solve IK for the end state and then interpolate and fill out the seed trajectory and change init type passed to trajopt.
- Case 4: If it is not the first case we can just error if this init type was selected and require the user to provide a seed.

mpowelson: So the reason I am using this is because I am adding a bunch of "freespace" waypoints to the plan rather than using the freespace planner. I add a bunch of joint waypoints and turn all of the coefficients off except for the first and last. I find this easier as if I want to do multiple chained moves I can do it all in one by just changing the coefficients of waypoints in the middle. Also I just find the config structure confusing in general, and this is similar.

With that said, joint interpolated probably doesn't make much sense in the default config if you are not doing this. So I think I might recommend that we keep the Tesseract planner side transparent unless we are going to add something like a freespace waypoint officially (in which case we could make a more intelligent "auto" initializer). If I am going to be hacking the planner, I'd rather know what the results will be coming out of the other side than have it change what I gave it under the hood. That said, an easy way to do joint interpolated b/n a joint position and a cartesian position would be welcome (and perhaps is already in the freespace config. I'm not sure).

PR #233, opened 2020-02-17
Add ability to add constraints to ompl planner

Levi-Armstrong:


tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp
```diff
@@ -304,6 +303,91 @@ TEST(OMPLMultiPlanner, OMPLMultiPlannerUnit)  // NOLINT
...
+  ~GlassUprightConstraint() override = default;
+
+  void function(const Eigen::Ref<const Eigen::VectorXd>& x, Eigen::Ref<Eigen::VectorXd> out) const override
+  {
+    Eigen::Isometry3d pose;
+    fwd_kin_->calcFwdKin(pose, x);
+
+    Eigen::Vector3d z_axis = pose.matrix().col(2).template head<3>().normalized();
+
+    out[0] = z_axis.dot(normal_);
```
marip8: I think your function here should actually set `out[0] = 1.0 - z_axis.dot(normal_)` such when the vectors are aligned (dot product = 1.0), your return value is zero

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp
```diff
@@ -304,6 +303,91 @@ TEST(OMPLMultiPlanner, OMPLMultiPlannerUnit)  // NOLINT
...
+
+  tesseract_motion_planners::PlannerResponse ompl_planning_response;
+  tesseract_common::StatusCode status = ompl_planner.solve(ompl_planning_response);
+
+  if (!status)
+  {
+    CONSOLE_BRIDGE_logError("CI Error: %s", status.message().c_str());
+  }
+  EXPECT_TRUE(status);
+  EXPECT_TRUE(ompl_planning_response.joint_trajectory.trajectory.rows() >= ompl_config->n_output_states);
```
marip8: It may also be worth looping through the trajectory, solving FK at each waypoint, and verifying the constraint was met

Levi-Armstrong: Done.

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_planner_tests.cpp
Levi-Armstrong: Fixed.

Levi-Armstrong:
@mpowelson This is ready to review. Although the constrained planning in OMPL does not seem to work for the example case when you have collision objects. I have speculation on why it is not working but we will keep it around and continue to investigate why it does not perform well. May need to also reach out to the OMPL developers and see if they have ever used it for robotics application and if they had challenges.

tesseract/tesseract_planning/tesseract_motion_planners/test/ompl_constrained_planner_tests.cpp
```diff
@@ -0,0 +1,219 @@
...
+  ~GlassUprightConstraint() override = default;
+
+  void function(const Eigen::Ref<const Eigen::VectorXd>& x, Eigen::Ref<Eigen::VectorXd> out) const override
+  {
+    Eigen::Isometry3d pose;
+    fwd_kin_->calcFwdKin(pose, x);
+
+    Eigen::Vector3d z_axis = pose.matrix().col(2).template head<3>().normalized();
+
+    out[0] = std::atan2(z_axis.cross(normal_).norm(), z_axis.dot(normal_));
```
marip8: Can you clarify your method for calculating the angle error here? I don't see how the magnitude of the cross product is relevant in this case because geometrically  the area of the parallelogram between the two vectors. Also this would not be well-behaved in the case that the vectors are orthogonal (the dot product, i.e. the denominator, is zero)

Levi-Armstrong: github_pic

Above is a picture of the proof. One assumption I made was the normal was of unit length and probable should check if the input normal vector is unit length and if not normalize it. 

> Also this would not be well-behaved in the case that the vectors are orthogonal (the dot product, i.e. the denominator, is zero)

This is the reason for using `atan2` which checks for of these issues and returns the absolute angle (Not the acute angle) betweep -pi and pi.

marip8: Okay, I understand now. I personally would still suggest using `1.0 - z_axis.dot(normal_)` (or `z_axis.norm() * normal_.norm() - z_axis.dot(normal_)` if you don't want to assume unit vectors) to reduce complexity/understand-ability, but it looks to produce the same result

Levi-Armstrong: I agree. This is for testing because it is easier to understand since the units are radians.

Levi-Armstrong:
@marip8 Do you have a kinetic install to test why it is failing?

Levi-Armstrong:
Moved to #249

PR #235, opened 2020-02-21
Add fields to OptimizationObjectiveAllocator for Tesseract environment and FK manager

schornakj:
This allows the use of custom optimization objective functions that evaluate costs based on the Tesseract collision environment.

Is there a more flexible way to do this so the scope of what can be used in an objective function isn't limited to the arguments of the function allocator defined here?

e: Based on talking to @Levi-Armstrong it would probably be best to just pass in the planner config object.


schornakj:
Replaced the Tesseract forward kinematics and environment pointers with a const reference to the OMPLPlannerConfig.

Levi-Armstrong:
@schornakj This should be addressed by #249. I made several similar changes to accommodate constraints.

Levi-Armstrong:
Addressed by #249

PR #237, opened 2020-02-27
Add/extension approach generator

DavidMerzJr:
Complements the previously added departure generator.


DavidMerzJr:
@Levi-Armstrong 
This seems like a natural counterpart to the previously created extension departure generator.  I needed it for the same project, so I thought I'd push it here.

DavidMerzJr:
@Levi-Armstrong I have addressed the clang warnings.

Issue #238, opened 2020-03-02
Tesseract with Python 3


mpowelson:
What does it take to use Tesseract_python with Python 3? Building on melodic and trying to import tesseract with Python 3 gives the following error. It imports fine with Python 2.
```
from . import _tesseract_python
ImportError: dynamic module does not define module export function (PyInit__tesseract_python)
```

Motivation is that I'm interested in using Tesseract Python in ROS 2.

mpowelson:
I should also note that I have been unsuccessful getting this to build in ROS 2. It always finds Python 2 first.

mpowelson:
For me the problem is solved by adding the following. 
```
set(Python_ADDITIONAL_VERSIONS 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0)

find_package(PythonInterp 3 REQUIRED)
find_package(PythonLibs REQUIRED)
```

Neither adding the 3 to the find_package nor the additional versions alone would make it work. Obviously this is a bit of a problem since we still need Python 2 for ROS 1.When we split out tesseract_ros_python we'll need to add a check for the correct python version based on the ROS version. It looks like we can just grab ROS_PYTHON_VERSION

mpowelson:
This was addressed in #241 . There is now a cmake flag PYTHON_VERSION that can be used to manually select what Python version to use. Additionally, if that is not set it now uses the ROS_PYTHON_VERSION if available.

PR #239, opened 2020-03-03
Set SWIG Python version based on ROS_PYTHON_VERSION

mpowelson:
Partially addresses #238 . The correct version of Python 3 is now found in ROS 2. However tesseract_python still does not build because of tessseract_ros. Breaking that out will be handled in a future PR. Hopefully this will preemptively fix Noetic assuming they set that variable.


mpowelson:
Replaced by #241

PR #243, opened 2020-03-06
Make passing waypoints as ConstPtr

Levi-Armstrong:
This is open for discussion but I thought it was odd that they were being passed in as non-const waypoints.


Levi-Armstrong:
This is being replaced by tesseract command language.

PR #247, opened 2020-03-09
Configurable post-plan collision check

marip8:
By default, the current motion planners perform a post-planning collision check using the continuous collision checker. In some cases, it is desirable to use the discrete collision checker (or no collision checker) instead. This PR adds an argument to the base class for specifying which type of post-planning collision check should be performed


marip8:
@Levi-Armstrong how do I go about modifying the python interface for C++ members that I have changed?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
```diff
@@ -39,6 +39,14 @@ namespace tesseract_motion_planners
 class MotionPlanner
 {
 public:
+  /** @brief Enumeration of the types of collision checks that can be performed on a planned trajectory */
+  enum class CollisionCheckType
+  {
+    NONE = 0,
+    DISCRETE = 1,
+    CONTINUOUS = 2
+  };
```
Levi-Armstrong: Do you think we should add single timestep option?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
```diff
@@ -39,6 +39,14 @@ namespace tesseract_motion_planners
 class MotionPlanner
 {
 public:
+  /** @brief Enumeration of the types of collision checks that can be performed on a planned trajectory */
```
Levi-Armstrong: Should we name this something slight different? The naming may be interpreted such that they think they are setting the collision check type used internal to the motion planner where this is controlling the post check of the trajectory.

Levi-Armstrong:
> @Levi-Armstrong how do I go about modifying the python interface for C++ members that I have changed?

In most cases you can get away with copying in the change header content in to the file with the same name except with the extension `.i`.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
marip8: What do you think about `PostPlanCollisionCheckType`? I initially had something like that, but I changed it because it seemed a bit wordy.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
marip8: What would a single timestep option mean in the context of collision checking a planned trajectory?

Levi-Armstrong: It would just check the individual waypoint and would not interpolate using longest valid segment.

Single Timestep - check only joint states in trajectory
Discrete Continuous - Interpolates based on longest valid segment and performs discrete collision check
Cast Continuous - Interpolates based on longest valid segment and performs a cast collision check.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
Levi-Armstrong: What about PostPlanCheckType? It would allow for us to integrated different checks other than collision if needed. If you agree I would also change the values to represent flags where you specify more than one check on the trajectory.

marip8: For this, are you thinking an enumeration where you could combine multiple checks together with a binary `|` operator? Like this:
```
enum PostPlanCheckType
{
  TYPE_1 = 1,
  TYPE_2 = 2,
  TYPE_3 = 4,
  TYPE_4 = 8
};

// Perform checks for type 1 and type 4
PostPlanCheckType my_check = TYPE_1 | TYPE_4;
```

Levi-Armstrong: That is correct.

marip8:
@Levi-Armstrong @mpowelson take a look at the trajectory validation class I created [here](https://github.com/ros-industrial-consortium/tesseract/pull/247/files#diff-15b4837fff57f66f346a5e2cd9ad02fbR61) when you get the chance and let me know what you think about this solution

Levi-Armstrong:
@marip8 Looks good. If you rebase using master CI should pass and this can be merged.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
```diff
@@ -55,7 +56,17 @@ class MotionPlanner
   void setRequest(const PlannerRequest& request) { request_ = request; }
 
   /** @brief Solve the planner request problem */
```
Levi-Armstrong: This is a duplicate of the line below.

PR #248, opened 2020-03-11
Remove OMPL EST planner from the unit tests

Levi-Armstrong:


Levi-Armstrong:
@DavidMerzJr Can you please review?

DavidMerzJr:
It might be nice to leave a comment as to why an extant struct is excluded from the tests.
```
// tesseract_motion_planners::ESTConfigurator,  // Do not run this one during the test.  As a random planner with a short test-length timeout, it frequently causes the test to fail.
```
Other than that, looks good.

DavidMerzJr:
I don't seem to have sufficient permission to merge, or even to approve.

PR #251, opened 2020-03-14
update readme and rosinstall to include tesseract_ros

Levi-Armstrong:
- Update readme by removing information related to tesseract_ros.
- Switch to using GitHub Actions


Levi-Armstrong:
@mpowelson I switched CI to use github actions because you get more resources and everything now builds faster. Here is the [PR CI](https://github.com/Levi-Armstrong/tesseract-1/actions/runs/55837298) to show that it passes. Once merged it will run on this repository.

PR #259, opened 2020-03-18
Configurable trajopt collision pairs

marrts:
When using the default trajopt planner config structure all links have the same collision distance setting between them. This PR adds a feature to specify constraint and cost distances between specific link pairs if increased control is desired. Additionally you can specify negative distances and these changes allow those specified negative distances to pass the post-plan collision check as long as the collisions are within the specified threshold.


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_config.h
```diff
@@ -63,6 +63,13 @@ struct TrajOptPlannerConfig
   /** @brief Trajopt problem to be solved (Required) */
   trajopt::TrajOptProb::Ptr prob;
 
+  /**@brief Special link collision cost distances
+   * This is a vector of std::tuple<link1, link2, collision distance, collision coefficient */
+  std::vector<std::tuple<std::string, std::string, double, double>> special_collision_cost;
```
Levi-Armstrong: Why not use the Trajopt SafetyMarginData class here?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_config.h
```diff
@@ -63,6 +63,13 @@ struct TrajOptPlannerConfig
...
+  /**@brief Special link collision constraint distances
+   * This is a vector of std::tuple<link1, link2, collision distance, collision coefficient */
+  std::vector<std::tuple<std::string, std::string, double, double>> special_collision_constraint;
```
Levi-Armstrong: Same as above.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_config.h
marrts: I don't fully understand how those work. In order to use them would I have to have both a SafetyMarginData struct and a vector of pairs of strings for both the cost and collision terms so that I can call getPairSafetyMarginData later when I want to modify the default collision terms being created in trajopt_planner_default_config.cpp? I am unaware of a way to iterate through the SafetyMarginData class, but I'm all for using something already created rather than a vector of an arbitrary tuple like I have defined here.

Levi-Armstrong: > I don't fully understand how those work. In order to use them would I have to have both a SafetyMarginData struct and a vector of pairs of strings for both the cost and collision terms so that I can call getPairSafetyMarginData later when I want to modify the default collision terms being created in trajopt_planner_default_config.cpp?

I recommend taking a look at the class which is documented. It stores the pair internally and if you want to overwrite an existing one you just setPairSafetyMarginData again. Currently there is no way to remove one, but I would add a removePairSafetyMarginData function to the class. 

> I am unaware of a way to iterate through the SafetyMarginData class, but I'm all for using something already created rather than a vector of an arbitrary tuple like I have defined here.

In most case you iterate over the contacts and query getPairSafetyMarginData to check if it is a valid contact.

Levi-Armstrong:
Also clang tidy is failing. The readme explains how to build locally with clang tidy enabled.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_config.h
marrts: Okay, I was able to get SafetyMarginData classes to work for me

marrts:
Thanks, I didn't realize I could check it offline

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/config/trajopt_planner_config.h
```diff
@@ -63,6 +63,11 @@ struct TrajOptPlannerConfig
   /** @brief Trajopt problem to be solved (Required) */
   trajopt::TrajOptProb::Ptr prob;
 
+  /**@brief Special link collision cost distances */
+  trajopt::SafetyMarginData::Ptr special_collision_cost = nullptr;
```
Levi-Armstrong: I recommend using `trajopt::SafetyMarginData::Ptr special_collision_cost { nullpltr };`

marrts: Fixed.

marrts:
> I recommend updating the checkTrajectory that takes a trajopt::SafetyMarginData::Ptr and if it nullptr it does not use it. If it is not a nullptr it uses this to decide if it is a valid contact. Also update the TrajectoryValidator constructor to take a trajopt::SafetyMarginData::Ptr which then gets passed to the checkTrajectory.

Now it passes the `trajopt::SafetyMarginData::Ptr` to TrajectoryValidator, but I can't pass it to checkTrajectory because that would require tesseract_environment to depend on trajopt which would create a circular dependency. It still follows the general idea you proposed of checking for a nullptr and performing additional collision checks only if it is not a nullptr.

> Second does there need to be two different SafetyMarginData::Ptr one for cost and one for constraint?

It is not necessary to have two different ones, but the idea was you could also add specific costs that aren't necessary for the trajectory to pass, e.g. you want the end effector to stay as far as possible from a structure as possible but it doesn't have to maintain that distance and you don't want to bother checking that every link is trying to be that far away from each other

> Also does it need to go in the base class and could it go in the collision cost and collision constraint struct?

I would prefer it to be in the collision structs, but those are in the default config and the trajopt_motion_planner.cpp can only access members of the TrajOptPlannerConfig struct as opposed to the TrajOptPlannerDefaultConfig struct. Perhaps the collision cost and contraint config structs could be moved to the TrajOptPlannerConfig struct instead of the TrajOptPlannerDefaultConfig struct.

marrts:
I forgot to run the offline build check, I'll run that first before recommitting.

marrts:
Also I have a [PR against trajopt](https://github.com/ros-industrial-consortium/trajopt_ros/pull/177) that allows for checking if there are any negative collisions allowed in the `trajopt::SafetyMarginData::Ptr`. The other option is to default to `tesseract_collision::ContactTestType::ALL` if any collision pairs are given, but I think this would waste a lot of computation in cases where no negative constraint distances are given despite using special collision pairs

Levi-Armstrong:
> Now it passes the trajopt::SafetyMarginData::Ptr to TrajectoryValidator, but I can't pass it to checkTrajectory because that would require tesseract_environment to depend on trajopt which would create a circular dependency. It still follows the general idea you proposed of checking for a nullptr and performing additional collision checks only if it is not a nullptr.

It may be useful to update the checkTrajectory that takes a function `isContactResultsValid(const ContactResults&)`  then developers can pass there own post check on the contact results and tell the function to discard/keep this contact results. Then you could pass in a function that wraps the safety margin data check.

>I would prefer it to be in the collision structs, but those are in the default config and the trajopt_motion_planner.cpp can only access members of the TrajOptPlannerConfig struct as opposed to the TrajOptPlannerDefaultConfig struct. Perhaps the collision cost and contraint config structs could be moved to the TrajOptPlannerConfig struct instead of the TrajOptPlannerDefaultConfig struct.

Since it is part of the trajectory validator function that gets provided to the config is this still valid?

marrts:
> It may be useful to update the checkTrajectory that takes a function isContactResultsValid(const ContactResults&) then developers can pass there own post check on the contact results and tell the function to discard/keep this contact results. Then you could pass in a function that wraps the safety margin data check.

I'll look into doing this

> Since it is part of the trajectory validator function that gets provided to the config is this still valid?

You can access the trajopt problem but not the pci. Is there a good way to pull out the collision constraints from a trajopt problem?

Levi-Armstrong:
>You can access the trajopt problem but not the pci. Is there a good way to pull out the collision constraints from a trajopt problem?

I would add it to both collision config structures.

marrts:
It doesn't look like I can access the collision constraints in any way while in trajopt_motion_planner.cpp. The vector of inequality constraints can be seen from the trajopt problem, but this doesn't allow access to the SafetyMarginData structs. I think some reference to the collision pairs needs to be in the main TrajOptPlannerConfig struct. Would you prefer I move the collision config structs into the TrajOptPlannerConfig struct and just add the special_collision SafetyMarginData structs to there?

Also in regards to passing a function to the checkTrajectory function, I can't find a good clean way of doing that. I don't think I have the knowledge to implement it. It reminds me of the type erasure stuff you guys have discussed some, since a function needs to be passed without inheritance and I don't fully understand how it works.

Levi-Armstrong:
I will pull down what you have and see if it is possible.

Levi-Armstrong:
@marrts I created a PR against your branch.

marrts:
I don't know why this is failing

PR #262, opened 2020-03-20
WIP: Switch to using type erasure for forward kinematics

Levi-Armstrong:
We have been discussing for a while to switch to suing Type Erasure for interface within Tesseract. I have an initial pass for the Forward kinematics and I wanted to get this up sooner to begin discussion.

### DIFF
--- a/tesseract/tesseract/include/tesseract/forward_kinematics_manager.h
+++ b/tesseract/tesseract/include/tesseract/forward_kinematics_manager.h
@@ -183,7 +183,7 @@ public:
   {
     auto it = fwd_kin_manipulators_.find(std::make_pair(manipulator, name));
     if (it != fwd_kin_manipulators_.end())
-      return it->second->clone();
+      return std::make_shared<tesseract_kinematics::ForwardKinematics>(*(it->second));
 
     return nullptr;
   }
@@ -197,7 +197,7 @@ public:
   {
     auto it = fwd_kin_manipulators_default_.find(manipulator);
     if (it != fwd_kin_manipulators_default_.end())
-      return it->second->clone();
+      return std::make_shared<tesseract_kinematics::ForwardKinematics>(*(it->second));
 
     return nullptr;
   }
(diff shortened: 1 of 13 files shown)


mpowelson:
So perhaps you should clarify the motivation behind this. I can see doing this in TrajOpt in order to remove the dependency on Tesseract. What exactly is the goal here?

gavanderhoorn:
I was indeed going to ask the same question.

Levi-Armstrong:
The main reason is it remove  inheritance for the developer and leverage Template Polymorphism for interfaces. It also remove the requirement on the user to implement a clone method because the new interface is type aware and can call the copy constructor. Though I am not sure which is a better approach for interfaces but from what I have read suggests that this may be a preferred approach for interfaces. Is there any disadvantages with this approach compared to inheritance?

marip8:
An immediate downside I can see is that type erasure is a lot more complicated (for the implementer, not necessarily the user) than inheritance. If we could come up with a solid and consistent paradigm/model for implementation and document it well, it may not be too difficult to use type erasure with other classes. I think the kinematics class could be a good test case to see what it takes to make type erasure work, but, like @mpowelson mentioned, there might be more immediate value in doing this for TrajOpt

Levi-Armstrong:
[Here](https://github.com/esa/pagmo2/wiki/From-1.x-to-2.x) is what started the discussions around type erasure. The video is a good watch.

Levi-Armstrong:
Adding @jrgnicho and @jdlangs to the discussion.

johnwason:
Using type erasure instead of inheritance will most likely make Python interfaces a lot more complicated. I have also found that type erasure often has a pretty heavy performance penalty.

Have you looked at the boost range and boost assign libraries? A lot of the clunky stl operations can be done very cleanly using the various boost libraries.

Levi-Armstrong:
> I have also found that type erasure often has a pretty heavy performance penalty.

Do you know why it has a performance penalty?

johnwason:
Type erasure has to dispatch each function in some way. This is done using virtual functions or complex case statements normally. In the past, virtual functions in C++ were quite slow. This is no longer the case in modern C++ compilers so the benefits of type erasure aren't as great.

The type erasure in boost like Boost.Any and Boost.Function both have a pretty heavy performance penalties compared to virtual functions.

Boost provides template metaprogramming functions that can do a lot of what type erasure does.

johnwason:
I meant to say that type erasure uses function pointers and switch statements before

PR #264, opened 2020-03-26
Create a test suite for tesseract_collision to be used by external packages

Levi-Armstrong:
This breaks out the tesseract_collision unit tests such that they could be used by external packages. This is useful for anyone who wants to implement a contact manager and run it though all of the tests that tesseract_collision uses.


tesseract/tesseract_collision/CMakeLists.txt
```diff
@@ -83,15 +83,32 @@ target_include_directories(create_convex_hull SYSTEM PUBLIC
...
+# Create test suite interface
+add_library(${PROJECT_NAME}_test_suite INTERFACE)
+target_link_libraries(${PROJECT_NAME}_test_suite INTERFACE tesseract::tesseract_geometry)
+tesseract_target_compile_options(${PROJECT_NAME}_test_suite INTERFACE)
+tesseract_clang_tidy(${PROJECT_NAME}_test_suite)
+tesseract_code_coverage(${PROJECT_NAME}_test_suite ALL EXCLUDE ${COVERAGE_EXCLUDE})
+target_compile_definitions(${PROJECT_NAME}_test_suite INTERFACE TEST_SUITE_DATA_DIR="${CMAKE_INSTALL_PREFIX}/share/${PROJECT_NAME}/test_suite/data")
+target_include_directories(${PROJECT_NAME}_test_suite INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+    "$<INSTALL_INTERFACE:include>")
```
mpowelson: So this interface library includes all of the headers in tesseract_collision? Could this be something like below so that it only includes the hpp from test_suite (or something else)?
```
target_include_directories(${PROJECT_NAME}_test_suite INTERFACE
    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/test_suite>"
    "$<INSTALL_INTERFACE:include>")
```

mpowelson: You obviously know more about this than me, but mostly I'm just asking are you sure this is right?

Levi-Armstrong: It just encodes the location to find includes shown below `${_IMPORT_PREFIX}/include`.  Interface are mainly for header only so it just stores the location where it should find headers. Ideally when you create a Debian you then create a separate package for this target which only installs the headers under test_suite.

```
# Create imported target tesseract::tesseract_collision_test_suite
add_library(tesseract::tesseract_collision_test_suite INTERFACE IMPORTED)

set_target_properties(tesseract::tesseract_collision_test_suite PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "TEST_SUITE_DATA_DIR=\"/home/larmstrong/catkin_ws/trajopt_only_ws/devel/share/tesseract_collision/test_suite/data\""
  INTERFACE_COMPILE_FEATURES "cxx_std_14"
  INTERFACE_COMPILE_OPTIONS "-mno-avx;-Werror=all;-Werror=extra;-Werror=conversion;-Werror=sign-conversion;-Wno-sign-compare"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/usr/include/eigen3;\$<TARGET_PROPERTY:tesseract::tesseract_common,INTERFACE_INCLUDE_DIRECTORIES>"
  INTERFACE_LINK_LIBRARIES "tesseract::tesseract_geometry"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/usr/include/eigen3;\$<TARGET_PROPERTY:tesseract::tesseract_common,INTERFACE_INCLUDE_DIRECTORIES>"
)
```

Levi-Armstrong: It looks like there is another option which is more direct, which is to leverage the target_source and remove the add target_include_directory. What do you think? I think they achieve the same thing but the only difference is this target will only provide these files when linking/building.

```
Complications For Installing

Specifying PRIVATE sources is relatively easy and has few difficulties. The location of each source file is clear and only needs to be considered within the build. Any PUBLIC or INTERFACE sources give rise to additional factors which must be considered. Within the build, paths to non-private sources are handled just like private ones, but when the project is installed, non-private paths have to make sense not only in the project’s own build, but also in the build of anything consuming the installed project. The paths used for the project’s own build will be specific to the machine on which that build is performed and to the directory in which it is built. Once installed, those directories will likely not be accessible, so the paths would need to be replaced with something else that makes sense for the installed set of files. This is made possible using the BUILD_INTERFACE and INSTALL_INTERFACE generator expressions. The following example shows how to provide different paths for the same file in the two different contexts:

target_sources(myHeaderOnly
    PUBLIC
        $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/algo.h>
        $<INSTALL_INTERFACE:include/algo.h>
)
install(FILES algo.h DESTINATION include) 

While the above solves the build/install differences, it also has the drawback of requiring us to spell out every BUILD_INTERFACE path as absolute once again. We can no longer use our target_sources_local() helper function that we defined earlier. This is the cost of supporting the installation of non-private source files. Projects should weigh up the loss of convenience and added complexity against the expected benefits to be gained by making sources non-private.
```

Issue #263, opened 2020-03-26
How to retrieve objects from the environment ?


sduminy:
Hi all,
I have been working on a pick & place task with object recognition. I am working from a code initially based on OpenRave and trying to switch to another library based on tesseract for motion planning. 
I would like to know if there is a method to retrieve object's position and size from the environment with the object's name (those features have been added previously during an objects detection phase.
I am looking into tessereact_environment source code but without success...
Does someone know if such a method is implemented on tesseract ?
Thanks,

mpowelson:
Perhaps [getLink](https://github.com/ros-industrial-consortium/tesseract/blob/a564b4d0d29d8783ae165d1e0d14fa7b94fe6b73/tesseract/tesseract_environment/include/tesseract_environment/core/environment.h#L177) is what you are looking for? Then the  link will have visual and collision elements associated with them which have some geometry type (e.g. box, cylinder, etc) .

sduminy:
Yes I am trying to work with that ! Do you know what method can I use with a tesseract_scene_graph::Link to get the pose and size of the object attached to this link ?

Levi-Armstrong:
I would us the environment getCurrentState or if you may also use the getState where you provide joint values. This will provide an object that contains a map of link names to transforms in world coordinate system. Then using getLink you have the geometry and the geometry pose relative to the links origin so you can calculate the location of the geometry in world coordinate system.

The size of the object will require you to calculate. For primitive shapes this is straight forward but for meshes and convex hulls you would need to do the calculation. Are you just wanting the bounding box?

mpowelson:
@Levi-Armstrong We could use some doxygen on EnvState https://github.com/ros-industrial-consortium/tesseract/blob/a564b4d0d29d8783ae165d1e0d14fa7b94fe6b73/tesseract/tesseract_environment/include/tesseract_environment/core/types.h#L54

sduminy:
Yes so actually the objects I want to retrieve the pose and size from, have first been added as link with the `tesseract_environment::addLink(tesseract_scene_graph::Link link, tesseract_scene_graph::Joint joint)` method.
Using `tesseract_environment::getLink(const std::string& name)` I should have the right link. But then, how can I extract the pose and size of it ?

mpowelson:
So if you want the link itself and all the geometry information associated with it (like size), you will need to do getLink. Presumably the size information you want is in either the visual or collision geometry.

[...]

PR #265, opened 2020-03-27
Store joint transforms in EnvState structure

Levi-Armstrong:
This PR stores joint_transform map in the EnvState.


tesseract/tesseract_environment/src/kdl/kdl_state_solver.cpp
```diff
@@ -218,21 +214,23 @@ void KDLStateSolver::calculateTransformsHelper(tesseract_common::TransformMap& t
     Eigen::Isometry3d local_frame, global_frame;
     KDLToEigen(current_frame, local_frame);
     global_frame = parent_frame * local_frame;
-    transforms[current_element.segment.getName()] = global_frame;
+    state.link_transforms[current_element.segment.getName()] = global_frame;
+    if (current_element.segment.getName() != scene_graph_->getRoot())
+      state.joint_transforms[current_element.segment.getJoint().getName()] = global_frame;
```
mpowelson: What does this do?

mpowelson: perhaps that should go in the detailed description of the docs?

marip8: Does this set both the link transform and joint transform to the same transformation (`global_frame`)?

Levi-Armstrong: Yea, so a child link of a joint has the same transform as the joint.

tesseract/tesseract_environment/include/tesseract_environment/core/types.h
```diff
@@ -42,16 +42,28 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_environment
 {
-/** @brief This holds a state of the environment */
+/**
+ * @brief This holds a state of the environment
+ *
+ * It provides a way to look up the location of a link/joint in world coordinates system by link/joint name. It is
+ * possible to get the joint transform using the child link transfrom of the joint, but they are both provided for
```
mpowelson: `transfrom` misspelled

mpowelson:
Note that this is a breaking change. We should comment over on #263 to be friendly.

PR #270, opened 2020-03-30
Update types.i

Levi-Armstrong:


mpowelson:
I guess coverage went down since it only did a partial build last time?

PR #271, opened 2020-03-30
Remove tesseract_ext submodule

mpowelson:


mpowelson:
This pull request will require that repos relying on wstool clong submodules add tesseract_ext to the rosinstall

PR #274, opened 2020-03-31
Improve tesseract_collision test coverage.

Levi-Armstrong:
This PR improves the unit test coverage of tesseract_collision. Most of the remaining code not covered is section that should not occur.


Levi-Armstrong:
@mpowelson I will rebase after approval to reduce the commits.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
```diff
@@ -261,32 +253,47 @@ struct CastHullShape : public btConvexShape
...
+    throw std::runtime_error("not implemented");
+  }
+
   void getAabbSlow(const btTransform& /*t*/, btVector3& /*aabbMin*/, btVector3& /*aabbMax*/) const override
   {
     throw std::runtime_error("shouldn't happen");
   }
 
-  void setLocalScaling(const btVector3& /*scaling*/) override {}
+  void setLocalScaling(const btVector3& /*scaling*/) override { throw std::runtime_error("shouldn't happen"); }
```
mpowelson: This is not a very descriptive error message

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
```diff
@@ -476,13 +483,17 @@ inline void calculateContinuousData(ContactResult* col,
   // TODO: this section is potentially problematic. think hard about the math
   if (shape_sup0 - shape_sup1 > BULLET_SUPPORT_FUNC_TOLERANCE)
   {
+    // LCOV_EXCL_START
     col->cc_time[link_index] = 0;
     col->cc_type[link_index] = ContinuousCollisionType::CCType_Time0;
+    // LCOV_EXCL_STOP
```
mpowelson: Why are these excluded?

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
```diff
@@ -699,7 +710,7 @@ struct DiscreteBroadphaseContactResultCallback : public BroadphaseContactResultC
                            int /*index1*/) override
   {
     if (cp.m_distance1 > static_cast<btScalar>(contact_distance_))
-      return 0;
+      return 0;  // LCOV_EXCL_LINE
```
mpowelson: There are a bunch of lines like this. Why are these being excluded?

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
Levi-Armstrong: Yea all of these functions should never get called but must be implemented. This is to catch if anything changes within Bullet.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
Levi-Armstrong: This is controlled by Bullet and I do not know of a good way create a condition where this gets triggered. I believe these are here to catch degenerate case of collision checking which are hard to create and in theory should not occur.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
Levi-Armstrong: I do not believe this should ever be reached but left it here. I am not sure if we should add an assert.

Levi-Armstrong: I removed these ones.

mpowelson:
It looks like Xenial ran out of disk space. You suggestion [HERE](https://github.com/swri-robotics/collaborative-robotic-sanding/blob/68bfb7a001909982a7211a915b899d542ea62fe2/.github/workflows/main.yml#L39) worked in CRS (at least in the short term)

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
mpowelson: Well if they are useful to keep that's fine. Don't feel pressured to remove it if you think it is useful. I just wanted to make sure we weren't checking a condition that will never happen or excluding from coverage something that could sometimes happen.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
mpowelson: Perhaps we should change the message to something like
"If you are seeing this error message then something in Bullet must have changed. <Insert some idea about what you might do to fix it>"

Levi-Armstrong: I updated the comment.

tesseract/tesseract_collision/include/tesseract_collision/bullet/bullet_utils.h
Levi-Armstrong: I will remove it until I have time to further investigate and confirm that this never happens.

Levi-Armstrong:
@mpowelson What do you think about moving the upstream (tesseract_ros) build and running test to a cron job. You are allowed multiple workflows so I would create one just for the cron job which runs nightly.

mpowelson:
I would say its probably not ideal. If I broke something in tesseract_ros I'd rather find out immediately than the next morning. I suppose you could use something like [this](https://github.com/Khan/pull-request-comment-trigger) to allow us to trigger a tesseract_ros CI build manually if we think a PR might effect it. What's the motivation?

mpowelson:
If you are going to do it (I see you already did), would you want to add [tesseract_ros2](https://github.com/ros-industrial-consortium/tesseract_ros2/blob/master/.github/workflows/main.yml)? It has a shiny new eloquent build.

PR #277, opened 2020-04-06
Fix Tesseract Nightly Workflow

mpowelson:
Previously was getting `pull_request requires a map value` error. 

See [this example](https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request)


Levi-Armstrong:
It triggered!

PR #275, opened 2020-04-03
Add Bullet Discrete Simple ContactTest Benchmarks

mpowelson:
Adds the beginnings of a benchmarking framework for Tesseract Collision. Currently it adds tests for one manager running contactTest on primitives (this is 300 tests by itself). The idea is that I will add another benchmark executable for each manager. Then we can compare the outputs of each one as well as the outputs of each over time.

Additions
1.  Macro to run benchmarks similar to tests
2.  Benchmarking library in test_suite
3.  Google benchmark executable for BulletDiscreteManager

To test, build with `-DTESSERACT_ENABLE_BENCHMARKING=ON` and run the executable `/tests/benchmarks/tesseract_collision_bullet_discrete_simple_benchmarks`

PR #278, opened 2020-04-06
Add Benchmarking Workflow

mpowelson:
Depends on #275 

Adds a separate workflow that builds and runs the benchmarks. The benchmark json output is then used to populate a gh_pages. [Here is an example result on my fork](https://mpowelson.github.io/tesseract/dev/bench/).[THIS](https://rhysd.github.io/github-action-benchmark/dev/bench/) is what it should look like over time. We will likely want to change the html later to make it a little more friendly since it currently makes one plot per benchmark. However, this should be a place to start. It works by reading in the json file, making changes to the html, and committing those changes to a gh-pages branch. 

Note: This will require some more setup from someone with more permissions that me before we will see any output (@Levi-Armstrong ). Notably we will need to set up an access token, add it to the repo secrets, and create the gh_pages branch. See [HERE](https://github.com/rhysd/github-action-benchmark#charts-on-github-pages-1)

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -1,0 +1,69 @@
+name: Benchmarking
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 5 * * *'
+
+jobs:
+  industrial_ci:
+    name: ${{ matrix.env.CI_NAME }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        env:
+          - {CI_NAME: Bionic-Benchmark,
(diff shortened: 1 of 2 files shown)


mpowelson:
@Levi-Armstrong It doesn't like the token. It is called PERSONAL_GITHUB_TOKEN, right?

Levi-Armstrong:
I think the issue is because of the following statement: `Secrets are not passed to workflows that are triggered by a pull request from a fork.` It looks like this may only run after a merge or the nightly.

mpowelson:
It worked. Results are here https://ros-industrial-consortium.github.io/tesseract/dev/bench/

PR #281, opened 2020-04-07
Add broadphase overlap filter callback

Levi-Armstrong:


mpowelson:
This adds a callback to the broadphase that filters out things in the ACM, objects in the same link, etc.

PR #282, opened 2020-04-07
Add other collision checkers to benchmarks

mpowelson:


mpowelson:
@Levi-Armstrong Is this failing because the run_benchmarks target is getting built before that directory is installed? If so, is there a way around that other than adding running the benchmarks to the post script rather than using a cmake target?

mpowelson:
I added tesseract_collision_test_suite_benchmarks to the exported targets. I had missed it before

mpowelson:
Apparently `${{ secrets.GITHUB_TOKEN }}` expires in 60 minutes which for this run was longer than it took to get to the point where it comments. I guess we'll have to switch back to the personal access token and figure out which permissions it needs.

@Levi-Armstrong the other option is that I just turn off the pull request trigger, and we stop wasting time on this.

mpowelson:
So the large scale benchmarks apparently take a _really_ long time on the github actions servers.

mpowelson:
@Levi-Armstrong This is working again, so it is ready for review

PR #283, opened 2020-04-08
Move tesseract_collision test resources to tesseract_support

Levi-Armstrong:
@mpowelson This should solve the issue in #282.


mpowelson:
It looks good to me other than the fact the unit tests are failing - presumably because you changed the number of vertices in that mesh?

Levi-Armstrong:
@mpowelson Everything is now resolved with CI. You good with merging?

PR #289, opened 2020-05-11
add process definition generator interface

colin-lewis-19:
An abstract process definition generator


Levi-Armstrong:
Looks like you just need to run clang format. Run the bash file `.run_clang_format`

PR #295, opened 2020-05-27
Add IFOPT to the rosinstall

mpowelson:
This is a new dependency for TrajOpt (and therefore Tesseract) until IFOPT is released again


mpowelson:
@Levi-Armstrong This should work after the trajopt PR is merged.

dependencies.rosinstall
```diff
@@ -14,3 +14,7 @@
     local-name: opw_kinematics
     uri: https://github.com/Jmeyer1292/opw_kinematics.git
     version: master
+- git:
```
Levi-Armstrong, line 17: I believe this can be removed. It should be included from the trajopt rosinstall.

mpowelson: Are we merging them? I didn't think wstool could be run recursively, hence descartes_light in trajopt's rosinstall.

PR #299, opened 2020-06-01
Add Noetic CI build

Levi-Armstrong:


mpowelson:
@Levi-Armstrong I added some changes to this branch in my branch [HERE](https://github.com/ros-industrial-consortium/tesseract/compare/master...mpowelson:feature/NoeticCI). It now fails because OMPL is not in the docker. When I added installing it to the before script it timed out.

Levi-Armstrong:
Replaced by #305

PR #300, opened 2020-06-01
Update Docs index.html

Levi-Armstrong:
The path was incorrect to the doxygen.

The Doxygen can be viewed through the link below.
https://ros-industrial-consortium.github.io/tesseract/doxygen/html/index.html


Levi-Armstrong:
> So will this make the github io address in the readme correct?

Yep

PR #301, opened 2020-06-11
Switch octomap to use CMake targets

mpowelson:
Something was not being exported correctly such that anything that used environment.h had to link against octomap. This has been fixed by switching to use the octomap targets.

### DIFF
--- a/tesseract/tesseract_collision/CMakeLists.txt
+++ b/tesseract/tesseract_collision/CMakeLists.txt
@@ -14,6 +14,18 @@ find_package(fcl REQUIRED)
 # We need to find ccd ourselves because FCL doesn't do it in their fclConfig.cmake
 find_package(ccd REQUIRED)
 
+# These targets are necessary for 16.04 builds. Remove when Kinetic support is dropped
+if(NOT TARGET octomap)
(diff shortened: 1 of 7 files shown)


Levi-Armstrong:
For kinetic builds to work we will need to add this to the CMakeLists and config file.

```
if(NOT TARGET octomap)
  add_library(octomap INTERFACE IMPORTED)
  set_target_properties(octomap PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${OCTOMAP_INCLUDE_DIRS}")
  set_target_properties(octomap PROPERTIES INTERFACE_LINK_LIBRARIES "${OCTOMAP_LIBRARIES}")
endif()

if(NOT TARGET octomath)
  add_library(octomap INTERFACE IMPORTED)
  set_target_properties(octomath PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${OCTOMAP_INCLUDE_DIRS}")
  set_target_properties(octomath PROPERTIES INTERFACE_LINK_LIBRARIES "${OCTOMAP_LIBRARIES}")
endif()
```

mpowelson:
There was a copy/paste error.

PR #305, opened 2020-06-16
Add Noetic CI Build

Levi-Armstrong:


mpowelson:
I removed the ompl commit because it takes too long to run in CI for some reason. We'll have to add it to the docker. I'm a bit surprised it wasn't already in it.

Levi-Armstrong:
So libompl-dev is not sufficient?

Levi-Armstrong:
Or it needs to be added to the docker?

mpowelson:
It should be. It just failed yesterday not finding it. Maybe something else was wrong. Here is the build
https://github.com/mpowelson/tesseract/runs/773620881?check_suite_focus=true

mpowelson:
libompl-dev worked locally. Since it failed in CI I assumed that it was not in the docker. When I added installing it to the before script, CI took 6 hours and timed out for some reason.

Levi-Armstrong:
Ok, I just wanted to make sure I did not need to build from source. I am rebuilding the docker now for noetic.

mpowelson:
Someday we should really fix industrial docker to build them automatically.

Levi-Armstrong:
Yea, it is added to the repo but it is not pushing them to docker hub. I need to check with @gavanderhoorn to see if there is something I need to do so they show up on docker hub.

Levi-Armstrong:
A header needs to be removed that no longer exists. See below.

```
  /root/target_ws/src/tesseract/tesseract/tesseract_planning/tesseract_motion_planners/test/trajopt_planner_tests.cpp:35:10: fatal error: urdf_parser/urdf_parser.h: No such file or directory
     35 | #include <urdf_parser/urdf_parser.h>
```

mpowelson:
> A header needs to be removed that no longer exists. See below.
> 
> ```
>   /root/target_ws/src/tesseract/tesseract/tesseract_planning/tesseract_motion_planners/test/trajopt_planner_tests.cpp:35:10: fatal error: urdf_parser/urdf_parser.h: No such file or directory
>      35 | #include <urdf_parser/urdf_parser.h>
> ```

Does it have to be replaced with something?

Levi-Armstrong:
> Does it have to be replaced with something?

yea by tesseract_urdf, the newer compiler must be catching this now.

Levi-Armstrong:
@johnwason Any thoughts on what may be wrong when building tesseract_python on Noetic?

johnwason:
Have you tried building locally? Does Noetic add the tesseract_python directories to PYTHON_PATH correctly? Is there a Python 2 / Python 3 mismatch?

Levi-Armstrong:
It looks like it is an issues when building with colcon. We are adding an environment hook to update the pythonpath variable.

Levi-Armstrong:
@ipa-mdl We added a Focal CI build and we are running into a python run tests issue. We were getting the same issue locally when using Colcon and we had to add another environment hook so the PYTHONPATH environment variable so it will build locally. We had to source before running the python tests, but in CI it is still failing so not sure where the issues is.

ipa-mdl:
> We had to source before running the python tests,

Which file did you source?
You can even do this with `industrial_ci` now: https://github.com/ros-industrial/industrial_ci/pull/524
In your case `BEFORE_RUN_TARGET_TEST_EMBED`.

Normally, it is not necessary to source the to-be-tested workspace in order to run the tests.
Another issue with colcon?
Can you use catkin-tools with noetic?
(The last time I checked it did not work yet, so I set colcon as the default builder)

Levi-Armstrong:
@mpowelson Did you source install/local_setup.bash on your machine to get the python tests to build?

Levi-Armstrong:
@ipa-mdl we can use catkin-tools.

mpowelson:
Yeah, catkin_tools worked fine. With colcon it worked with our added hooks and sourcing install/setup.bash

gavanderhoorn:
> Another issue with colcon?
> Can you use catkin-tools with noetic?
> (The last time I checked it did not work yet, so I set colcon as the default builder)

`catkin_tools` works if you install it with `pip` from current `HEAD` of the repository.

Levi-Armstrong:
@mpowelson I just created a PR against your branch to see if switching to use catkin_tools will work. I may need to update the docker to install catkin_tools using pip if it fails.

ipa-mdl:
So either
* `AFTER_INIT="pip install 'git+https://github.com/catkin/catkin_tools.git#egg=catkin_tools'"` and `BUILDER=catkin_tools`
* colcon and sourcing install/setup.bash in `BEFORE_RUN_TARGET_TEST_EMBED`
* BUILDER=catkin_make_isolated (not yet tested)

Levi-Armstrong:
@ipa-mdl Adding `BEFORE_RUN_TARGET_TEST_EMBED` along with the environment hooks for PYTHONPATH solved the issue. Thank you for the help.

PR #308, opened 2020-06-24
Add COLCON environment hooks to update ROS_PACKAGE_PATH

mpowelson:


.github/workflows/nightly.yml
```diff
@@ -43,7 +43,7 @@ jobs:
              CCACHE_DIR: "/home/runner/work/tesseract/tesseract/Xenial-Build-Nightly/.ccache",
              UPSTREAM_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Release",
              TARGET_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Release",
-             DOWNSTREAM_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Release"}
+             DOWNSTREAM_CMAKE_ARGS: "-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=14"}
```
Levi-Armstrong, line 46: Do you know why we are having to added this now?

mpowelson: No. We added it in tesseract_ros as well because Xenial was failing ([here](https://github.com/ros-industrial-consortium/tesseract_ros/pull/16)). I assume something about conflicting C++ versions specified in dependencies.

Levi-Armstrong: Ok

gavanderhoorn: Shouldn't the C++14 requirement be exported by the dependencies instead of fudging it in here?

Levi-Armstrong: I agree. I think there is an issue with how we are setting c++14 on the target. The feature option is not available on Xenial, but is on Bionic and later which is maybe where the issue is. In the case when the feature option is not available we set it as a compiler definition. May try switching to using the code below in the cmake macro over using the compiler definition to see if the issue gets resolved.

Current:
```
target_compile_options("${target}" INTERFACE -std=c++14 -mno-avx ${warning_flags})
```

Proposed Change:

```
target_compile_options("${target}" INTERFACE -mno-avx ${warning_flags})
set_property(TARGET ${target} PROPERTY CXX_STANDARD 14)
```

PR #309, opened 2020-06-24
Bugfix when OMPL simplifies down to two states 

pbeeson:
TrajOpt was happily copying the 2 state solution and moving on, which caused segfaults later on when > 2 states are assumed (it was indexing from -1 in places).

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/src/hybrid/ompl_trajopt_freespace_planner.cpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/src/hybrid/ompl_trajopt_freespace_planner.cpp
@@ -120,6 +120,14 @@ tesseract_common::StatusCode OMPLTrajOptFreespacePlanner::solve(PlannerResponse&
     return response.status;
   }
 
+  if (ompl_planning_response.joint_trajectory.trajectory.rows() == 2)
+  {
+    auto new_trajectory = tesseract_common::TrajArray(3, ompl_planning_response.joint_trajectory.trajectory.cols());
+    new_trajectory.topRows(2) = ompl_planning_response.joint_trajectory.trajectory.topRows(2);
+    new_trajectory.row(2) = new_trajectory.row(1);
+    ompl_planning_response.joint_trajectory.trajectory = new_trajectory;
+  }
+
   if (trajopt_config_)
   {
     // The trajectory from OMPL may not be the same size as the number of steps requested in the TrajOpt config
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/hybrid/ompl_trajopt_freespace_planner.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/hybrid/ompl_trajopt_freespace_planner.h
@@ -48,7 +48,7 @@ public:
   bool setConfiguration(OMPLPlannerConfig::Ptr ompl_config, TrajOptPlannerFreespaceConfig::Ptr trajopt_config);
 
   /**
-   * @brief Sets up the opimizer and solves a SQP problem read from json with no callbacks and dafault parameterss
+   * @brief Sets up the optimizer and solves an SQP problem read from json with no callbacks and default parameters
    * @param response The results of the optimization. Primary output is the optimized joint trajectory
    * @param check_type The type of validation check to be performed on the planned trajectory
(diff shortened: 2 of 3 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/src/hybrid/ompl_trajopt_freespace_planner.cpp
```diff
@@ -120,6 +120,14 @@ tesseract_common::StatusCode OMPLTrajOptFreespacePlanner::solve(PlannerResponse&
     return response.status;
   }
 
+  if (ompl_planning_response.joint_trajectory.trajectory.rows() == 2)
```
Levi-Armstrong, line 123: When you ran into this was your number of output states set to 2? If not then there may be an issue in the logic above as it should leverage OMPL interpolate to generate the set number of output states. This may be an issue in OMPL and if so we should interpolate using the interpolate function in the core folder utils.h for this conditional.

```
inline Eigen::MatrixXd interpolate(const Eigen::VectorXd& start,
                                   const Eigen::VectorXd& stop,
                                   int steps)
{
  assert(start.size() == stop.size());

  Eigen::MatrixXd result(start.size(), steps + 1);

  for (int i = 0; i < start.size(); ++i)
    result.row(i) = Eigen::VectorXd::LinSpaced(steps + 1, start(i), stop(i));

  return result;
}
```

pbeeson: If you turn on OMPL smoothing, it can get to 2 states when there are no obstacles nearby.  I'm fine replacing my functionality with the one above if you like, though I'm unsure that it will change the final results.

Levi-Armstrong: I looked at the tesseract ompl config doxygen and it does state that if simplify is set to true the number of output states is ignored. I think what you done is good. Though when using ompl + trajopt for I would recommend setting at minimum 10 to 20 states and simplify to false. In the case above trajopt will only have 3 states to adjust and you may have issues on the trajopt side if velocity, acceleration or jerk smoothing is enable because it is using finite difference to calculate these and there will not be enough states to do so.

pbeeson: Well, if you interpolate as you indicated, that may be better.  The only issue would be if someone called the planner from a location they were already at for some reason.  Then you'd likely get 2 equal states that go to 3 equal states.  If TrajOpt can handle that, then great.

Levi-Armstrong: > The only issue would be if someone called the planner from a location they were already at for some reason. Then you'd likely get 2 equal states that go to 3 equal states. If TrajOpt can handle that, then great.

Sorry, I am not following.

pbeeson: If a person calls the planner and the start/goal are the same (let' say this is programmatic and maybe they didn't mean to, but it happened).  Then (given you are allowing the simplify=true option) OMPL is going to return two states, that are identical.  So even if you interpolate and/or just copy the last state to make a 3 state plan, they will all be the same joint angles.  

My comment on this was that if TrajOpt can handle this without blowing up (diving by zero or something), then it should be fine.  So, I'm inclined to like your interpolation solution better if you think that handles the more general case as well as this case.

Levi-Armstrong: Now I understand. Should we maybe add a check early in the pipeline to catch this case and automatically return the input in the return response to prevent an unnecessary plan?

pbeeson: You could.  But then there's always going to be some corner case right.  I'd say as long as TrajOpt is numerically stable in the 3 point case (even if they are all the same 3 points), then we use your interpolation change above and let it pass through.  My two cents.

Levi-Armstrong: Sounds good.

PR #310, opened 2020-06-25
Set CXX_STANDARD 14 instead of -std=c++14 for xenial builds

mpowelson:


mpowelson:
See discussion on #308

PR #314, opened 2020-06-27
Add cube, box, and sphere shapes support to tesseract_viewer_python

johnwason:
Currently tesseract_viewer_python only supports meshes. This pull request adds support for cube, box, and sphere shapes. Other shapes can be added, but are more complicated and will require some effort to implement.


johnwason:
Updated to also include a grid pattern for the ground plane.

PR #321, opened 2020-07-09
Add cmake_common_scripts to rosinstall file

mpowelson:
This is a new dependency of opw_kinematics


Levi-Armstrong:
It looks like some of the builds are missing iwyu as a skip key.

mpowelson:
The codeCov build had a typo, but Focal appears to be genuinely broken. It has a bunch of PCL errors. Maybe it isn't being compiled with C++14?

mpowelson:
> The codeCov build had a typo, but Focal appears to be genuinely broken. It has a bunch of PCL errors. Maybe it isn't being compiled with C++14?

Should be fixed in https://github.com/Jmeyer1292/opw_kinematics/pull/42

PR #326, opened 2020-07-13
Tesseract planners: Make solve method const

mpowelson:
These should be able to be const now since the request and response hold all of the data.


Levi-Armstrong:
I am not sure if this will affect the ompl re-planning where you call solve multiple times and it continues to grow the graph. We still need to look at how this is affected with the new interface.

mpowelson:
I imagine calling solve multiple times is broken now. Currently all of the solve methods construct everything they need to run internally to the function call (including the problem). I think this may be important when calling solve inside a CPP Taskflow process for instance since each process must be threadsafe.

PR #327, opened 2020-07-14
Simplify instruction class signature

Levi-Armstrong:


tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/ompl_motion_planner.cpp
```diff
@@ -211,12 +211,12 @@ tesseract_common::StatusCode OMPLMotionPlanner::solve(const PlannerRequest& requ
 
     // Loop over the flattened results and add them to response if the input was a plan instruction
     Eigen::Index result_index = 0;
-    for (std::size_t plan_index = 0; plan_index < results_flattened.size(); plan_index++)
+    for (auto& instruction : results_flattened)
     {
-      if (instructions_flattened.at(plan_index).get().isPlan())
+      if (isPlanInstruction(instruction.get().getType()))
```
mpowelson, line 216: Perhaps we should make a utility `bool isPlanInstruction(const Instruction&)`. It is not terribly obvious from the name that you have to call getType on the instruction. We could leave this version as well if you want.

Levi-Armstrong: Updated along with waypoint type checks.

PR #331, opened 2020-07-15
Add Tesseract_process_managers

mpowelson:
Adds process managers based on [Taskfow](https://github.com/taskflow/taskflow/blob/master/CMakeLists.txt) for orchestrating Tesseract processes. Currently these include freespace and raster path motion planning, where each individual motion plan is executing in parallel based on the required dependencies. For example, freespace motions between rasters require that each raster already be planned. Additionally, sub-Taskflows are defined to facilitate defining a "fail-active" motion planning approach where a failure tree of planners can be defined for each raster or freespace motion.


mpowelson:
Some of these changes are in #332, so I will rebase as soon as that is merged.

mpowelson:
Ok @Levi-Armstrong Kinetic is going to fail unless we add your PPA to the docker or upgrade the cmake version. I'm not sure what that Focal error is.

But this should be ready for at least an architectural look. I'm going to move on from tracking minor bugs for now and start on the  raster process

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/move_instruction.h
```diff
@@ -54,16 +55,17 @@ class MoveInstruction
   using Ptr = std::shared_ptr<MoveInstruction>;
   using ConstPtr = std::shared_ptr<const MoveInstruction>;
 
-  MoveInstruction(Waypoint waypoint, MoveInstructionType type, const std::string& profile = "DEFAULT");
+  MoveInstruction(Waypoint waypoint,
+                  MoveInstructionType type,
+                  const std::string& profile = "DEFAULT",
+                  ManipulatorInfo manipulator_info = ManipulatorInfo());
```
Levi-Armstrong, line 61: Should this be a required by the user to provide? The only field that must be field out is `manipulator`

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/manipulator_info.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+struct ManipulatorInfo
+{
+  /** @brief Name of the manipulator group */
+  std::string manipulator;
+
+  /** @brief (Optional) IK Solver to be used */
+  std::string manipulator_ik_solver;
+
+  /** @brief The tool center point */
```
Levi-Armstrong, line 51: Mark this optional?

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/manipulator_info.h
```diff
@@ -0,0 +1,59 @@
...
+  Eigen::Isometry3d tcp{ Eigen::Isometry3d::Identity() };
+
+  /** @brief The working frame the waypoint is relative to */
```
Levi-Armstrong, line 54: Mark this optional and add that if empty the base link of the environment will be used?

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/move_instruction.h
```diff
@@ -54,16 +55,17 @@ class MoveInstruction
...
   void setWaypoint(Waypoint waypoint);
   const Waypoint& getWaypoint() const;
 
-  void setTCP(const Eigen::Isometry3d& tcp);
-  const Eigen::Isometry3d& getTCP() const;
-
-  void setWorkingFrame(std::string working_frame);
-  const std::string& getWorkingFrame() const;
+  void setManipulatorInfo(ManipulatorInfo info);
+  const ManipulatorInfo& getManipulatorInfoConst() const;
```
Levi-Armstrong, line 67: You can remove the `Const` from the name.

mpowelson: When they are the same name it errors saying I am trying to overload based on return type.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/move_instruction.h
mpowelson: Maybe. I guess it's unlikely that things would work if you don't fill it out, but you could set it later. 

Would you rather?
```
  MoveInstruction(Waypoint waypoint,
                  MoveInstructionType type,
                 ManipulatorInfo manipulator_info
                  const std::string& profile = "DEFAULT");
```

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/manipulator_info.h
mpowelson: done

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/move_instruction.h
mpowelson: done

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/manipulator_info.h
mpowelson: done

mpowelson:
Added the SimpleMotionPlanner. This should fill the need for a way to initialize the seed and also acts as a planner that does interpolation depending on what functions you put in the profile.  I still need to make a generateSeed-like wrapper for it and integrate it into my process manager examples. An eagle-eyed reviewer will also notice that many of the functions I have defined are not yet implemented. These shouldn't take long to do, but I don't know if I'll get to them this week. I imagine we will also want to make one that initializes them to a value (e.g. null or the current joint value).

mpowelson:
@Levi-Armstrong Would it be possible/advisable to make a default constructor for Instruction that constructs it as a NullInstruction? Ie so you can do things like
```
Instruction instruction;
...
instruction = CompositeInstruction()
```

Levi-Armstrong:
I don't see an issue with that.

mpowelson:
@Levi-Armstrong The tests are still failing. If I had to make a todo list I'd say

- [ ] Fix broken unit tests
- [ ] Implement the task validators for 
  - [ ] Format check (for checking that seed matches instruction)
  - [ ] Discrete Collision
  - [ ] Continous collision
  - [ ] etc
- [ ] Implement the rest of the step generators
- [ ] Address all of the todos placed everywhere in the code.
- [ ] Unit tests. I tried to structure everything such that components are easily testable. However, not many of them are currently tested.

Levi-Armstrong:
continued in PR #337

PR #338, opened 2020-07-29
Add visualization loader

Levi-Armstrong:
We are in the process of releasing a visualization tool for tesseract which leverages Ignition. This tool will allow for dynamically loading a visualization class along with default constructor which looks for the library in the snap location.

### DIFF
--- a/tesseract/tesseract_visualization/src/visualization_loader.cpp
+++ b/tesseract/tesseract_visualization/src/visualization_loader.cpp
@@ -1,0 +1,86 @@
+/**
+ * @file visualization_loader.h
+ * @brief Visualization Loader
+ *
+ * @author Levi Armstrong
+ * @date July 29, 2020
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2020, Southwest Research Institute
+ *
+ * @par License
+ * Software License Agreement (Apache License)
+ * @par
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


Levi-Armstrong:
@marrts Could you review? Once this is merged I will create a new snap leveraging the new interface and update the planning example to leverage this for visualization.

tesseract/tesseract_visualization/src/visualization_loader.cpp
```diff
@@ -0,0 +1,96 @@
...
+Visualization::Ptr VisualizationLoader::get()
+{
+  // Dynamically load visualization library
+  if (loader_)
+  {
+    try
+    {
+      if (loader_->isClassAvailable<tesseract_visualization::Visualization>(derived_class_))
+      {
+#ifndef CLASS_LOADER_LESS_4_0_0
```
marrts, line 61: Is this checking if class_loader is less than 4.0.0? I think you intended to check if it is at least 4.0.0. I am not too familiar with this syntax so I may just be misreading it too.

Levi-Armstrong: In the cmake file it checks if the version is less than 0.4.0 and if so it set a the compiler definition `CLASS_LOADER_LESS_0_4_0`. Then here I check if it is not defined, then if not defined then the version of class loader is at least version 0.4.0 or greater. This is because kinetic is still using boost shared pointer instead of std::shared_ptr so this will not work on kinetic.

marrts: Ah, that makes sense. All seems good to me then

Levi-Armstrong:
Had wrong version number. It should be `0.4.0` instead of `4.0.0`.

Levi-Armstrong:
Here is a [video](https://www.youtube.com/watch?v=gN0EM3Bm5bc) using the tesseract ignition visualization for the motion planner example. The first is the descartes results and the second is the results passing descartes as the seed to trajopt.

PR #342, opened 2020-07-30
Fix flatten utils to leverage start instruction

Levi-Armstrong:
- Adds example for simple freespace motion plan
- Adds non-const method for getting waypoint from Plan and Move Instructions
- Update motion planners to leverage new flatten utils and non-const getWaypoint
- Updates example to leverage new VisualizationLoader so you can use the new Tesseract Ignition Snap package for visualization.
- Updated visualization to support composite instruction
- Switched to using pointers instead of references in ProcessInput. Somewhere a copy was being made and it was returning a reference to a temporary variable. When switching to a raw pointer the issue was resolved but not sure where.
- Reworked utility function in the command language to be generic
- Switched to using unique pointers for SequentialTaskflow and ProcessGenerator because of the ability to abort. Do not want to pass multiple of the the same copy around.


Levi-Armstrong:
@marrts Do you time to review?

marrts:
Yeah, I'll take a look at this

tesseract/tesseract_planning/tesseract_motion_planners/examples/freespace_example.cpp
```diff
@@ -0,0 +1,235 @@
...
+  auto tesseract = std::make_shared<tesseract::Tesseract>();
+  boost::filesystem::path urdf_path(std::string(TESSERACT_SUPPORT_DIR) + "/urdf/abb_irb2400.urdf");
+  boost::filesystem::path srdf_path(std::string(TESSERACT_SUPPORT_DIR) + "/urdf/abb_irb2400.srdf");
+  tesseract->init(urdf_path, srdf_path, locator);
+
+  // Dynamically load ignition visualizer if exist
+  std::string library_path = "/home/larmstrong/catkin_ws/trajopt_only_ws/devel/lib/"
+                             "libtesseract_ignition_visualization_plugin.so";
+  std::string class_name = "tesseract_ignition::TesseractIgnitionVisualization";
+  tesseract_visualization::VisualizationLoader loader(library_path, class_name);
```
marrts, line 69: You are using a hardcoded file path from your system. 
I think it should work with just `tesseract_visualization::VisualizationLoader loader;`

marrts: I also cannot get the visualization to load on my system, I just installed the tesseract-ignition snap, but I get a warning that it fails to load the library whenever I try and run the examples

tesseract/tesseract_planning/tesseract_motion_planners/examples/freespace_example.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  // Solve OMPL Plan
+  PlannerResponse ompl_response;
+  OMPLMotionPlanner ompl_planner;
+  ompl_planner.plan_profiles["DEFAULT"] = ompl_plan_profile;
+  ompl_planner.problem_generator = tesseract_planning::DefaultOMPLProblemGenerator;
+  auto ompl_status = ompl_planner.solve(request, ompl_response);
+  assert(ompl_status);
+
+  // Plot Descartes Trajectory
```
marrts, line 153: I believe this is supposed to be OMPL trajectory

tesseract/tesseract_planning/tesseract_motion_planners/examples/raster_example.cpp
```diff
@@ -61,6 +65,16 @@ int main(int /*argc*/, char** /*argv*/)
   boost::filesystem::path srdf_path(std::string(TESSERACT_SUPPORT_DIR) + "/urdf/abb_irb2400.srdf");
   tesseract->init(urdf_path, srdf_path, locator);
 
+  // Dynamically load ignition visualizer if exist
+  std::string library_path = "/home/larmstrong/catkin_ws/trajopt_only_ws/devel/lib/"
+                             "libtesseract_ignition_visualization_plugin.so";
+  std::string class_name = "tesseract_ignition::TesseractIgnitionVisualization";
+  tesseract_visualization::VisualizationLoader loader(library_path, class_name);
```
marrts, line 72: Same thing regarding hardcoded filepath

tesseract/tesseract_planning/tesseract_process_managers/src/examples/freespace_manager_example.cpp
```diff
@@ -6,10 +6,10 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
 #include <tesseract/tesseract.h>
 #include <tesseract_command_language/command_language.h>
-#include <tesseract_motion_planners/trajopt/trajopt_motion_planner.h>
-#include <tesseract_motion_planners/trajopt/profile/trajopt_default_plan_profile.h>
-#include <tesseract_motion_planners/trajopt/profile/trajopt_default_composite_profile.h>
-#include <tesseract_motion_planners/core/utils.h>
+//#include <tesseract_motion_planners/trajopt/trajopt_motion_planner.h>
+//#include <tesseract_motion_planners/trajopt/profile/trajopt_default_plan_profile.h>
+//#include <tesseract_motion_planners/trajopt/profile/trajopt_default_composite_profile.h>
+//#include <tesseract_motion_planners/core/utils.h>
```
marrts, line 12: Not sure if you want to leave these here commented out or remove them

Levi-Armstrong: Removed.

tesseract/tesseract_planning/tesseract_motion_planners/examples/freespace_example.cpp
Levi-Armstrong: Version 10 is the one that should work, but it is currently in manual review and should be released soon. Once it is then should work.

tesseract/tesseract_planning/tesseract_motion_planners/examples/freespace_example.cpp
Levi-Armstrong: Fixed.

tesseract/tesseract_planning/tesseract_motion_planners/examples/raster_example.cpp
Levi-Armstrong: Removed, these were for testing and should have been removed. The default constructor will initialize with paths to the snap location.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
```diff
@@ -138,40 +138,24 @@ long getPlanInstructionsCount(const CompositeInstruction& composite_instruction,
...
- * @param include_composite If true, CompositeInstructions will be included in the final flattened vector
- */
-void flattenHelper(std::vector<std::reference_wrapper<Instruction>>& flattened,
-                   CompositeInstruction& composite,
-                   const bool& process_child_composites);
-
 /**
  * @brief Flattens a CompositeInstruction into a vector of Instruction&
+ *
+ * If instruction has a start instruction it is added but subsequent composites are not check for start instructions.
```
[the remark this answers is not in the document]
Levi-Armstrong: I reworded it. Let me know if it makes sense.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
```diff
@@ -196,9 +200,29 @@ std::vector<std::reference_wrapper<Instruction>> flattenToPattern(CompositeInstr
...
+ * @brief This loops over the instructions validates the structure
+ *
+ * Every plan instruction in /p composite_instruction should have a cooresponding CompositeInstruction
+ *
+ * @param composite_instructions
+ * @param composite_seed
+ * @return
+ */
+bool validateSeedStructure(const CompositeInstruction& composite_instructions,
+                           const CompositeInstruction& composite_seed);
```
marrts, line 225: This doesn't appear to be implemented yet

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
```diff
@@ -196,9 +200,29 @@ std::vector<std::reference_wrapper<Instruction>> flattenToPattern(CompositeInstr
  * vector
  * @return A new flattened vector referencing the original instruction elements
  */
-std::vector<std::reference_wrapper<const Instruction>> flattenToPattern(const CompositeInstruction& instruction,
-                                                                        const CompositeInstruction& pattern,
-                                                                        bool process_child_composites = false);
+std::vector<std::reference_wrapper<const Instruction>>
+flattenToPattern(const CompositeInstruction& composite_instruction,
+                 const CompositeInstruction& pattern,
+                 bool process_child_composites = false);
```
marrts, line 206: parameter `process_child_composites` doesn't match param in description, `include_composite`

marrts: There are a few instances of this, not sure if they're effectively supposed to mean the same thing or if the parameter is misidentified

tesseract/tesseract_planning/tesseract_process_managers/CMakeLists.txt
```diff
@@ -68,8 +68,8 @@ install(DIRECTORY include/${PROJECT_NAME}
...
 )
 
-if (TESSERACT_ENABLE_TESTING)
-  enable_testing()
+#if (TESSERACT_ENABLE_TESTING)
+#  enable_testing()
   tesseract_add_run_tests_target()
   add_subdirectory(test)
-endif()
+#endif()
```
marrts, line 75: Is this intentionally commented out?

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/default_processes/default_raster_processes.h
```diff
@@ -48,21 +48,24 @@ inline std::vector<ProcessGenerator::Ptr> defaultRasterProcesses()
...
   descartes_planner->plan_profiles["RASTER"] = std::make_shared<DescartesDefaultPlanProfileD>();
+  auto descartes_generator = std::make_shared<MotionPlannerProcessGenerator>(descartes_planner);
 
-  auto descartes = std::make_shared<MotionPlannerProcessGenerator>(descartes_planner);
-  descartes->planner = descartes_planner;
+  // Setup TrajOpt
+  auto trajopt_planner = std::make_shared<TrajOptMotionPlanner>();
+  trajopt_planner->problem_generator = &DefaultTrajoptProblemGenerator;
+  trajopt_planner->plan_profiles["FREESPACE"] = std::make_shared<TrajOptDefaultPlanProfile>();
+  trajopt_planner->composite_profiles["FREESPACE"] = std::make_shared<TrajOptDefaultCompositeProfile>();
```
marrts, line 65: Is this supposed to be RASTER?

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
```diff
@@ -138,64 +138,42 @@ long getPlanInstructionsCount(const CompositeInstruction& composite_instruction,
...
+/** @brief This is used for filtering only what you want in the vector **/
+using flattenFilter = std::function<bool(const Instruction&)>;
 
 /**
  * @brief Flattens a CompositeInstruction into a vector of Instruction&
- * @param instruction Input composite instruction to be flattened
- * @param include_composite Default: false. If true, CompositeInstructions will be included in the final flattened
- * vector
+ *
+ * If /p composite_instruction parameter has a start instruction it is added but child composites are not check for
```
mpowelson, line 147: I tried to post this earlier but I think it got lost somehow.

This seems like a somewhat not readily apparent logic to have in a flatten utility. This logic is a bit specific to the planning. Some processes might not want the start instruction and then they have to know to remove it. Others might want to include all of the start instructions.

For reference, the start instruction was the reason I added the include composite flag. That includes a reference to the composite itself so the caller can check if there is a start instruction. Then you get to decide if you use only the first, all l, or none of them.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
mpowelson: Unless it has changed, process_child_compisites is a misnomer. It used to specify where or not the raw composite was stored in the result (so that  you could access things like start_instructions). Either way the children get processed (since that is what flatten means).

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
```diff
@@ -68,13 +68,13 @@ struct ProcessInput
   const tesseract::Tesseract::ConstPtr tesseract;
 
   /** @brief Instructions to be carried out by process */
-  const Instruction& instruction;
```
mpowelson, line 67: Perhaps you could comment in the PR as to why this change was necessary for posterity

tesseract/tesseract_planning/tesseract_process_managers/CMakeLists.txt
Levi-Armstrong: fixed

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/default_processes/default_raster_processes.h
Levi-Armstrong: fixed.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
Levi-Armstrong: I am restructuring these utility function to make them generic where you can provide a function callback to indicate if something should be kept or not. Then I will create one that are specific to motion planning in the planning utilities.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
Levi-Armstrong: I am restructuring these utility function to make them generic where you can provide a function callback to indicate if something should be kept or not. Then I will create one that are specific to motion planning in the planning utilities.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
Levi-Armstrong: Updated

Levi-Armstrong:
@mpowelson I also update the Sequential Tree TaskFlow where you can indicate if a generator should exit on succes or on failure. This way you easily build a taskflow the chains them all together or exits on first success. This also solves the issue with leveraging the taskflow to generate the seed.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/default_processes/default_freespace_processes.h
```diff
@@ -49,28 +50,33 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+inline SequentialProcesses defaultFreespaceProcesses()
 {
+  SequentialProcesses sp;
+
   // Setup Interpolator
   auto interpolator = std::make_shared<SimpleMotionPlanner>("INTERPOLATOR");
   interpolator->plan_profiles["FREESPACE"] = std::make_shared<SimplePlannerDefaultPlanProfile>();
   auto interpolator_generator = std::make_shared<MotionPlannerProcessGenerator>(interpolator);
-  interpolator_generator->validators.emplace_back(&randomValidator);
+  sp.emplace_back(interpolator_generator, SequentialTaskType::CONDITIONAL_EXIT_ON_FAILURE);
```
mpowelson, line 61: So this is a bit backwards from what I was intending (though not necessarily incorrect). I was intending it to try one until it succeeded. Now if I understand correctly it will always run all 3 (in a successful case).

How would we handle a case like
- Always run initialize seed (which should succeed)
- Run trajopt. Return if succeeds
- Run OMPL.
- Run OMPL with different parameters
- etc

Levi-Armstrong: You  are correct, but you can easily create what you intended by switching out the task type for trajopt and ompl to CONDITIONAL_EXIT_ON_SUCCESS.

Levi-Armstrong: @mpowelson Does the name SequentialFailureTreeTaskFlow still fit? Should it be renamed to something more general like MotionPlanningTaskFlow?

mpowelson: Maybe just SequentialTaskFlow. You could add tasks that aren't motion planning specific. But it also isn't strictly a cascading error tree anymore.

Levi-Armstrong: Sound good, SequentialTaskFlow it is.

Levi-Armstrong:
Process Manager Example for a raster path:

image

image

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
Levi-Armstrong: @mpowelson Take a look at how I restructured the command language utilities and let me know what you think.

Levi-Armstrong:
@marrts So everything is now working all the way up to the tesseract_process_managers package now. The only think left to do is create the contact checker process generator for validating the trajectory after it has finished, but that can go in another PR if you have time to review the latest commits.

mpowelson:
> @marrts So everything is now working all the way up to the tesseract_process_managers package now. The only think left to do is create the contact checker process generator for validating the trajectory after it has finished, but that can go in another PR if you have time to review the latest commits.

So did you decide to add the contact checks as tasks? That seems logical now that you added the exit flags to the taskflow generators. Does each process still have the validators as well?

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/command_language_utils.h
mpowelson: That looks good. The unit tests for the filter functionality look a bit sparse. It only tests when it returns true. I imagine we want to test the default value and return false (and includes only the correct types). 

Should we add some extra information about what exactly those filters should do? e.g. They should return true when the instruction passed in should be included, and they should not throw. Including some premade ones would be convenient. I'm thinking one to include all planning instructions, all non-user-implemented, etc.

Levi-Armstrong:
> > @marrts So everything is now working all the way up to the tesseract_process_managers package now. The only think left to do is create the contact checker process generator for validating the trajectory after it has finished, but that can go in another PR if you have time to review the latest commits.
> 
> So did you decide to add the contact checks as tasks? That seems logical now that you added the exit flags to the taskflow generators. Does each process still have the validators as well?

That is what I am working on now. I expect that it will be a separate task but the validators still exist. I am think we maybe should remove them?

mpowelson:
@Levi-Armstrong Yeah you probably can. If you wanted, you could add a validator process generator that could dynamically create a bunch of parallel tasks for validators. But for now, a single validator for collision is probably sufficient.

mpowelson:
Bug report: If you try the freespace_manager_example going to a joint waypoint, it doesn't work. To replicate, just change freespaceExampleProgram() to this
```
inline CompositeInstruction freespaceExampleProgram()
{
  CompositeInstruction program(
      "freespace_composite", CompositeInstructionOrder::ORDERED, ManipulatorInfo("manipulator"));

  // Start Joint Position for the program
  std::vector<std::string> joint_names = { "joint_a1", "joint_a2", "joint_a3", "joint_a4",
                                           "joint_a5", "joint_a6", "joint_a7" };
  Waypoint wp1 = StateWaypoint(joint_names, Eigen::VectorXd::Zero(7));
  PlanInstruction start_instruction(wp1, PlanInstructionType::START);
  program.setStartInstruction(start_instruction);

  // Define target pose
  Waypoint wp2 = JointWaypoint(Eigen::VectorXd::Ones(7) * 0.1);
  PlanInstruction plan_f0(wp2, PlanInstructionType::FREESPACE, "DEFAULT");
  plan_f0.setDescription("freespace_motion");
  program.push_back(plan_f0);

  return program;
}
```

Here's the error: 
tesseract_process_managers_freespace_manager_example: /home/matthew/workspaces/test/src/tesseract/tesseract/tesseract_planning/tesseract_motion_planners/src/simple/step_generators/fixed_size_interpolation.cpp:51: tesseract_planning::CompositeInstruction tesseract_planning::fixedSizeJointInterpolation(const tesseract_planning::JointWaypoint&, const tesseract_planning::JointWaypoint&, const tesseract_planning::PlanInstruction&, const tesseract_planning::PlannerRequest&, const tesseract_planning::ManipulatorInfo&, int): Assertion `static_cast<long>(end.joint_names.size()) == end.size()' failed.

Levi-Armstrong:
Since you are using a joint waypoint you must set the joint names. The wp2 does not have joint names assigned to it. Add `wp2.joint_names = joint_names` and it should work.

mpowelson:
> Since you are using a joint waypoint you must set the joint names. The wp2 does not have joint names assigned to it. Add `wp2.joint_names = joint_names` and it should work.

Should we delete the constructor that just takes a vectorXd or at least add one that takes joint names as well? Right now I have to do
```
  auto tmp = JointWaypoint(Eigen::VectorXd::Ones(7) * 0.1);
  tmp.joint_names = joint_names;
  Waypoint wp2 = tmp;
```

Levi-Armstrong:
I will add a constructor that takes a joint names and eigen vector.

Levi-Armstrong:
@mpowelson Added the JointWaypoint constructor and fixed the raster example.

tesseract/tesseract_planning/tesseract_process_managers/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ add_library(${PROJECT_NAME} SHARED
     src/process_generators/motion_planner_process_generator.cpp
     src/process_managers/freespace_process_manager.cpp
     src/process_managers/raster_process_manager.cpp
-    src/taskflow_generators/sequential_failure_tree_taskflow.cpp)
+    src/taskflow_generators/sequential_taskflow.cpp)
 target_link_libraries(${PROJECT_NAME} PUBLIC
```
mpowelson, line 25: Should this be linking tesseract::tesseract_command_language?

Levi-Armstrong: Yes. It is in on my machine but may have added it over the weekend. I will push my changes soon.

Levi-Armstrong: pushed my changes

Levi-Armstrong:
Discrete and Continuous Contact Check Process Generators have been added.

image

mpowelson:
@Levi-Armstrong Perhaps we should be using [declaration.hpp](https://github.com/taskflow/taskflow/blob/master/taskflow/declarations.hpp) in the headers to avoid the long compile times.

Levi-Armstrong:
Iterative Spline Parameterization has been integrated.

image

mpowelson:
@Levi-Armstrong One comment, the freespace manager does not take into account whether or not a process is exiting because it was successful or because it failed. Any early exit is considered a failure. If I've got something like below then what I really want is for planning to return failed if interpolate failed, succeeded if the first collision check succeeded, failed if OMPL failed, and succeeded if the second collision check succeeded.

```
  // Setup Interpolator
  auto interpolator = std::make_shared<SimpleMotionPlanner>("INTERPOLATOR");
  interpolator->plan_profiles["FREESPACE"] = std::make_shared<SimplePlannerDefaultPlanProfile>();
  auto interpolator_generator = std::make_unique<MotionPlannerProcessGenerator>(interpolator);
  sp.emplace_back(std::move(interpolator_generator), SequentialTaskType::CONDITIONAL_EXIT_ON_FAILURE);

  {
    auto contact_check_generator = std::make_unique<DiscreteContactCheckProcessGenerator>();
    sp.emplace_back(std::move(contact_check_generator), SequentialTaskType::CONDITIONAL_EXIT_ON_SUCCESS);
  }

  // Setup OMPL
  auto ompl_planner = std::make_shared<OMPLMotionPlanner>();
  ompl_planner->problem_generator = &DefaultOMPLProblemGenerator;
  ompl_planner->plan_profiles["FREESPACE"] = std::make_shared<OMPLDefaultPlanProfile>();
  auto ompl_generator = std::make_unique<MotionPlannerProcessGenerator>(ompl_planner);
  sp.emplace_back(std::move(ompl_generator), SequentialTaskType::CONDITIONAL_EXIT_ON_FAILURE);

  {
    auto contact_check_generator = std::make_unique<DiscreteContactCheckProcessGenerator>();
    sp.emplace_back(std::move(contact_check_generator), SequentialTaskType::CONDITIONAL_EXIT_ON_SUCCESS);
  }
```

mpowelson:
Also, I would probably put the time parameterization task outside of the sequential taskflow. In order to do a planning structure that takes the first successful method, you would want that to exit succeeded and then run time parameterization.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/utils.h
```diff
@@ -147,6 +153,582 @@ inline std::vector<Waypoint> interpolate_waypoint(const Waypoint& start, const W
...
+                                    tesseract_collision::ContactRequest(tesseract_collision::ContactTestType::FIRST),
+                                bool verbose = false)
+{
+  bool found = false;
+
+  // Flatten results
+  std::vector<std::reference_wrapper<const Instruction>> mi = flatten(program, moveFilter);
+
+  contacts.reserve(mi.size());
+  for (std::size_t iStep = 0; iStep < mi.size(); ++iStep)
```
mpowelson, line 657: I think this needs to be `for (std::size_t iStep = 0; iStep < mi.size() - 1; ++iStep)` or else swp1 needs to just go away

tesseract/tesseract_planning/tesseract_process_managers/src/process_generators/discrete_contact_check_process_generator.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // Get state solver
+  tesseract_environment::StateSolver::Ptr state_solver = input.tesseract->getEnvironmentConst()->getStateSolver();
+  tesseract_collision::DiscreteContactManager::Ptr manager =
+      input.tesseract->getEnvironmentConst()->getDiscreteContactManager();
+  manager->setContactDistanceThreshold(contact_distance_);
+
+  const auto* ci = input.results->cast_const<CompositeInstruction>();
+  std::vector<tesseract_collision::ContactResultMap> contacts;
+  if (contactCheckProgram(contacts, *manager, *state_solver, *ci, longest_valid_segment_length_))
```
mpowelson, line 82: What would you think about adding `console_bridge::getLogLevel() == console_bridge::LogLevel::CONSOLE_BRIDGE_LOG_DEBUG` to set verbose?

mpowelson: `  if (contactCheckProgram(contacts, *manager, *state_solver, *ci, longest_valid_segment_length_, tesseract_collision::ContactRequest(tesseract_collision::ContactTestType::FIRST), console_bridge::getLogLevel() == console_bridge::LogLevel::CONSOLE_BRIDGE_LOG_DEBUG))`

mpowelson:
> Also, I would probably put the time parameterization task outside of the sequential taskflow. In order to do a planning structure that takes the first successful method, you would want that to exit succeeded and then run time parameterization.

Thinking about this more, perhaps the freespace manager takes a vector of taskflow generators. Then you could have similar to what MoveIt does with FixStartState tasks in the first, planning, in the second, and time param in the third

tesseract/tesseract_planning/tesseract_process_managers/src/process_generators/discrete_contact_check_process_generator.cpp
Levi-Armstrong: I guess we should maybe remove the verbose and then switch the messages to console messages debug. Then the developer can just set the verbosity in the using program to trigger these messages without having to pass it through function calls. What do you think?

mpowelson: Yeah that's fine. I imagine the compiler will optimize all of the loops away if it is disabled anyway.

Levi-Armstrong:
> > Also, I would probably put the time parameterization task outside of the sequential taskflow. In order to do a planning structure that takes the first successful method, you would want that to exit succeeded and then run time parameterization.
> 
> Thinking about this more, perhaps the freespace manager takes a vector of taskflow generators. Then you could have similar to what [MoveIt](https://moveit.ros.org/documentation/concepts/) does with FixStartState tasks in the first, planning, in the second, and time param in the third

Sound good. I will look at doing something similar to how I did the process generators so you can attach a task flow to either the done process or failure process.

mpowelson:
> > > Also, I would probably put the time parameterization task outside of the sequential taskflow. In order to do a planning structure that takes the first successful method, you would want that to exit succeeded and then run time parameterization.
> > 
> > 
> > Thinking about this more, perhaps the freespace manager takes a vector of taskflow generators. Then you could have similar to what [MoveIt](https://moveit.ros.org/documentation/concepts/) does with FixStartState tasks in the first, planning, in the second, and time param in the third
> 
> Sound good. I will look at doing something similar to how I did the process generators so you can attach a task flow to either the done process or failure process.

Ok. Let me know if/when you do that and I can test it. FixStartStateBounds and FixStartStateCollision tasks would also be valuable since those are somewhat common problems. Both could potentially be a specialized trajopt call, or we could look at grabbing whatever MoveIt has. The path constraints and workspace bounds don't seem as useful to me, but maybe they are.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/utils.h
Levi-Armstrong: Since this is discrete it needs to handle the last state in the trajectory. There may be a cleaner way to handle this but I am doing by the logic highlighted below in the function.

```
    double dist = -1;
    if (iStep < mi.size() - 1)
      dist = (swp1->position - swp0->position).norm();

    if (dist > 0 && dist > longest_valid_segment_length)
    {
    }
    else
    {
      ... This just checks iStep for collision
    }
```

mpowelson: Will line 660 not segfault here?

mpowelson: I did not look into where, but I was getting an out of bounds iterator error in this function.

Levi-Armstrong: You are correct. I am fixing it now.

Levi-Armstrong:
I believe I have fixed the indexing out of bounds issue. I also have added a trajectory interrogator and trajectory player to tesseract_visualization.

Levi-Armstrong:
Added graph taskflow producing the same results as the sequential taskflow but more flexible. 
image

Levi-Armstrong:
@mpowelson GraphTaskflow is available.

Levi-Armstrong:
Do you think we should remove the sequential taskflow and just create a helper function that takes a vector of ProcessGenerators?

mpowelson:
Also, this PR is getting difficult to review. I would recommend that we choose some point to merge in and then continue opening new feature pull requests against feature/CommandLanguage

tesseract/tesseract_planning/tesseract_process_managers/src/examples/freespace_manager_graph_example.cpp
```diff
@@ -0,0 +1,205 @@
...
+  /////////////////
+  /// Add Edges ///
+  /////////////////
+  graph->addEdge(interpolator_idx,
+                 GraphTaskflow::SourceChannel::ON_SUCCESS,
+                 ompl_idx,
+                 GraphTaskflow::DestinationChannel::PROCESS_NODE);
+  graph->addEdge(interpolator_idx,
+                 GraphTaskflow::SourceChannel::ON_FAILURE,
+                 -1,
```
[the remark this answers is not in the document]
Levi-Armstrong: It just needs an integer, but is not used. Should I create an overload function that does not take a node? It is only allowed if your destination is either error callback or done callback.

mpowelson:
> ProcessGenerators

Either way. It might be useful to have some function (either a helper or a class method) to add a vector of nodes. I'm thinking

```

addSequentialNodes(std::vector<SequentialProcesses>)
addSequentialNodes(std::vector<SequentialProcesses>, int success_node, int failure node)
```

Then you could easily add a couple of sequential sections and then just connect up the edges.

Levi-Armstrong:
Yea, I am good with merging now.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/taskflow_generators/graph_taskflow.h
```diff
@@ -0,0 +1,110 @@
...
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
+
+#include <tesseract_process_managers/taskflow_generator.h>
+#include <tesseract_process_managers/process_generator.h>
+
+namespace tesseract_planning
+{
+/** @brief This class generates taskflows for a sequential failure tree. Each process is executed in order until one
+ * succeeds. Between each process, the validator tasks are executed (if not empty). For a process to succeed, the
+ * process itself must succeed and all of the validators must succeed*/
```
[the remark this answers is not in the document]
Levi-Armstrong: fixed

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+void GraphTaskflow::addEdge(int src, SourceChannel src_channel, int dest, DestinationChannel dest_channel)
+{
+  Edge e;
+  e.src_channel = src_channel;
+  e.dest = dest;
+  e.dest_channel = dest_channel;
+
+  nodes_[static_cast<std::size_t>(src)].edges.push_back(e);
```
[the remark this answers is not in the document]
Levi-Armstrong: done

PR #346, opened 2020-08-14
Add clone method to Tesseract

mpowelson:
I tried just adding below to clone, but that does not work because the links are not in the graph when the joints are added. 

```
SceneGraph::Ptr cloned_graph = std::make_shared<SceneGraph>();
cloned_graph->insertSceneGraph(*this, getOutboundJoints(getRoot())[0], "");
return cloned_graph;
```


Levi-Armstrong:
@mpowelson Could you add a test that instead of cloning creates an empty scene graph and call the insertSceneGraph and run through the same check?

mpowelson:
> @mpowelson Could you add a test that instead of cloning creates an empty scene graph and call the insertSceneGraph and run through the same check?

I tried to do this in the clone method, but it fails because the joints that are added reference links that have not been added yet. That's why I switched to just inserting everything. While that function was copied from the environment, I'm not sure it really does what we want.

mpowelson:
@Levi-Armstrong Apparently I need changes in your PR. So we can just wait until that is merged and then merge this into feature/command_language.

Levi-Armstrong:
These changes have been included in PR #349.

PR #348, opened 2020-08-15
Fix/Add clearing of graph and sequential taskflow

Levi-Armstrong:


tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generator.h
```diff
@@ -46,7 +46,12 @@ class ProcessGenerator
 public:
   using UPtr = std::unique_ptr<ProcessGenerator>;
 
+  ProcessGenerator() = default;
   virtual ~ProcessGenerator() = default;
+  ProcessGenerator(const ProcessGenerator&) = delete;
+  ProcessGenerator& operator=(const ProcessGenerator&) = delete;
+  ProcessGenerator(ProcessGenerator&&) = delete;
+  ProcessGenerator& operator=(ProcessGenerator&&) = delete;
```
mpowelson, line 54: Why can't these be copied?

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/taskflow_generator.h
```diff
@@ -43,7 +43,12 @@ class TaskflowGenerator
 public:
   using UPtr = std::unique_ptr<TaskflowGenerator>;
 
+  TaskflowGenerator() = default;
   virtual ~TaskflowGenerator() = default;
+  TaskflowGenerator(const TaskflowGenerator&) = delete;
+  TaskflowGenerator& operator=(const TaskflowGenerator&) = delete;
+  TaskflowGenerator(TaskflowGenerator&&) = delete;
+  TaskflowGenerator& operator=(TaskflowGenerator&&) = delete;
```
mpowelson, line 51: This one as well. I don't know why you would want to copy it, but it seems like it should be safe to do so.

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -192,16 +192,25 @@ tf::Taskflow& GraphTaskflow::generateTaskflow(ProcessInput input,
 
 void GraphTaskflow::abort()
 {
+  abort_ = true;
```
mpowelson, line 195: Should generateTaskflow set abort to false?

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/sequential_taskflow.cpp
```diff
@@ -139,16 +139,25 @@ tf::Taskflow& SequentialTaskflow::generateTaskflow(ProcessInput input,
 
 void SequentialTaskflow::abort()
 {
+  abort_ = true;
```
mpowelson, line 142: same here

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generator.h
Levi-Armstrong: It is good to delete the base class copy constructors because copying is slicing. Also I believe adding an atomic implicitly deletes the copy constructor and maybe the move constructor.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/taskflow_generator.h
Levi-Armstrong: same as above.

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -192,16 +192,25 @@ tf::Taskflow& GraphTaskflow::generateTaskflow(ProcessInput input,
...
   for (auto& node : nodes_)
     node.process->setAbort(true);
 }
 
 void GraphTaskflow::reset()
 {
+  abort_ = false;
```
mpowelson, line 202: This can segfault. 
```
 for (auto& node : nodes_)
    if (node.process)
       node.process->setAbort(false);

```

mpowelson: Perhaps this check should go in the addNode. It came about because I was doing this.
```
  auto contact_check_generator = std::make_unique<DiscreteContactCheckProcessGenerator>();
  int contact_check_idx_1 = graph->addNode(std::move(contact_check_generator), GraphTaskflow::NodeType::CONDITIONAL);
  int contact_check_idx_2 = graph->addNode(std::move(contact_check_generator), GraphTaskflow::NodeType::CONDITIONAL);
```

It seems odd conceptually that the same generator can't be passed into multiple nodes.

Levi-Armstrong: Now that I think about it, is it ever possible that it would be a null pointer. It is a unique pointer which this class owns so if it is in the vector of nodes it should have a valid pointer. If it does not, then this class deleted it which should not happen.

mpowelson: Is there some reason addNode has to take a unique pointer? Or could it take a reference? The Node stores a ProcessGenerator::UPtr, but there shouldn't be any reason multiple nodes couldn't use the same ProcessGenerator. That is sort of the purpose of the generators.

Levi-Armstrong: > Perhaps this check should go in the addNode. It came about because I was doing this.
> 
> ```
>   auto contact_check_generator = std::make_unique<DiscreteContactCheckProcessGenerator>();
>   int contact_check_idx_1 = graph->addNode(std::move(contact_check_generator), GraphTaskflow::NodeType::CONDITIONAL);
>   int contact_check_idx_2 = graph->addNode(std::move(contact_check_generator), GraphTaskflow::NodeType::CONDITIONAL);
> ```
> 
> It seems odd conceptually that the same generator can't be passed into multiple nodes.

The main reason was the ability to cancel things. If they were added to two different taskflows and you called cancel in one it would kill the processes in the other taskflow. We could switch it back to Unique Pointers and leave it up to the developer to make sure they do not do this.

mpowelson:
Is this safe for a Graph Taskflow? Specifically how there are 2 paths to ISP. It should be right?

image

Levi-Armstrong:
> Is this safe for a Graph Taskflow? Specifically how there are 2 paths to ISP. It should be right?
> 
> image
> 
> I think I should be able to use the same contact checker too actually since the same ProcessInput is going to all of them. Edit: Nevermind, I need it to go to different places after each collision check, so they have to be different tasks

It should be good. No two process will be operating on the same data at the same time.

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
Levi-Armstrong: Now that things are working we can look at restructuring to see what should be unique. It was easier to do this to start to insure that I did not run into race conditions while trouble shooting the how the data flows through everything.

Levi-Armstrong: @mpowelson If we could change the ProcessGenerator to where it returns an object that gets stored at the higher level this would allows for the same process generator to be used.

mpowelson: Maybe what should happen is that the process generators should clone and store all of the information they need when generate is called. e.g. clone the planner. As for the abort, I think it is reasonable that any process created from the same generator shares an abort flag, but that's just my opinion.

PR #349, opened 2020-08-17
Add scene graph clone and environment clone

Levi-Armstrong:
@mpowelson I improved the scene graph insertSceneGraph along with add a new command to the environment AddSceneGraphCommand where now every environment starts with one command which is the AddSceneGraphCommand. I will need to create a message for SceneGraph to complete tesseract_ros. Now should have everything in the command history to fully recreate an environment.


Levi-Armstrong:
Also I tried rebasing yours on the latest, but in the end it was easier for me to cherry-pick the commits.

Levi-Armstrong:
Depends on PR #348

tesseract/tesseract_environment/include/tesseract_environment/core/environment.h
```diff
@@ -64,12 +64,64 @@ class Environment
    *
    * Inherited class must call the protected create function
    *
-   * @param scene_graph
-   * @return
+   * @param scene_graph The scene graph to initialize the environment.
+   * @return True if successful, otherwise false
    */
-  virtual bool init(tesseract_scene_graph::SceneGraph::Ptr scene_graph) = 0;
+  template <typename S>
```
mpowelson, line 70: Should the documentation state what should be passed into this template? ie, a state solver?

mpowelson: Also, do you want to do an explicit instantiation of a KDL state solver (maybe you already do).

mpowelson:
This shouldn't be a breaking change for most people, but the biggest change is that rather than having a custom KDL environment class, the base is now templated by state solver. To migrate, make changes similar to what was done in tesseract.cpp
```
#include <tesseract_environment/kdl/kdl_env.h>

environment_ = std::make_shared<tesseract_environment::KDLEnv>();
environment_->init(scene_graph)
```
becomes
```
#include <tesseract_environment/kdl/kdl_state_solver.h>

environment_ = std::make_shared<tesseract_environment::Environment>();
environment_->init<tesseract_environment::KDLStateSolver>(scene_graph)
```

PR #352, opened 2020-08-18
Add debugging information when planning fails due to collision

mpowelson:


mpowelson:
> Should we add a print out here? Maybe cache the latest one and at the end if none are found to be collision free it prints them.
> 
> https://github.com/ros-industrial-consortium/tesseract/blob/567ab08a81d971bc4a5185750d98474ceebceef2/tesseract/tesseract_planning/tesseract_motion_planners/src/ompl/profile/ompl_default_plan_profile.cpp#L154

Yeah we can. More broadly, I was wondering if we want to have some sort of logging about how different tasks executed in tesseract_process_managers. Perhaps some record of whether or not a task suceeded or failed and error messages along the way. That way if a planning pipeline fails there is something to print out. That might be a good use of the observer interface.

Levi-Armstrong:
The observer looks good. I saw some code for it; did you ever get anything working?

mpowelson:
> The observer looks good. I saw some code for it; did you ever get anything working?

I played with the example he had for the simple observer that just said what task you were in. I don't know if I pushed it or not though.

Levi-Armstrong:
Were you going to add the debugging to the Cartesian or do you want to merge it as is?

mpowelson:
> Were you going to add the debugging to the Cartesian or do you want to merge it as is?

Should be done now.

Levi-Armstrong:
It looks like there is some compiler warnings causing CI to fail.

PR #353, opened 2020-08-19
Expose velocity and acceleration scaling factors in process generators

mpowelson:


mpowelson:
I was gettting some odd results when setting this to less than 1, but that might be on my execution side. Regardless, I think we want these exposed.

Levi-Armstrong:
This ready to merge but waiting on PR #352

PR #355, opened 2020-08-20
Add simple process manager and planner profile remapping

Levi-Armstrong:
This PR adds two things. The first is a simple process manager which is very similar to freespace process manager which maybe deprecated in the near future. This is currently be used in the planning server to support a simple freespace process manager and simple cartesian process manager.

The second adds the ability to provide profile mapping for specific planners. This way in the command language you can have one profile for say Cartesian, then in this mapping you could provide a planner mapping to a specific profile. For example you could map the Cartesian profile for Descartes to DescartesToolZFree45 which has 45 degree sampling and for TrajOpt you could map it to a more specific or not provide a mapping.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
@@ -33,9 +33,18 @@
 
 namespace tesseract_planning
 {
+/**
+ * This used to store planner specific profile mapping with the request
+ *
+ * For example say you have a profile named Raster in your command language. Say you have multiple Raster profiles
+ * for descartes {Raster, Raster1, Rester2}. This allows you to remap the meaning of Raster in the command language to
+ * say Raster2 for the specific planner Descartes by Map<Descartes, Map<Raster, Raster1>>.
+ */
+using PlannerProfileRemapping = std::unordered_map<std::string, std::unordered_map<std::string, std::string>>;
+
 struct PlannerRequest
 {
-  std::string name;                                    /**< @brief The name of the planner to use */
+  std::string name;                                    /**< @brief The name of the process manager to use */
   tesseract::Tesseract::ConstPtr tesseract;            /**< @brief Tesseract */
   tesseract_environment::EnvState::ConstPtr env_state; /**< @brief The start state to use for planning */
 
@@ -49,6 +58,18 @@ struct PlannerRequest
    * CompositeInstruction of MoveInstructions.
    */
   CompositeInstruction seed;
+
+  /**
+   * @brief This allows the remapping of the Plan Profile identified in the command language to a specific profile for a
+   * given motion planner.
+   */
+  PlannerProfileRemapping plan_profile_remapping;
+
+  /**
+   * @brief This allows the remapping of the Composite Profile identified in the command language to a specific profile
+   * for a given motion planner.
+   */
+  PlannerProfileRemapping composite_profile_remapping;
 
   /**
    * @brief data Planner specific data. For planners included in Tesseract_planning this is the planner problem that
--- a/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
+++ b/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
@@ -32,6 +32,8 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <tesseract_command_language/command_language.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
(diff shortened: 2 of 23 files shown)


tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/simple_process_manager.cpp
```diff
@@ -0,0 +1,97 @@
...
+    CONSOLE_BRIDGE_logError("ProcessInput Invalid: input.instructions should have a start instruction");
+    return false;
+  }
+
+  // Create the dependency graph
+  input.instruction->print("Generating Taskflow for: ");
+  auto task = taskflow_
+                  .composed_of(taskflow_generator_->generateTaskflow(
+                      input, [this]() { successCallback(); }, [this]() { failureCallback(); }))
+                  .name("freespace");
```
mpowelson, line 52: Perhaps default name should be "simple"?

tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/simple_process_manager.cpp
```diff
@@ -0,0 +1,97 @@
...
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
+
+#include <tesseract_process_managers/process_managers/simple_process_manager.h>
+
+#include <tesseract_process_managers/process_generators/motion_planner_process_generator.h>
+#include <tesseract_process_managers/taskflow_generators/sequential_taskflow.h>
+
+#include <tesseract_motion_planners/trajopt/profile/trajopt_default_plan_profile.h>
+#include <tesseract_motion_planners/trajopt/profile/trajopt_default_composite_profile.h>
+#include <tesseract_motion_planners/trajopt/problem_generators/default_problem_generator.h>
```
mpowelson, line 16: These should be able to go.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
```diff
@@ -33,9 +33,18 @@
 
 namespace tesseract_planning
 {
+/**
+ * This used to store planner specific profile mapping with the request
+ *
+ * For example say you have a profile named Raster in your command language. Say you have multiple Raster profiles
+ * for descartes {Raster, Raster1, Rester2}. This allows you to remap the meaning of Raster in the command languate to
```
mpowelson, line 40: typo `language`

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
```diff
@@ -73,6 +90,12 @@ struct ProcessInput
   /** @brief Global Manipulator Information */
   const ManipulatorInfo& manip_info;
 
+  /** @brief Planner Plan Profile mapping */
+  const PlannerProfileMapping& plan_profile_mapping;
+
+  /** @brief Planner Composite Profile mapping */
+  const PlannerProfileMapping& composite_profile_mapping;
+
```
mpowelson, line 98: I'm not sure I would put this in the process input.  While I understand wanting to put it in the Motion Planning request so we don't have to have multiple copies of the planner with different mappings, I think I would prefer that each generator has a set of mappings and we keep the ProcessInput simple and independent of the processes it is passed into.

mpowelson: This also means that if these mappings were loaded from a file, that would happen here at the process manager rather than at the caller.

tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/simple_process_manager.cpp
Levi-Armstrong: removed

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/types.h
Levi-Armstrong: fixed

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
Levi-Armstrong: Based on our discussion, I renamed things and updated the documentation in the ProcessInput class.

tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/simple_process_manager.cpp
Levi-Armstrong: It has been renamed.

PR #363, opened 2020-08-25
Add Profile to ISP time parameterization

mpowelson:


Levi-Armstrong:
This can now be rebase.

mpowelson:
@Levi-Armstrong I've also got [this branch](https://github.com/mpowelson/tesseract/tree/archive/isp_plan_profile) where I started to add this as a plan profile like we discussed. However, I realized that there was an issue at the transitions where there will be a jump unless the two waypoints are on top of each other (since they will be in different locations with no time between them).

mpowelson:
@Levi-Armstrong What if we modified ISP to accept a vector of scaling factors? It seems that ISP supports changing the max velocity per point if we just set it [here](https://github.com/ros-industrial-consortium/tesseract/blob/519f05bb9e0a01911a58e2b1445c9c1345a7de5d/tesseract/tesseract_planning/tesseract_time_parameterization/src/iterative_spline_parameterization.cpp#L351)

Levi-Armstrong:
Works for me.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/iterative_spline_parameterization_process_generator.h
```diff
@@ -36,6 +36,27 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+  {
+  }
+
+  /** @brief max_velocity_scaling_factor The max velocity scaling factor passed to the solver */
+  double max_velocity_scaling_factor = 1.0;
+
+  /** @brief max_velocity_scaling_factor The max acceleration scaling factor passed to the solver */
+  double max_acceleration_scaling_factor = 1.0;
+};
+using IterativeSplineParameterizationPlanProfileMap =
```
Levi-Armstrong, line 57: Should this be ...CompositeProfileMap?

tesseract/tesseract_planning/tesseract_process_managers/src/process_generators/iterative_spline_parameterization_process_generator.cpp
```diff
@@ -69,11 +72,35 @@ int IterativeSplineParameterizationProcessGenerator::conditionalProcess(ProcessI
   const ManipulatorInfo& manip_info = ci->getManipulatorInfo();
   const auto fwd_kin = input.tesseract->getFwdKinematicsManagerConst()->getFwdKinematicSolver(manip_info.manipulator);
 
+  // Get Composite profile
+  std::string profile = ci->getProfile();
+  if (profile.empty())
+    profile = "DEFAULT";
+
+  // Check for remapping of profile
+  auto remap = input.plan_profile_remapping.find(name_);
```
Levi-Armstrong, line 81: Should this be pulling from the composite_profile remapping?

Levi-Armstrong:
@mpowelson I rebased feature/CommandLanguage on master. If you rebase on feature/CommandLanguage the commits should go away.

mpowelson:
@Levi-Armstrong I think I will add a composite profile map and a plan profile map. I'll initialize all of the params to the composite profile and then override individual instructions' params if a plan profile is available for it.

This of course assumes that the comment I made yesterday works and I can set the parameters on a point by point basis.

mpowelson:
@Levi-Armstrong Ready for review. The ISP algorithm did not allow changing scaling on a trajectory point basis (only on a per joint basis), so I added it. 

It's also worth noting that there is a lot of math in there that could potentially be vectorized someday. However, it may be more worthwhile to put that time into integrating time parameterization into TrajOpt. I haven't profiled it, but I imagine it is still pretty fast.

tesseract/tesseract_planning/tesseract_time_parameterization/src/iterative_spline_parameterization.cpp
```diff
@@ -340,33 +405,33 @@ bool IterativeSplineParameterization::compute(const std::vector<std::reference_w
     t2[j].accelerations_[num_points - 1] = t2[j].final_acceleration_;
 
     // Set bounds based on inputs
-    t2[j].max_velocity_ = max_velocity[j];
-    t2[j].min_velocity_ = -max_velocity[j];
-    t2[j].max_velocity_ *= velocity_scaling_factor;
-    t2[j].min_velocity_ *= velocity_scaling_factor;
+    t2[j].max_velocity_ = Eigen::VectorXd::Ones(static_cast<Eigen::Index>(num_points)) * max_velocity[j];
+    t2[j].min_velocity_ = Eigen::VectorXd::Ones(static_cast<Eigen::Index>(num_points)) * -max_velocity[j];
+    t2[j].max_velocity_.array() *= velocity_scaling_factor.array();
```
Levi-Armstrong, line 410: Does a VectorXd not have cwiseProduct? I am not sure which performs better.

```
t2[j].max_velocity_ = t2[j].max_velocity_.cwiseProduct(velocity_scaling_factor);
```

Levi-Armstrong: I found this in the documentation so there is no performance difference. I will leave it up to you which you prefer. 

```
The main difference between the two API is that the one based on cwise* methods returns an expression in the matrix world, while the second one (based on .array()) returns an array expression. Recall that .array() has no cost, it only changes the available API and interpretation of the data.
```

mpowelson: I'd probably leave it like this as it seems a bit more concise

Issue #377, opened 2020-09-07
Is there a quick way to get the collisions that are allowed with a specific link?


mpowelson:
Specifically, I want to get a vector of all of the links that are allowed to collide with a manipulator group. Is there a quick way to do this?

Levi-Armstrong:
I do not believe so. The allowed collision matrix class could be modified to store this information

PR #382, opened 2020-09-08
Feature/get allowed collisions

mpowelson:
Addresses #377 . Depends on #379


mpowelson:
Rebased

Levi-Armstrong:
As a separate utility it could be a time consuming call if this was made repeatedly. Not sure how widely this will be used, is there a reason not to store this in the allowed collision matrix and provide getters for the information?

mpowelson:
Well I thought about how to store it, and I figured it would basically become some graph structure like the scene graph. I also thought about storing an unordered_map<string, vector<string>> in the acm. I'll do whatever you think is best. I just didn't know if this would actually be a common thing people needed to do.

mpowelson:
But yeah. I'm just processing this once at startup and storing it at the moment.

tesseract/tesseract_scene_graph/test/CMakeLists.txt
```diff
@@ -11,3 +11,13 @@ target_code_coverage(${PROJECT_NAME}_unit ALL EXCLUDE ${COVERAGE_EXCLUDE} ENABLE
 add_gtest_discover_tests(${PROJECT_NAME}_unit)
 add_dependencies(${PROJECT_NAME}_unit ${PROJECT_NAME})
 add_dependencies(run_tests ${PROJECT_NAME}_unit)
+
+add_executable(${PROJECT_NAME}_utils_unit utils_unit.cpp)
+target_link_libraries(${PROJECT_NAME}_utils_unit PRIVATE GTest::GTest GTest::Main ${PROJECT_NAME} tesseract::tesseract_support tesseract::tesseract_geometry)
+tesseract_target_compile_options(${PROJECT_NAME}_utils_unit PRIVATE)
+tesseract_clang_tidy(${PROJECT_NAME}_utils_unit)
+tesseract_code_coverage(${PROJECT_NAME}_utils_unit ALL EXCLUDE ${COVERAGE_EXCLUDE})
+target_include_directories(${PROJECT_NAME}_utils_unit PRIVATE ${orocos_kdl_INCLUDE_DIRS})
```
Levi-Armstrong, line 20: The above macros have changed to leveraging the one in cmake_common_scripts. I would take a look at another package for the new syntax.

Levi-Armstrong:
I think storing it as an unordered_map with a vector of string is a good approach to keep it simple. I could then leverage this in the Rviz widget.

mpowelson:
Should this always get populated or you have to request it? Like you said, it could be relatively expensive.

Levi-Armstrong:
I think every time a pair is added or removed it would update the information. This will not affect the lookup during normal operation so should not see a penalty. Only would affect add/removal of objects. Do you see a case where we would be constantly adding and remove objects from the allowed collision matrix? If so, I would say we leave the way you have it.

mpowelson:
> I think every time a pair is added or removed it would update the information. This will not affect the lookup during normal operation so should not see a penalty. Only would affect add/removal of objects. Do you see a case where we would be constantly adding and remove objects from the allowed collision matrix? If so, I would say we leave the way you have it.

Well my usecase is that every time I plan I need to add a link and a bunch of things to the ACM. However the environment I use is a clone that isn't reused. So in that case it would just be wasted overhead. Perhaps I can add a getter that on the first time runs the utility and stores the results. Then I can have a bool to keep track of when it needs to be updated. Essentially, just cache the results.

tesseract/tesseract_scene_graph/test/CMakeLists.txt
Levi-Armstrong: > > I think every time a pair is added or removed it would update the information. This will not affect the lookup during normal operation so should not see a penalty. Only would affect add/removal of objects. Do you see a case where we would be constantly adding and remove objects from the allowed collision matrix? If so, I would say we leave the way you have it.
> 
> Well my usecase is that every time I plan I need to add a link and a bunch of things to the ACM. However the environment I use is a clone that isn't reused. So in that case it would just be wasted overhead. Perhaps I can add a getter that on the first time runs the utility and stores the results. Then I can have a bool to keep track of when it needs to be updated. Essentially, just cache the results.



> > I think every time a pair is added or removed it would update the information. This will not affect the lookup during normal operation so should not see a penalty. Only would affect add/removal of objects. Do you see a case where we would be constantly adding and remove objects from the allowed collision matrix? If so, I would say we leave the way you have it.
> 
> Well my usecase is that every time I plan I need to add a link and a bunch of things to the ACM. However the environment I use is a clone that isn't reused. So in that case it would just be wasted overhead. Perhaps I can add a getter that on the first time runs the utility and stores the results. Then I can have a bool to keep track of when it needs to be updated. Essentially, just cache the results.

I think that is a good middle ground for now.

Levi-Armstrong:
I believe this is good to go, just needs to updated to the latest.

mpowelson:
> I believe this is good to go, just needs to updated to the latest.

I just updated it. If you want to merge it in its current form, I'm onboard. Otherwise, I'm not sure when/if I will get to making the cached version we discussed above. I have been using this for several months without problems.

PR #384, opened 2020-09-14
Add global raster variant

Levi-Armstrong:
This add a global raster taskflow which will run descartes on the whole raster followed by trajopt at the individual raster and transitions. This helps with limiting the number of configuration flips between rasters.

image

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_default_plan_profile.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_default_plan_profile.h
@@ -37,7 +37,8 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_planning
 {
-class SimplePlannerDefaultPlanProfile : public SimplePlannerPlanProfile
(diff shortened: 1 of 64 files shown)


Levi-Armstrong:
@marrts Currently this will not be useful until the LVS in integrated to limit the number waypoints to prevent running out of memory with Descartes.

Levi-Armstrong:
@mpowelson Just want to let you know I had to change how the ProcessInput function to support global planning first in the raster planners. Before it was storing pointer into the results data structure for start and end instruction modification, but when using a global planner first these were getting overridden. In order to solve the issues it now stores the index sequence need to access the information from the topmost instruction.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/simple/profile/simple_planner_default_plan_profile.h
```diff
@@ -37,7 +37,8 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_planning
 {
-class SimplePlannerDefaultPlanProfile : public SimplePlannerPlanProfile
+class DEPRECATED("Please use SimplePlannerLVSDefaultPlanProfile") SimplePlannerDefaultPlanProfile
+  : public SimplePlannerPlanProfile
```
mpowelson, line 41: Is SimplePlannerLVSDefaultPlanProfile coming in a future PR?

PR #385, opened 2020-09-15
Feature/lvs interpolation

marrts:
LVS interpolation

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/test/simple_planner_lvs_interpolation.cpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/test/simple_planner_lvs_interpolation.cpp
@@ -88,7 +88,7 @@ protected:
   }
 };
 
-TEST_F(TesseractPlanningSimplePlannerLVSInterpolationUnit, JointJoint_JointInterpolation)  // NOLINT
+TEST_F(TesseractPlanningSimplePlannerLVSInterpolationUnit, InterpolateStateWaypoint_JointJoint_Freespace)  // NOLINT
 {
   PlannerRequest request;
   request.tesseract = tesseract_ptr_;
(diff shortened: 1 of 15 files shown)


Levi-Armstrong:
Before it is merged, has this been tested out on the demo workcell?

marrts:
No, I haven't tested it on the workcell yet.

marrts:
Tested on workcell, it ran with RASTER_FT_PLANNER_NAME, but not global. Also I couldn't get the plotTrajectory to work so I couldn't see the individual points, but I did seem to notice a difference in behavior just from watching the motions.

marrts:
Tested and working on the tesseract workcell example now, there's a PR against that one too.

Levi-Armstrong (approved):

tesseract/tesseract_planning/tesseract_motion_planners/test/simple_planner_lvs_interpolation.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  auto composite = LVSInterpolateStateWaypoint(wp1, wp2, instr, request, ManipulatorInfo(), 3.14, 0.5, 1.57, 5);
+  for (const auto& c : composite)
+  {
+    EXPECT_TRUE(isMoveInstruction(c));
+    EXPECT_TRUE(isStateWaypoint(c.cast_const<MoveInstruction>()->getWaypoint()));
+  }
+  const auto* mi = composite.back().cast_const<MoveInstruction>();
+  EXPECT_TRUE(wp2.isApprox(mi->getWaypoint().cast_const<StateWaypoint>()->position, 1e-5));
+}
```
mpowelson, line 107: Sorry to be nit picky, but these unit tests don't test the most important features of these functions - that the LVS logic works correctly and that the interpolation is happening correctly. I bring this up because I was having an issue when switching to use this code. When looking at the tests here, they don't really guarantee that it is working at all. Has this code been used on some project so that we are reasonably certain it is doing what it should be doing?

Levi-Armstrong: @marrts Can you take a look at this?

mpowelson: Specifically, interpolation seems to be working fine. However on a joint joint freespace moves I seem to always get 11 points. Again, it could be my issue. I just don't know for sure since I don't think these tests test that.

marrts: I'll take a look at this, I think I've only really tested with cartesian to cartesian in an actual planning environment.

marrts: See #398

PR #387, opened 2020-09-22
Fix descartes handeling of freespace plan types

Levi-Armstrong:
Also improves logging for the checkKinematics functions.

### DIFF
--- a/tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
+++ b/tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
@@ -58,40 +58,85 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
 {
   // Check name
   if (fwd_kin->getName() != inv_kin->getName())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator names do not match!");
     return false;
+  }
 
   // Check if number of joints
   if (fwd_kin->numJoints() != inv_kin->numJoints())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator number of joints do not match!");
     return false;
+  }
 
   // Check link names
-  if (fwd_kin->getLinkNames().size() != inv_kin->getLinkNames().size() ||
-      !std::equal(fwd_kin->getLinkNames().begin(), fwd_kin->getLinkNames().end(), inv_kin->getLinkNames().begin()))
+  std::vector<std::string> l1 = fwd_kin->getLinkNames();
+  std::vector<std::string> l2 = inv_kin->getLinkNames();
+  std::sort(l1.begin(), l1.end());
+  std::sort(l2.begin(), l2.end());
+  if (l1.size() != l1.size())
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator link names size do not match!");
     return false;
+  }
+
+  if (!std::equal(l1.begin(), l1.end(), l2.begin()))
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator link names do not match!");
+    return false;
+  }
 
   // Check active link names
-  if (fwd_kin->getActiveLinkNames().size() != inv_kin->getActiveLinkNames().size() ||
-      !std::equal(fwd_kin->getActiveLinkNames().begin(),
-                  fwd_kin->getActiveLinkNames().end(),
-                  inv_kin->getActiveLinkNames().begin()))
+  l1 = fwd_kin->getActiveLinkNames();
+  l2 = inv_kin->getActiveLinkNames();
+  std::sort(l1.begin(), l1.end());
+  std::sort(l2.begin(), l2.end());
+  if (l1.size() != l2.size())
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator active link names size do not match!");
     return false;
+  }
+
+  if (!std::equal(l1.begin(), l1.end(), l2.begin()))
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator active link names do not match!");
+    return false;
+  }
 
   // Check joint names
-  if (fwd_kin->getJointNames().size() != inv_kin->getJointNames().size() ||
-      !std::equal(fwd_kin->getJointNames().begin(), fwd_kin->getJointNames().end(), inv_kin->getJointNames().begin()))
+  if (fwd_kin->getJointNames().size() != inv_kin->getJointNames().size())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator joint names size do not match!");
     return false;
+  }
+
+  if (!std::equal(fwd_kin->getJointNames().begin(), fwd_kin->getJointNames().end(), inv_kin->getJointNames().begin()))
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator joint names do not match!");
+    return false;
+  }
 
   // Check joint limits
   if (!fwd_kin->getLimits().joint_limits.isApprox(inv_kin->getLimits().joint_limits, tol))
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator joint limits do not match within tolerance %f!", tol);
     return false;
+  }
 
   // Check velocity limits
   if (!fwd_kin->getLimits().velocity_limits.isApprox(inv_kin->getLimits().velocity_limits, tol))
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator velocity limits do not match within tolerance %f!", tol);
     return false;
+  }
 
   // Check acceleration limits
   if (!fwd_kin->getLimits().acceleration_limits.isApprox(inv_kin->getLimits().acceleration_limits, tol))
+  {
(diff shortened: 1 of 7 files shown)


tesseract/tesseract_visualization/src/trajectory_interpolator.cpp
```diff
@@ -61,7 +61,7 @@ TrajectoryInterpolator::TrajectoryInterpolator(tesseract_planning::CompositeInst
 
     // It is possible for sub composites to start back from zero, this accounts for it
     if (current_time < last_time)
-      last_time = current_time;
```
mpowelson, line 64: This is correct?

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -58,40 +58,85 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
...
   if (fwd_kin->getName() != inv_kin->getName())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator names do not match!");
     return false;
+  }
 
   // Check if number of joints
   if (fwd_kin->numJoints() != inv_kin->numJoints())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator number of joints do not match!");
```
mpowelson, line 69: Maybe list them for convenience?
`CONSOLE_BRIDGE_logError("checkKinematics: Manipulator number of joints do not match! Fwd: %d, Inv: %d", fwd_kin->numJoints(), inv_kin->numJoints() );`

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -58,40 +58,85 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
 {
   // Check name
   if (fwd_kin->getName() != inv_kin->getName())
+  {
+    CONSOLE_BRIDGE_logError("checkKinematics: Manipulator names do not match!");
```
mpowelson, line 62: Same here

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -58,40 +58,85 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
...
   // Check link names
-  if (fwd_kin->getLinkNames().size() != inv_kin->getLinkNames().size() ||
-      !std::equal(fwd_kin->getLinkNames().begin(), fwd_kin->getLinkNames().end(), inv_kin->getLinkNames().begin()))
+  std::vector<std::string> l1 = fwd_kin->getLinkNames();
+  std::vector<std::string> l2 = inv_kin->getLinkNames();
+  std::sort(l1.begin(), l1.end());
+  std::sort(l2.begin(), l2.end());
+  if (l1.size() != l1.size())
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator link names size do not match!");
```
mpowelson, line 80: Should this be logError? It returns false.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -100,19 +145,32 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
   seed_angles.setZero();
   joint_angles2.setZero();
 
-  for (int t = 0; t < fwd_kin->numJoints(); ++t)
+  int nj = static_cast<int>(fwd_kin->numJoints());
```
mpowelson, line 148: const int?

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -100,19 +145,32 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
...
-      if (!test1.isApprox(test2, tol))
+      fwd_kin->calcFwdKin(test2, sols.middleRows(nj * i, nj));
+
+      if ((test1.translation() - test2.translation()).norm() > tol)
+      {
+        CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator translation norm is greater than tolerance %f!", tol);
         return false;
+      }
+
+      if (Eigen::Quaterniond(test1.linear()).angularDistance(Eigen::Quaterniond(test2.linear())) > tol)
```
mpowelson, line 167: angularDistance looks like quite a useful function. Should we be using that in trajopt?

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -100,19 +145,32 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
...
+      {
+        CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator orientation angular distance is greater than tolerance "
+                               "%f!",
+                               tol);
```
mpowelson, line 171: Does the same tolerance work for all of these or should they be broken out?

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
```diff
@@ -142,55 +200,111 @@ inline bool checkKinematics(const tesseract_kinematics::ForwardKinematics::Const
...
+  if (l1.size() != l1.size())
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator link names size do not match!");
     return false;
+  }
+
+  if (!std::equal(l1.begin(), l1.end(), l2.begin()))
+  {
+    CONSOLE_BRIDGE_logWarn("checkKinematics: Manipulator link names do not match!");
+    return false;
```
mpowelson, line 229: Perhaps this would be a useful utility?

```
bool isIdentical(const std::vector<std::string>& vec_1, const std::vector<std::string>& vec_2, std::string name = "")
```

mpowelson (approved):
Nothing major. Some optional suggestions

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: There could be a case where if you used a chain compared to a joint list the link list could be different. Even though they are different they may still produce the same adjacency map used for getting the active links. It should not return false, I will fix that.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: I need to check, but I think this should match the same as the angle of the angle axis, but I have not tested it. I know with angle axis we have to do some checking to get the acute angle which this may not have that issues. If not it, would be a better approach.

tesseract/tesseract_visualization/src/trajectory_interpolator.cpp
Levi-Armstrong: Yea, so if the current_time is less than the last_time then the time has started over. When it is set to the current_time the calculation of dt will be zero . The dt should be the current_time, because this only occurs on the first instruction when time restarts.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: I know clang complains about this in the function arguments but not sure about in the body. I will change it and see.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: I believe Eigen Approx only excepts one input since it is an equal comparison.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: I like it. I am adding it to tesseract_common utils.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: Updated

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: updated

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: It looks like clang only cares when it is one of the function arguments.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
Levi-Armstrong: fixed.

tesseract/tesseract_kinematics/include/tesseract_kinematics/core/validate.h
mpowelson: I meant you use the same tolerance to do multiple comparisons. Is the same tolerance ok for all of them or would you expect some comparisons to need different tolerances?

PR #389, opened 2020-09-23
Issue/fix graph taskflow

Levi-Armstrong:
This fixes a few issues found and changes the freespace taskflow to always try trajopt even if ompl fails.

image

### DIFF
--- a/tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/utils/utils.h
+++ b/tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/utils/utils.h
@@ -38,10 +38,20 @@ namespace tesseract_planning
  * @brief Gets joint position from waypoints that contain that information.
  *
  * Throws if waypoint does not directly contain that information
- *
+ * @param waypoint The waypoint to try and extract the joint position from
  * @return The joint values
  */
 const Eigen::VectorXd& getJointPosition(const Waypoint& waypoint);
+
+/**
+ * @brief Gets joint names from waypoints that contain that information.
+ *
+ * Throws if waypoint does not directly contain that information
+ *
+ * @param waypoint The waypoint to try and extract the joint position from
+ * @return The joint names
+ */
+const std::vector<std::string>& getJointNames(const Waypoint& waypoint);
 
 /**
  * @brief Get the joint positions ordered by the provided joint names
--- a/tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
+++ b/tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
@@ -182,8 +182,8 @@ tf::Taskflow& GraphTaskflow::generateTaskflow(ProcessInput input,
       {
         throw std::runtime_error("Invalide Edges for process index: " + std::to_string(src_idx));
       }
-      ++src_idx;
     }
+    ++src_idx;
   }
 
   return *taskflow;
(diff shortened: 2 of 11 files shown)


tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/utils/utils.h
```diff
@@ -43,6 +43,15 @@ namespace tesseract_planning
  */
 const Eigen::VectorXd& getJointPosition(const Waypoint& waypoint);
 
+/**
+ * @brief Gets joint names from waypoints that contain that information.
+ *
+ * Throws if waypoint does not directly contain that information
+ *
```
mpowelson, line 50: Missing @param waypoint

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
```diff
@@ -182,8 +182,8 @@ tf::Taskflow& GraphTaskflow::generateTaskflow(ProcessInput input,
       {
         throw std::runtime_error("Invalide Edges for process index: " + std::to_string(src_idx));
       }
-      ++src_idx;
     }
+    ++src_idx;
```
mpowelson, line 186: Ah. I noticed that TASK nodes didn't seem to work before. I just made it a conditional with both conditions going to the same node as a workaround. Good find.

tesseract/tesseract_visualization/include/tesseract_visualization/trajectory_interpolator.h
```diff
@@ -52,7 +52,7 @@ class TrajectoryInterpolator
 
   tesseract_planning::MoveInstruction getMoveInstruction(double request_duration) const;
 
-  double getMoveInstructionDuration(std::size_t index) const;
```
mpowelson, line 55: Not upset either way, but duration seems like a size. It seems like you are just trading a cast in trajectory_player for a cast in trajectory interpolator. I wonder if index should just be a size_t. It wouldn't make since to play an index larger than the size it can hold anyway.

mpowelson (commented):
Looks good.

tesseract/tesseract_planning/tesseract_process_managers/src/taskflow_generators/graph_taskflow.cpp
Levi-Armstrong: Yea it took a while to figure out what was going on. I was getting weird graphs.

tesseract/tesseract_visualization/include/tesseract_visualization/trajectory_interpolator.h
Levi-Armstrong: These are a good read. As a result I plan to avoid using std::size_t in class interface. Let me know what your take is on the information below.

[Google Style Guide](https://google.github.io/styleguide/cppguide.html#Integer_Types)
[Issue Discussion](https://github.com/fish-shell/fish-shell/issues/3493)

mpowelson: My take away is that there is a lot of debate. Switching to an int is probably best in this case. I don't know that it is going to stop me from using size_t in loops to avoid clang warnings though.

Levi-Armstrong: I agree, I think that it is one case I have seen it mentioned as acceptable. My thought is just not use it in interfaces/function parameters, but if it makes sense to use it internal to functions that is acceptable as you pointed out.

PR #390, opened 2020-09-25
Add taskflow debug observer

Levi-Armstrong:
This allow you enable debug with in a ProcessManager which will attache and observer which will output additional information as ti executes the taskfflow.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
+++ b/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
@@ -1,0 +1,61 @@
+/**
+ * @file debug_observer.h
+ * @brief Debug Taskflow Observer
+ *
+ * @author Matthew Powelson
+ * @date July 15. 2020
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2020, Southwest Research Institute
+ *
+ * @par License
+ * Software License Agreement (Apache License)
+ * @par
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_PROCESS_MANAGERS_DEBUG_OBSERVER_H
+#define TESSERACT_PROCESS_MANAGERS_DEBUG_OBSERVER_H
+
+#include <tesseract_common/macros.h>
+TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
+#include <taskflow/taskflow.hpp>
+#include <memory>
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
+
+namespace tesseract_planning
+{
+/**
+ * @brief This is currently simple debug observer which prints a message when each of the functions below are called.
+ */
+class DebugObserver : public tf::ObserverInterface
+{
+public:
+  using Ptr = std::shared_ptr<DebugObserver>;
+  using ConstPtr = std::shared_ptr<const DebugObserver>;
+
+  /**
+   * @brief Constructor
+   * @param name The name given to the Observer which also gets printed out.
+   */
+  DebugObserver(const std::string& name);
+
+  void set_up(size_t num_workers) final;
+
+  void on_entry(size_t w, tf::TaskView tv) final;
+
+  void on_exit(size_t w, tf::TaskView tv) final;
+};
+
+}  // namespace tesseract_planning
+
+#endif  // TESSERACT_PROCESS_MANAGERS_DEBUG_OBSERVER_H
--- a/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_manager.h
+++ b/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_manager.h
@@ -73,6 +73,25 @@ public:
    * @return True if successful
    */
   virtual bool clear() = 0;
+
+  /**
+   * @brief Enable/Disable debug for the process manager
+   *
+   * This will add an observer which will print out information on the task as it is executed.
+   *
+   * @param enabled
+   */
+  virtual void enableDebug(bool enabled) = 0;
+
+  /**
+   * @brief Enable/Disable profiling for the process manager
+   *
+   * This will add an observer which will profile task as it is executed. It will save a file in the temp directory
+   * which can be loaded into the taskflow profile webpage.
+   *
+   * @param enabled
+   */
+  virtual void enableProfile(bool enabled) = 0;
 };
 
 }  // namespace tesseract_planning
--- a/tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/raster_dt_process_manager.cpp
+++ b/tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/raster_dt_process_manager.cpp
@@ -27,14 +27,17 @@
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <functional>
 #include <taskflow/taskflow.hpp>
-#include <fstream>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 #include <tesseract_process_managers/process_managers/raster_dt_process_manager.h>
+#include <tesseract_process_managers/debug_observer.h>
+
 #include <tesseract_command_language/instruction_type.h>
 #include <tesseract_command_language/composite_instruction.h>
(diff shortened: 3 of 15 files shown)


tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
```diff
@@ -0,0 +1,54 @@
+/**
+ * @file debug_observer.h
+ * @brief Debug Taskflow Observer
+ *
+ * @author Matthew Powelson
+ * @date July 15. 2020
```
mpowelson, line 6: Needs updating

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <tesseract_common/macros.h>
+TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
+#include <taskflow/taskflow.hpp>
+#include <memory>
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
+
+namespace tesseract_planning
+{
+class DebugObserver : public tf::ObserverInterface
```
mpowelson, line 37: Doxygen

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/continuous_contact_check_process_generator.h
```diff
@@ -68,7 +68,7 @@ class ContinuousContactCheckProcessGenerator : public ProcessGenerator
 
   std::string name_;
 
-  double longest_valid_segment_length_{ 0.1 };
+  double longest_valid_segment_length_{ 0.005 };
```
mpowelson, line 71: This seems pretty small.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_manager.h
```diff
@@ -73,6 +73,25 @@ class ProcessManager
...
+
+  /**
+   * @brief Enable/Disable profiling for the process manager
+   *
+   * This will add an observer which will profile task as it is executed. It will save a file in the temp directory
+   * which can be loaded into the taskflow profile webpage.
+   *
+   * @param enabled
+   */
+  virtual void setProfile(bool enabled) = 0;
```
mpowelson, line 94: Maybe enableDebug and enableProfiling? SetProfile sounds like the motion planning profiles

mpowelson (approved):
I got excited for a minute. Can't wait to see the actual debug and profiling observers.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
Levi-Armstrong: I copied this from the you created so kept the same date.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
Levi-Armstrong: Done

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_manager.h
Levi-Armstrong: I like it.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/continuous_contact_check_process_generator.h
Levi-Armstrong: Yea, I will try 0.05 instead. The challenge I keep running into with continuous collision checking is that if you have two kinematic links that are always within say 2cm and are not disabled in ACM. Then if you move more than 2cm in a single time step it will report collision because the end state of the first shape overlaps with the start state of the second shape. This is why I have switched to using Discrete Continuous collision checking for motion planning and I am getting much better performance then I run the post check a fine resolution.

Levi-Armstrong: Also I did not mean to include this in this PR. I will remove it and create a different PR to discuss further.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/debug_observer.h
mpowelson: Oh ok. Got it.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/continuous_contact_check_process_generator.h
mpowelson: I was actually running into the same thing. Do you think DISCRETE_CONTINUOUS or SINGLE_TIMESTEP should be the default? It is also more difficult to debug in this framework since the trajopt plotting has to sort of be hacked in there.

mpowelson: Is `composite_profile->collision_constraint_config.type = trajopt::CollisionEvaluatorType::DISCRETE_CONTINUOUS;` all you have to do to set that? I'm still getting blue arrows

Levi-Armstrong:
> I got excited for a minute. Can't wait to see the actual debug and profiling observers.

I just added the profile observer take a look and let me know if anything should be changed. I switched to adding a string timestamp to the file names.

Levi-Armstrong:
It will save a file in the tmp directory and you can copy the contents into this [tool](https://taskflow.github.io/tfprof/).

mpowelson:
Did you push it? I don't see it

Levi-Armstrong:
Sorry I forced pushed. The observer is already available in Taskflow I just needed to create it and add it to the executor.

tesseract/tesseract_planning/tesseract_process_managers/src/process_managers/raster_dt_process_manager.cpp
```diff
@@ -201,6 +207,14 @@ bool RasterDTProcessManager::execute()
 {
   success_ = true;
 
+  DebugObserver::Ptr debug_observer;
+  std::shared_ptr<tf::TFProfObserver> profile_observer;
+  if (debug_)
+    debug_observer = executor_.make_observer<DebugObserver>("RasterDTProcessManagerObserver");
+
+  if (profile_)
+    profile_observer = executor_.make_observer<tf::TFProfObserver>();
```
Levi-Armstrong, line 216: Updated section along with saving it to a file below.

Levi-Armstrong: I am also updating in the Planning Request on the ROS side

mpowelson: Oh ok. I didn't realize it was in Taskflow. I'll have to try it sometime

Levi-Armstrong: Yea neither did I until recently. I ask a question looking for a way to trace the path taken through the graph and they pointed me to this.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/continuous_contact_check_process_generator.h
Levi-Armstrong: > I was actually running into the same thing. Do you think DISCRETE_CONTINUOUS or SINGLE_TIMESTEP should be the default? It is also more difficult to debug in this framework since the trajopt plotting has to sort of be hacked in there.

I do believe we should set the default to Discrete Continuous. I think because of the previous issue it makes motion planning difficult to debug. I do think the post check is still good to use continuous but you need to do this at a fine resolution to no run into issues.

Levi-Armstrong: > Is `composite_profile->collision_constraint_config.type = trajopt::CollisionEvaluatorType::DISCRETE_CONTINUOUS;` all you have to do to set that? I'm still getting blue arrows

You should still get the blue arrows with Discrete continuous. When interpolating between two states it stores the contact results with the deepest penetration and then shows the contact point on the start and end state like what is done with Continuous Collision checking.

Levi-Armstrong: I reset it for this PR and will open up a new one changing these settings.

Levi-Armstrong: I have create PR #392 to further discuss with proposed changes.

PR #392, opened 2020-09-27
Update default longest valid segment length and set default to use discrete continuous.

Levi-Armstrong:
The challenge I keep running into with continuous collision checking is that if you have two kinematic links that are always within say 2cm and are not disabled in ACM. Then if you move more than 2cm in a single time step it will report collision because the end state of the first shape overlaps with the start state of the second shape. This switches the default to leverage Discrete Continuous which will make motion planning more stable and will use casted collision checking in the post check with a fine resolution to catch collisions if the planners longest valid segment length was not set fine enough.


mpowelson (approved):

mpowelson:
In the future, maybe we could add a 2 step process where we do a continuous check and then only run the smaller LVS discrete check on the timesteps (and maybe even links) that fail. Sort of like a BVH but for time. Probably more important things to work on at the moment, but you could imagine a system where we take the continuous collision time and segment the discretization based on that and keep doing that until it is no longer in collision or the step size is sufficiently small that we decide it actually is in collision.

PR #393, opened 2020-09-28
Update trajectory player to support rviz visualization

Levi-Armstrong:


mpowelson (approved):

mpowelson:
@Levi-Armstrong I just tried out the archive viewer again. I have not looked into why, but trail display mode doesn't seem to work. Single and loop work fine.

PR #394, opened 2020-09-28
Improve global raster taskflow

Levi-Armstrong:
Through testing it was found for the global planner that since Descartes was already ran as the global planner that running TrajOpt first succeeds more often that it does not. This switches it to try TrajOpt first then if it fails it will then try OMPL followed by TrajOpt.

image


Levi-Armstrong:
@colin-lewis-19 Do mind also taking a look?

colin-lewis-19:
Builds and runs more predictably than starting with OMPL. This looks good to me.

PR #396, opened 2020-09-29
Remove const from return type in ikfast.h

Levi-Armstrong:
The [Cpp Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con2-by-default-make-member-functions-const) recommends against forcing `const` on function return types and function parameters that are not references or pointers. The user can decide if they want to store the function return variable into a const or not. If you use Clang-Tidy, warnings will be produced, but since this does not have a test that leverages this header it was not found until recently when using it in another project.

Given:

`const int test(const int i)`

Should be:

`int test(int i)`

### DIFF
--- a/tesseract/tesseract_kinematics/include/tesseract_kinematics/ikfast/external/ikfast.h
+++ b/tesseract/tesseract_kinematics/include/tesseract_kinematics/ikfast/external/ikfast.h
@@ -197,7 +197,7 @@ public:
   }
 
   virtual const std::vector<int>& GetFree() const { return _vfree; }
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
@Levi-Armstrong: could you add some rationale for *why* you've made this change?

The diff shows what the change is, but it doesn't record why the change was made. In 6 months from now I'm more interested in the *why*, but it isn't recorded anywhere.

Levi-Armstrong:
@gavanderhoorn Sure, I have updated the main description.

mpowelson (approved):

PR #397, opened 2020-09-29
Add support for user defined find tcp callbacks

Levi-Armstrong:
This was added to support finding tcp when using ROS where it is only available via TF.


Levi-Armstrong:
@colin-lewis-19 Ready to review.

colin-lewis-19:
Understood. Pulling down for testing

colin-lewis-19:
Seems to work as intended. Prevents planning if the TCP TF publisher is killed (if a camera driver were to crash, for instance), but those type of circumstances should probably be handled upstream anyways

Levi-Armstrong:
@colin-lewis-19 It has been update to where findTCP will throw an exception if not found and updated motion planners to handle this exception to return failed plan with status ErrorInvalidInput.

PR #398, opened 2020-10-01
Fix/simple lvs tests

marrts:
- Updated tests of LVS interpolation to check that the proper number of steps are created

- Changed calls of fixed size interpolation to LVS interpolation

@mpowelson I seem to be getting the proper number of steps for every scenario


mpowelson:
Thanks for looking into this. I see you reassigned it to Levi which is fine. It would likely be next week before I got to this. If it is still open then I can take a look.

Levi-Armstrong:
See PR #438

PR #399, opened 2020-10-01
Remove inheritance of std::vector from Composite Instruction

Levi-Armstrong:
Having the CompositeInstruction inherit directly form std:vector was going to make python wrapper very difficult or even impossible leveraging SWIG. Same issues exists for the JointWaypoint and CartesianWaypoint. This remove the inheritance and adds all of the container methods required.

### DIFF
--- a/tesseract/tesseract_collision/CMakeLists.txt
+++ b/tesseract/tesseract_collision/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.0)
+cmake_minimum_required(VERSION 3.6.0)
 project(tesseract_collision VERSION 0.1.0 LANGUAGES CXX)
 
 find_package(Eigen3 REQUIRED)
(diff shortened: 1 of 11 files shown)


Levi-Armstrong:
@johnwason Do you mind taking a look and let me know if you see any issues?

Levi-Armstrong:
@johnwason I have change the Cartesian and Joint Waypoint where they do not inherit from Eigen types but still operate the same. Let me know if any additional changes need to be made to the these classes.

Levi-Armstrong:
@mpowelson He made some possible breaking changes depending on how much you are using JointWaypoint and CartesianWaypoint as Eigen Types. I have added several methods so it plays nice with Eigen but you may need to further expose methods if you run into any issues.

mpowelson:
Thanks for letting me know

Levi-Armstrong:
Also added initializer_list support for the JointWaypoint.

Levi-Armstrong:
@johnwason  We have been using this for a while without any issues so merging but if further changes are required to better support python please let me know and I will take of it.

PR #400, opened 2020-10-01
Switch to c++17 and leverage shared_mutex in tesseract_environment

Levi-Armstrong:


tesseract/tesseract_environment/src/core/environment.cpp
```diff
@@ -443,23 +443,24 @@ void Environment::removeAllowedCollision(const std::string& link_name)
...
   tesseract_scene_graph::AllowedCollisionMatrix::ConstPtr acm = scene_graph_->getAllowedCollisionMatrix();
 
   return acm;
 }
 
 tesseract_scene_graph::Joint::ConstPtr Environment::getJoint(const std::string& name) const
 {
-  std::lock_guard<std::mutex> lock(mutex_);
+  std::shared_lock<std::shared_mutex> lock(mutex_);
+  ;
```
mpowelson, line 455: extra `;`

mpowelson (approved):
One inconsequential change

mpowelson:
> One inconsequential change

Actually, just looked at CI. That's not good.

mpowelson:
Yeah we may be more careful than I was thinking because have cases all over the place where unique locked things are calling unique locked functions.

Levi-Armstrong:
Yep, I need to look into this further to see where the double lock is occurring.

Levi-Armstrong:
This is included in #544

PR #401, opened 2020-10-03
Fix done and error callback in simple process manager

Levi-Armstrong:
The callback was using a reference to a local variable causing a runtime error.


Levi-Armstrong:
@colin-lewis-19 Can you please review?

PR #402, opened 2020-10-03
Update due to change with descartes compound edge evaluator

Levi-Armstrong:


Levi-Armstrong:
@colin-lewis-19 Ready for review.

PR #407, opened 2020-10-12
Add individual CI badges and Windows CI build

Levi-Armstrong:
I ran into a lot of issues using the visibility control related to weird build issues in boost. I have switched it such that all symbols are export and now everything builds. I did run into some issues related to FCL and Bullet because these libraries are already available in the ros melodic and noetic windows package. The cmake find_package was finding the correct config file but to get it to find the header and libraries I had to add the include directories and library directory as shown below and the critical thing was the `BEFORE` to make it actual use the version in my workspace instead of the one provided.

```
include_directories(BEFORE ${FCL_INCLUDE_DIRS})
link_directories(BEFORE ${FCL_LIBRARY_DIRS})

include_directories(BEFORE "${BULLET_ROOT_DIR}/${BULLET_INCLUDE_DIRS}")
link_directories(BEFORE "${BULLET_ROOT_DIR}/${BULLET_LIBRARY_DIRS}")
```

Also currently, not all unit tests run on windows but they do build.


Levi-Armstrong:
@johnwason I believe I have found a fix for the boost windows build error explained [here](https://github.com/boostorg/graph/issues/229).  Not sure if it is an issue in boost or if I should have always been using the hash map.

johnwason:
Have you tried std::set<>?

Levi-Armstrong:
I have not yet but the interesting thing is I copied the exact code into an isolated repository [here](https://github.com/Levi-Armstrong/boost_graph_issue) and it builds fine. I am so confused as to why it will build there but not here.

gavanderhoorn:
I'd almost give you a badge just for persevering .. :medal_sports:

Issue #391, opened 2020-09-25
Change profile to use profile named DEFAULT if not found


mpowelson:
How confident are we that setting the SafetyMarginData in our trajopt planner is getting correctly propagated? Even doing this like below I am getting collision distances reported unless I actually change the default values in trajop_collision_config.h. Perhaps I am doing something obviously wrong. Below does not work, but changing the default values (which I should be changing below) does. However, that obviously doesn't let you set just a single value like I was wanting to do.

```
    auto trajopt_planner = std::make_shared<TrajOptMotionPlanner>();
    trajopt_planner->problem_generator = &DefaultTrajoptProblemGenerator;
    trajopt_planner->plan_profiles["FREESPACE"] = std::make_shared<TrajOptDefaultPlanProfile>();
    auto composite_profile = std::make_shared<TrajOptDefaultCompositeProfile>();

    auto cost_safety_margin = std::make_shared<trajopt::SafetyMarginData>(0.00, 20);
    cost_safety_margin->setPairSafetyMarginData("link_1", "link_2", 0.0, 0);
    composite_profile->special_collision_cost = cost_safety_margin;

    auto cnt_safety_margin = std::make_shared<trajopt::SafetyMarginData>(0.0, 20); 
    cnt_safety_margin->setPairSafetyMarginData("link_1", "link_2", 0.0, 10);

    composite_profile->special_collision_constraint = cnt_safety_margin;
    composite_profile->collision_constraint_config.safety_margin_buffer = 0.0;
    composite_profile->collision_cost_config.safety_margin_buffer = 0.0;
    composite_profile->collision_constraint_config.safety_margin = 0.0;
    composite_profile->collision_cost_config.buffer_margin = 0.0;

    composite_profile->collision_cost_config.type = trajopt::CollisionEvaluatorType::DISCRETE_CONTINUOUS; 
    composite_profile->collision_constraint_config.type = trajopt::CollisionEvaluatorType::DISCRETE_CONTINUOUS;

    trajopt_planner->composite_profiles["FREESPACE"] = composite_profile;
    auto trajopt_generator = std::make_unique<MotionPlannerProcessGenerator>(trajopt_planner);
    trajopt_idx = graph->addNode(std::move(trajopt_generator), GraphTaskflow::NodeType::CONDITIONAL);
```

mpowelson:
@Levi-Armstrong Have you used the special_collision_cost?

Levi-Armstrong:
[...]

PR #412, opened 2020-10-16
Add utility for getting profiles

mpowelson:
Consolidates all of the profile remapping into a single utility that gets used everywhere. Addresses #391. Note that all planners should now set a "DEFAULT" profile

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
@@ -132,30 +132,29 @@ inline std::string getProfileString(const std::string& profile,
   return results;
 }
 
+/**
+ * @brief Gets the profile specified from the profile map
+ * @param profile The requested profile
+ * @param profile_map map that contains the profiles
+ * @param default_profile Profile that is returned if the requested profile is not found in the map. Default = nullptr
+ * @return The profile requested if found. Otherwise the default_profile
+ */
 template <typename ProfileType>
 std::shared_ptr<ProfileType>
-getProfile(const std::string& profile, const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map, std::string default_profile = "DEFAULT")
+getProfile(const std::string& profile,
+           const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map,
+           std::shared_ptr<ProfileType> default_profile = nullptr)
 {
-  std::string profile_string = profile;
   std::shared_ptr<ProfileType> results;
-  auto it = profile_map.find(profile_string);
+  auto it = profile_map.find(profile);
 
-  // If profile not found, look for default_profile.
   if (it == profile_map.end())
   {
-    profile_string = default_profile;
-    CONSOLE_BRIDGE_logDebug("Profile %s was not found. Setting to %s. Available profiles:", default_profile.c_str(), profile.c_str());
+    CONSOLE_BRIDGE_logDebug("Profile %s was not found. Using default if available. Available profiles:",
+                            profile.c_str());
     for (const auto& pair : profile_map)
       CONSOLE_BRIDGE_logDebug("%s", pair.first.c_str());
-    it = profile_map.find(profile_string);
-  }
-
-  // If default_profile not found, print an error
-  if (it == profile_map.end())
-  {
-    CONSOLE_BRIDGE_logError("Profile %s was not found. Available profiles:", profile.c_str());
-    for (const auto& pair : profile_map)
-      CONSOLE_BRIDGE_logError("%s", pair.first.c_str());
+    results = default_profile;
   }
   else
     results = it->second;
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
@@ -134,10 +134,23 @@ inline std::string getProfileString(const std::string& profile,
 
 template <typename ProfileType>
 std::shared_ptr<ProfileType>
-getProfile(const std::string& profile, const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map)
+getProfile(const std::string& profile, const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map, std::string default_profile = "DEFAULT")
 {
+  std::string profile_string = profile;
   std::shared_ptr<ProfileType> results;
-  auto it = profile_map.find(profile);
+  auto it = profile_map.find(profile_string);
+
+  // If profile not found, look for default_profile.
+  if (it == profile_map.end())
+  {
+    profile_string = default_profile;
+    CONSOLE_BRIDGE_logDebug("Profile %s was not found. Setting to %s. Available profiles:", default_profile.c_str(), profile.c_str());
+    for (const auto& pair : profile_map)
+      CONSOLE_BRIDGE_logDebug("%s", pair.first.c_str());
+    it = profile_map.find(profile_string);
+  }
+
+  // If default_profile not found, print an error
   if (it == profile_map.end())
   {
     CONSOLE_BRIDGE_logError("Profile %s was not found. Available profiles:", profile.c_str());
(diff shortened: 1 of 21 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
```diff
@@ -101,6 +104,52 @@ bool isValidState(const tesseract_kinematics::ForwardKinematics::ConstPtr& robot
...
+    if (p != remap->second.end())
+      results = p->second;
+  }
+  return results;
+}
+
+template <typename ProfileType>
+std::shared_ptr<ProfileType> getProfile(const std::string& profile,
+                                        const std::unordered_map<std::string, std::shared_ptr<ProfileType>>&
+                                            profile_map /*, std::shared_ptr<ProfileType> default_profile = nullptr*/)
```
Levi-Armstrong, line 138: Commented out code?

Levi-Armstrong (changes requested):
Looks good, just one minor comment.

Levi-Armstrong:
On a side note, it may be good to create a higher level class that contains all profile that gets passed to everything, for example we can then change say trajopt, ompl, descartes have collision information as a profile key this way we do not have to make sure we configure all of the planners the same. Thoughts? Also not saying that needs to be done as part of this PR.

mpowelson:
@Levi-Armstrong It looks like the Focal build is missing some profiles in the unit test for some reason?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
mpowelson: removed

mpowelson:
Fixed clang-tidy. It looks like the raster process unit tests are failing now because they can't find their profile. Not sure why it is only on Focal.

Levi-Armstrong:
Have you rebased on the latest?

mpowelson:
I just did and it is still missing.

tesseract/tesseract_planning/tesseract_process_managers/src/process_generators/iterative_spline_parameterization_process_generator.cpp
```diff
@@ -117,23 +101,18 @@ int IterativeSplineParameterizationProcessGenerator::conditionalProcess(ProcessI
   }
 
   Eigen::VectorXd velocity_scaling_factors = Eigen::VectorXd::Ones(static_cast<Eigen::Index>(flattened.size())) *
-                                             cur_plan_profile->max_velocity_scaling_factor;
+                                             cur_composite_profile->max_velocity_scaling_factor;
```
Levi-Armstrong, line 104: Was thing moved from plan profile to composite profile?

mpowelson: We were already pulling it from composite_profiles. The local variable we use here was just misnamed.

Levi-Armstrong: Ah.

Levi-Armstrong:
Found out why the other pass and Focal does not. All others have to use the AFTER_SCRIPT which never got updated to include tesseract_process_managers. I switch out the tesseract_process_planners with tesseract_process_managers in the CI script it should fail on those to.

Levi-Armstrong:
Also I expect all of the test may fail now since a default profile is not being created if one is not found, instead now the getProfile returns a nullptr.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
```diff
@@ -101,6 +104,51 @@ bool isValidState(const tesseract_kinematics::ForwardKinematics::ConstPtr& robot
...
+      results = p->second;
+  }
+  return results;
+}
+
+template <typename ProfileType>
+std::shared_ptr<ProfileType>
+getProfile(const std::string& profile, const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map)
+{
+  std::shared_ptr<ProfileType> results;
```
Levi-Armstrong, line 139: recommend changing this to `auto results = std::make_shared<ProfileType>();` this way it will return a profile with default setting if one is not found.

mpowelson: Ah. I forgot to set it to DEFAULT if it wasn't found. That should fix it.

mpowelson: Also, I can't do make_shared because ProfileType is an abstract class.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
```diff
@@ -101,6 +104,64 @@ bool isValidState(const tesseract_kinematics::ForwardKinematics::ConstPtr& robot
...
+    auto p = remap->second.find(profile);
+    if (p != remap->second.end())
+      results = p->second;
+  }
+  return results;
+}
+
+template <typename ProfileType>
+std::shared_ptr<ProfileType>
+getProfile(const std::string& profile, const std::unordered_map<std::string, std::shared_ptr<ProfileType>>& profile_map, std::string default_profile = "DEFAULT")
```
Levi-Armstrong, line 137: Maybe switch this to where the user passes in the actual default profile instead of the string?

mpowelson: Ok. I made this change

mpowelson:
@Levi-Armstrong When I try to run clang tidy locally, I always get a bunch of erroneus errors of things that are unused when they are actually used etc. Any idea what that is about?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
Levi-Armstrong: I like the change, but maybe we should put the "DEFAULT" in a static const variable that gets used everywhere?

mpowelson: You mean for each planner store the default profile in a separate const variable? We would have to pass them into the problem generators for ompl, trajopt, and descartes. Ideally, the default in getProfile would never be used and it would just use whatever has been set as DEFAULT. They are only used in the the raster managers because they currently replace the profile_map of the planners rather than inserting the new entries.

Levi-Armstrong: Sorry, I meant add the line below some where in a base class that all use then use this variable instead of `"DEFAULT"`.

```
static const std::string DEFAULT_PROFILE_KEY = "DEFAULT";
```

Levi-Armstrong:
> @Levi-Armstrong When I try to run clang tidy locally, I always get a bunch of erroneus errors of things that are unused when they are actually used etc. Any idea what that is about?

I have seen this before but do not remember what the issue was.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/planner_utils.h
mpowelson: I added it to MotionPlanner.

Levi-Armstrong: Maybe we should move somewhere into the command lanaguage which is the lowest level and then it could be used a few places in the command language along with in motion planning?

mpowelson: Where would you want it? The Instruction?

Levi-Armstrong: Maybe just create a types.h or constants.h in the same directory as composite_instruction.h?

mpowelson:
I added the default key like we discussed. I kept these commits separate for possibly easier review, but we can squash them when we merge it.

Levi-Armstrong (approved):

PR #413, opened 2020-10-17
Fix use-after-move in Tesseract::init.

a-price:


Levi-Armstrong (approved):

Levi-Armstrong:
Thank you for the fix. Surprised clang-tidy did not catch this. Once CI passes I will merge.

Levi-Armstrong:
Note to myself: This is only required in master, the feature/CommandLanguage branch switched to cloning the scene graph so no fix required in that branch.

PR #415, opened 2020-10-19
Add ability to change joint limits 

mpowelson:


mpowelson:
@Levi-Armstrong Could you take a look at this? Is this all I have to do or will I need to trigger the manipulator manager to update somehow?

Levi-Armstrong:
For the environment that is it, but we need to figure out how to update the manipulator manager. I starting to thing we should have the environment own the manipulator manager then it could handle updating things like this. What do you think?

mpowelson:
The ManipulatorManager has an Environment::ConstPtr as a class member. I'm really not sure. It doesn't not make sense. The environment already owns the collision checkers. I will say though, if we do that, there isn't really that much in the Tesseract anymore.

I wonder if we should instead move the CommandHistory aspect to the Tesseract. We could make the commands TesseractCommands then we could register a command history callback where each "manager" (environment, manipulator, maybe collision) registers a function to be called when the CommandHistory is updated.

mpowelson:
Also, I'm not entirely sure that this doesn't do it as is. The joint limits are a pointer in the scene graph. I assume the scene graph in the manipulator manager is the same one as in the environment. Do you think there is a copy somewhere I would need to update?

Levi-Armstrong:
> Also, I'm not entirely sure that this doesn't do it as is. The joint limits are a pointer in the scene graph. I assume the scene graph in the manipulator manager is the same one as in the environment. Do you think there is a copy somewhere I would need to update?

Not sure,  the kinematics objects make a copy so they would need to be updated.

Levi-Armstrong:
> The ManipulatorManager has an Environment::ConstPtr as a class member. I'm really not sure. It doesn't not make sense. The environment already owns the collision checkers. I will say though, if we do that, there isn't really that much in the Tesseract anymore.
> 
> I wonder if we should instead move the CommandHistory aspect to the Tesseract. We could make the commands TesseractCommands then we could register a command history callback where each "manager" (environment, manipulator, maybe collision) registers a function to be called when the CommandHistory is updated.

I think either way would work, though not sure which would be the better approach.

mpowelson:
I see now. They get parsed in to KDL structures.

For now, I could just add an update method to the FK and IK structures and the manipulator manager (which just calls update on all of the solvers). Internally, they could just call init again.

Levi-Armstrong:
> I see now. They get parsed in to KDL structures.
> 
> For now, I could just add an update method to the FK and IK structures and the manipulator manager (which just calls update on all of the solvers). Internally, they could just call init again.

sounds good to me.

mpowelson:
> > I see now. They get parsed in to KDL structures.
> > For now, I could just add an update method to the FK and IK structures and the manipulator manager (which just calls update on all of the solvers). Internally, they could just call init again.
> 
> sounds good to me.

@Levi-Armstrong I'll have to change fwd_kin_manipulators_default_ to store Ptrs instead of ConstPtr. Unless there is another way I should be doing this.

mpowelson:
Do fwd_kin_manipulators and fwd_kin_manipulators_default contain duplicates? If I just loop over fwd_kin_manipulators does that include fwd_kin_manipulators_default?

Levi-Armstrong:
> Do fwd_kin_manipulators and fwd_kin_manipulators_default contain duplicates? If I just loop over fwd_kin_manipulators does that include fwd_kin_manipulators_default?

It should.

mpowelson:
I added the update method. I'm not sure how useful it will be for the IK solves as they all store the limits as objects rather than pulling them from the SceneGraph::Ptr.

Levi-Armstrong:
> I added the update method. I'm not sure how useful it will be for the IK solves as they all store the limits as objects rather than pulling them from the SceneGraph::Ptr.

What about adding a `setLimits` to the forward and inverse kinematics interface. Then let the manager loop over all kinematics objects and set there limits. It loop over and get the list of active joints from the kinematics objects and extract the data form the scene graph to build a new kinematics limits object.

mpowelson:
With current changes, one breaking change is that ManipulatorManager is now in the namespace tesseract_environment instead of tesseract

Levi-Armstrong:
I like the direction this going.

mpowelson:
@Levi-Armstrong I think this is ready for a look. I wrote a unit test that I think shows that it is working. It is not checking the underlying KDL datastructures, but it should be changing them.

I am still recreating the KDL tree each time.  It should be nearly there for us to add functions to update each FK/IK solver individually as a performance improvement. However, I don't know if I will get to that right now.

mpowelson:
Still todo. 

- Update tesseract_rosutils
- Add back in Tesseract::init(Environment)

mpowelson:
@Levi-Armstrong Were you wanting me to merge this too?

Levi-Armstrong (approved):

PR #423, opened 2020-10-21
Separate public and private compiler option and add back -mno-avx

Levi-Armstrong:


marrts (approved):
This fixed my Eigen alignment issues

Issue #404, opened 2020-10-08
WaypointInCollision doesn't ever return collisions


mpowelson:
Somewhere we discussed that [WaypointInCollision](https://github.com/ros-industrial-consortium/tesseract/blob/cf27ce01e06b880e8bcb2e26f95ec04281f157e6/tesseract/tesseract_planning/tesseract_process_managers/src/process_generators/fix_state_collision_process_generator.cpp#L46) doesn't seem to be working. I've again run into this where inputs that pass through this function without problem don't satisfy trajopt contraints. We need to make sure that all of our collision checks are done consistently across, this, the post check processes, and the motion planners.

mpowelson:
This seems to work instead.

```

      std::vector<tesseract_collision::ContactResultMap> contacts;

      std::vector<std::string> joint_names;
      std::vector<double> joint_values(start_pos.data(), start_pos.data() + start_pos.size());
      auto state = input.tesseract->getEnvironment()->getState(joint_names, joint_values);

      const bool no_collisions = !tesseract_environment::checkTrajectoryState(contacts, *manager, state);

      {
        for (std::size_t i = 0; i < contacts.size(); i++)
          for (const auto& contact_vec : contacts[i])
            for (const auto& contact : contact_vec.second)
              CONSOLE_BRIDGE_logWarn("%s", ("Links: " + contact.link_names[0] + ", " +
                                       contact.link_names[1] + " Dist: " + std::to_string(contact.distance))
                                          .c_str());
      }
```

Levi-Armstrong:
The issue is here. It should not be `traj.rows() - 1` it should just be `traj.rows()`

https://github.com/ros-industrial-consortium/tesseract/blob/cf27ce01e06b880e8bcb2e26f95ec04281f157e6/tesseract/tesseract_environment/include/tesseract_environment/core/utils.h#L345

mpowelson:
Yep. That would do it.

mpowelson:
I wonder if we are exploiting that bug anywhere.

Levi-Armstrong:
Not sure. I looked through that file and did not see anything else.

Levi-Armstrong:
@mpowelson Do you have a PR in the works to fix the issue; If not I can create one?

mpowelson:
[...]

PR #426, opened 2020-10-23
Add MoveWaypointFromCollisionRandomSampler to FixStateCollisionProcess

mpowelson:
I was running into cases where the trajopt problem would fail to converge. This adds a fallback method that uses random sampling to find a non collision state like MoveIt does. It also addresses the two bullet points in #404 - adding unit tests and using the contact distance in the check function.

I included the boxbox urdf from trajopt to make testing easier.


mpowelson:
The unit test pass for me on melodic. I'm not sure why it failed on Noetic. I rebased and added an assert to catch incorrect kinematics size. We'll see if it passes now.

mpowelson:
This can be squashed

mpowelson:
@Levi-Armstrong Looks like that fixed it if you get a chance to look at this.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/fix_state_collision_process_generator.h
```diff
@@ -100,5 +117,47 @@ class FixStateCollisionProcessGenerator : public ProcessGenerator
...
   void process(ProcessInput input) const;
 };
+
+/**
+ * @brief Checks if a joint state is in collision
+ * @param start_pos Vector that represents a joint state
+ * @param input Process Input associated with waypoint. Needed for kinematics, etc.
+ * @return True if in collision
+ */
+bool StateInCollision(const Eigen::Ref<Eigen::VectorXd>& start_pos,
```
Levi-Armstrong, line 127: I think this can just be `Eigen::Ref<Eigen::VectorXd> start_pos`,

Levi-Armstrong (commented):
Looks good, I just one minor comment.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_generators/fix_state_collision_process_generator.h
mpowelson: Oh. This doesn't have to be writable. Based on the trajopt PR. I think it should be `const Eigen::Ref<const Eigen::VectorXd>& start_pos`

Levi-Armstrong: Ah, change looks good.

Levi-Armstrong (approved):

Levi-Armstrong:
You want to squash merge this when CI passes?

mpowelson:
Yeah that's fine

PR #428, opened 2020-10-26
Add ProfileSwitchProcessGenerator

mpowelson:
This generator simply returns a value specified in the composite profile. This can be used to switch execution based on the profile


Levi-Armstrong:
Just curious what the use case is for the switch profile generator? Is is doing the same thing as the remapping or is it used differently?

mpowelson:
This is used to switch the execution based on profile. So in my case I have a profile where I just want to quickly check if things are viable and another one where I want to actually motion plan. So when the first profile is active, this can be set to return 0 to fail quickly whereas when the other is active, it returns 1 to continue to planning.
Screenshot from 2020-10-27 09-08-01

mpowelson:
I guess the visibility control headers have been removed? Anyway, I rebased and fixed the headers. Hopefully it will pass now.

Levi-Armstrong:
Yea they were causing all sorts of issues with boost, so I removed it and just make all symbols public for windows build.

Levi-Armstrong (approved):

Issue #410, opened 2020-10-15
Instructions don't pass profiles to results


mpowelson:
I'm not sure where this should best be fixed, but the problem is that the results in the ProcessInput don't get populated with the same profile as the instructions (inputs). This means that any process that uses the results profile (currently only time parameterization I think, but potentially execution) always uses the defaults.

Levi-Armstrong:
I would believe it is the responsibility of the seed generator to fill this information out. I just checked the simple planner generators and they are not setting them.

mpowelson:
Yeah. I am using `generateSkeletonSeed`. It looks like it sets the Plan/Move Instruction profiles but not the composites.

mpowelson:
Nevermind. It looks like it is. Maybe I'm doing something wrong on my side.

mpowelson:
To turn this issue into concrete action items.
- [ ] Unit test verifying that generateSkeletonSeed and any other seed generators set the profile

mpowelson:
I added a unit test in #431. However, in doing that I realized the real problem. The profile is set correctly in generateSkeletonSeed. However that isn't passed on to the MoveInstructions in the simple planner does not pass the profile from the composite to the individual move instructions.

mpowelson:
In fact, now that I am looking at this, it isn't clear to me why we need the generateSkeletonSeed step. The simple planner should recursively handle composite instructions.

To address this issue though, I think I am going to change the step generators to take a PlanInstruction to get meta data from instead of just the manipulator info. Then we can set the profile and description as well.

mpowelson:
Actually, base_instructin is already what we want. I'm not sure why the manipulator info is being passed in anyway. There is some logic going to with manip_info.getCombined to combine manipulator infos. Regardless, I should just be able to add `setProfile(base_instruction.getProfile());`

PR #433, opened 2020-10-27
Fix SimplePlanner step generators to correctly set profile

mpowelson:
This should fully address #410


mpowelson:
Missed the composite instructions. Now they get the correct profile as well.

Levi-Armstrong (approved):

Levi-Armstrong:
It can be merged once CI passes.

PR #436, opened 2020-10-29
Add raster only process managers

Levi-Armstrong:
This PR breaks up the `RasterProcessManager `into two separate classes `RasterProcessManager` and `RasterGlobalProcessManager`. It also adds two additional raster process managers `(RasterOnlyProcessManager and RasterOnlyGlobalProcessManager)` which takes a program that does not include a from start and to home in the program. It also includes unit tests for the newly added managers.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
+++ b/tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
@@ -52,28 +52,37 @@ struct ProcessInput
   ProcessInput(tesseract::Tesseract::ConstPtr tesseract,
                const Instruction* instruction,
                const ManipulatorInfo& manip_info,
-               Instruction* seed);
+               Instruction* seed,
+               bool verbose = false);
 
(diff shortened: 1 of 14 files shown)


Levi-Armstrong:
This has been tested on system.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
```diff
@@ -52,29 +52,38 @@ struct ProcessInput
...
+               bool verbose = false);
 
   /** @brief Tesseract associated with current state of the system */
   const tesseract::Tesseract::ConstPtr tesseract;
 
   /** @brief Global Manipulator Information */
   const ManipulatorInfo& manip_info;
 
+  /** @brief Verbose Output */
+  bool verbose{ false };
```
mpowelson, line 85: I wonder if this should be an int so we can have different verbosity levels.

Levi-Armstrong: I was thinking the same thing after noticing that in some places we were using the console bridge log level to control output. Should we just use that as a parameter?

PR #438, opened 2020-11-02
Fix simple lvs tests

Levi-Armstrong:
This rebases PR #398 on master

I ran into issues getting this to fully function as the default because TrajOpt requires a minimum set states for joint velocity, acceleration, and jerk smoothing. This is also an issue with the interpolate and assign components but the defaults handled but you end up with a large number of states because this was manager a the plan instruction where the number of states were set. I looked at adding a composite profile for the simple planner but it quickly became difficult especially when you got to the raster planners because the simple planner is not aware of how the program would be broken up. Instead I created a process generator to handle this which increases the seed length only in the case where TrajOpt is used in the task flow. This check the input seed size and if it does not meet the minimum length it subdivides the trajectory until the min length is meet. @mpowelson 

image


colin-lewis-19:
This builds, runs, and fixes problematic motions for me.

Levi-Armstrong:
@mpowelson FYI, this PR changes the default behavior.

PR #442, opened 2020-11-12
Add vertex evaluator to descartes

Levi-Armstrong:
This PR creates a vertex validator and vertex allocator that take a descartes problem similar to the edge evaluator. Also adds missing delete constructors for the motion planners base class.

@mpowelson  I wanted to let you know, in case you are using Descartes planner, this is a breaking change if you are explicitly setting the is_valid for he Descartes profile.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
@@ -40,13 +40,15 @@ namespace tesseract_planning
 class MotionPlanner
 {
 public:
+  using Ptr = std::shared_ptr<MotionPlanner>;
+  using ConstPtr = std::shared_ptr<const MotionPlanner>;
   /** @brief Construct a basic planner */
   MotionPlanner(std::string name) : name_(std::move(name)) {}
   virtual ~MotionPlanner() = default;
-  MotionPlanner(const MotionPlanner&) = default;
-  MotionPlanner& operator=(const MotionPlanner&) = default;
-  MotionPlanner(MotionPlanner&&) = default;
-  MotionPlanner& operator=(MotionPlanner&&) = default;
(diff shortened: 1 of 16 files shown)


colin-lewis-19:
^ This works as intended on my project

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/core/planner.h
```diff
@@ -40,13 +40,15 @@ namespace tesseract_planning
...
   MotionPlanner(std::string name) : name_(std::move(name)) {}
   virtual ~MotionPlanner() = default;
-  MotionPlanner(const MotionPlanner&) = default;
-  MotionPlanner& operator=(const MotionPlanner&) = default;
-  MotionPlanner(MotionPlanner&&) = default;
-  MotionPlanner& operator=(MotionPlanner&&) = default;
+  MotionPlanner(const MotionPlanner&) = delete;
+  MotionPlanner& operator=(const MotionPlanner&) = delete;
+  MotionPlanner(MotionPlanner&&) = delete;
+  MotionPlanner& operator=(MotionPlanner&&) = delete;
```
mpowelson, line 51: Do we need a clone now? Are we using this anywhere?

Levi-Armstrong: Probably should add a clone, but I think right now we are creating shared and unique pointers which get passed around. It would be good to add a clone but nothing is currently using this library in that way.

Levi-Armstrong: I am adding the clone method now.

Levi-Armstrong: @mpowelson I added the clone method.

PR #443, opened 2020-11-12
Add TrajOpt Ifopt planner

mpowelson:
Adds a Tesseract planner for the trajopt_ifopt planner. This planner will eventually replace the trajopt planner and is meant to be functionally equivalent. Currently only the trajopt_sqp solver (equivalent to trajopt_sco) is supported via this planner. However, we may eventually set this up such that IPOPT of SNOPT can be used too.


mpowelson:
@Levi-Armstrong Just to keep you updated, I think this is nearing a workable state. The main thing left is the collision terms, but I think I will put that off until #426 is addressed.

Levi-Armstrong:
> @Levi-Armstrong Just to keep you updated, I think this is nearing a workable state. The main thing left is the collision terms, but I think I will put that off until #426 is addressed.
> 
> There are also the obvious things like acceleration, jerk, singularity avoidance, cost from error func, etc that are [not yet supported in trajopt_ifopt](https://github.com/ros-industrial-consortium/trajopt_ros/issues/183).

Thats great. Let me know if I can help with anything.

mpowelson:
The collision term skeleton has been added. However, it will require some work in trajopt before it is usable.

mpowelson:
@Levi-Armstrong Believe it or not, I think this is ready for review. It requires [trajopt #219](https://github.com/ros-industrial-consortium/trajopt_ros/pull/219) which has not been merged yet.

mpowelson:
Pushed a clang-tidy fix and deleted an unused cpp file

mpowelson:
@Levi-Armstrong It looks like it can't find tesseract_motion_planners-targets.cmake? [Here's the error](https://github.com/ros-industrial-consortium/tesseract/pull/443/checks?check_run_id=1631044489#step:6:732)

Levi-Armstrong:
@mpowelson Can you rebase on the latest?

mpowelson:
> @mpowelson Can you rebase on the latest?

Done

mpowelson:
@Levi-Armstrong This is ready for review. I don't know why Windows is upset.

Levi-Armstrong:
Ah, so the windows build is failing because there is no implementation for the toXML in the default composite and plan profiles. I would add them in the cpp and throw and exception if they are called until we have time to implement.

Levi-Armstrong (approved):

mpowelson:
Thanks for fixing this by the way.

PR #449, opened 2020-11-17
Update to leverage kinematics setLimits for environment changes

Levi-Armstrong:
This update code base to leverage the kinematics setLimits for environment commands that changes kinematic limits. This was required to solve the issue for opw and ikfast which do not store the scene_graph and did not what to change the interface. @mpowelson Do you mind testing to make sure this still works for you? I am currently testing it on my end to confirm it solves the issues we are having related to ikfast kinematics objects not having its limits updated.

### DIFF
--- a/tesseract/tesseract/src/tesseract.cpp
+++ b/tesseract/tesseract/src/tesseract.cpp
@@ -248,7 +248,7 @@ bool Tesseract::init(const tesseract_environment::Environment& env)
   clear();
 
   environment_ = env.clone();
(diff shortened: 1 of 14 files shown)


mpowelson:
I did a short test and it seems to work ok. No growing DOF for instance and the joint limits seem to get changed.

mpowelson (approved):
Looks good to me

Levi-Armstrong:
@mpowelson Thanks for the review.

PR #452, opened 2020-11-20
Fix issue with parsing OPW sign correction from SRDF.

Levi-Armstrong:
This PR also adds some error checking for Joint and State waypoint to make sure joint names and positions are the same length.

### DIFF
--- a/tesseract/tesseract_scene_graph/test/tesseract_scene_graph_unit.cpp
+++ b/tesseract/tesseract_scene_graph/test/tesseract_scene_graph_unit.cpp
@@ -13,6 +13,7 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 #include <tesseract_scene_graph/resource_locator.h>
 #include <tesseract_scene_graph/srdf_model.h>
 #include <tesseract_scene_graph/parser/kdl_parser.h>
+#include <tesseract_scene_graph/srdf/group_opw_kinematics.h>
 
 // getLinks and getJoint use an internal map so need to check against graph
 void checkSceneGraph(tesseract_scene_graph::SceneGraph& scene_graph)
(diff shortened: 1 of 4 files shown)


tesseract/tesseract_scene_graph/test/tesseract_scene_graph_unit.cpp
```diff
@@ -347,6 +348,42 @@ TEST(TesseractSceneGraphUnit, LoadSRDFUnit)  // NOLINT
...
+  EXPECT_NEAR(opw.offsets[4], 0.0, 1e-8);
+  EXPECT_NEAR(opw.offsets[5], 0.0, 1e-8);
+  EXPECT_EQ(opw.sign_corrections[0], 1);
+  EXPECT_EQ(opw.sign_corrections[1], 1);
+  EXPECT_EQ(opw.sign_corrections[2], 1);
+  EXPECT_EQ(opw.sign_corrections[3], -1);
+  EXPECT_EQ(opw.sign_corrections[4], 1);
+  EXPECT_EQ(opw.sign_corrections[5], 1);
+}
+
```
mpowelson, line 386: This test doesn't test when the OPW sign correction is wrong does it?

Levi-Armstrong: No, it is testing the issue with the parser of the xml. I was casting to a signed char which was not working if -1. So before this change the test would not pass but now it should.

Levi-Armstrong: I need to go through and create full set of unit tests for the SRDF similar to the URDF.

mpowelson: If you want to merge this before doing that, that is fine with me.

Levi-Armstrong: Lets wait. I should have time over the next month to go through and add these.

Levi-Armstrong:
@mpowelson Are you good with merging this one?

Levi-Armstrong:
Included in PR #453

PR #453, opened 2020-11-21
This add SFINAE utils for checking if a class passed to template has a member function.

Levi-Armstrong:
This build on PR #452. Once merged this will be rebased.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/core/waypoint.h
+++ b/tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/core/waypoint.h
@@ -33,10 +33,16 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <tinyxml2.h>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
+#include <tesseract_common/sfinae_utils.h>
+
 namespace tesseract_planning
 {
-namespace detail
+namespace detail_waypoint
 {
+CREATE_MEMBER_CHECK(getType);
+CREATE_MEMBER_CHECK(print);
+CREATE_MEMBER_CHECK(toXML);
+
 struct WaypointInnerBase
 {
   WaypointInnerBase() = default;
@@ -62,7 +68,12 @@ struct WaypointInnerBase
 template <typename T>
 struct WaypointInner final : WaypointInnerBase
 {
-  WaypointInner() = default;
+  WaypointInner()
+  {
+    static_assert(has_member_getType<T>::value, "Class does not have member function 'getType'");
+    static_assert(has_member_print<T>::value, "Class does not have member function 'print'");
+    static_assert(has_member_toXML<T>::value, "Class does not have member function 'toXML'");
+  }
   ~WaypointInner() override = default;
   WaypointInner(const WaypointInner&) = delete;
   WaypointInner(WaypointInner&&) = delete;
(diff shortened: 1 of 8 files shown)


Levi-Armstrong:
@mpowelson I add the SFINAE utils for checking if a member function exist. Take a look at Instruction and Waypoint class on how I added to the classes for the CloneCache.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/core/waypoint.h
```diff
@@ -62,7 +68,12 @@ struct WaypointInnerBase
 template <typename T>
 struct WaypointInner final : WaypointInnerBase
 {
-  WaypointInner() = default;
+  WaypointInner()
+  {
+    static_assert(has_member_Waypoint_getType<T>::value, "Class does not have member function 'getType'");
+    static_assert(has_member_Waypoint_print<T>::value, "Class does not have member function 'print'");
+    static_assert(has_member_Waypoint_toXML<T>::value, "Class does not have member function 'has_member_toXML'");
```
mpowelson, line 75: typo: `Class does not have member function 'has_member_toXML `should be `Class does not have member function 'toXML'`

mpowelson (changes requested):
Looks good as far as I can tell. This is some black magic. I think we should probably have a couple of unit tests for this before we use it on anything important.

tesseract/tesseract_planning/tesseract_command_language/include/tesseract_command_language/core/waypoint.h
Levi-Armstrong: Yea, not sure how to create a test for something like this where it causes compiler failures; any thoughts? I did tested it out in godbolt to verify that it works.

Levi-Armstrong: I think I have a solution for a unit test. Testing now.

mpowelson: Oh I just meant make 2 test classes and see if checking for that method works

Levi-Armstrong: Yea, that is what I did. In the process I did some testing with Instruction and Waypoint and realized I forgot to add the static asserts to the move and copy constructor. In addition I think there is also a way to leverage enable_if_t in the template definition, but not sure how to handle multiple conditions. Though I like the static assert because you get a meaningful message but you have to make sure to add them to all available constructors where the other automatically works for all.

``` c++
template <typename T, std::enable_if_t<has_member_getType<T>::value, bool> = true>
```

Levi-Armstrong:
@mpowelson I added a unit test to tesseract_common to verify the macro works and fixed Instruction and Waypoint.

mpowelson (approved):

Levi-Armstrong:
@mpowelson I wanted to check before merging. Did you review all commits? There first two were in another PR which is OK if you reviewed them here, just did not want to merge if still reviewing the first two commits in the other PR.

mpowelson:
I reviewed them, but you said you wanted to wait until you added the SRDF tests. I'm fine either way, but if you do decide to wait with the other PR I think we should cherrypick this one.

Levi-Armstrong:
> I reviewed them, but you said you wanted to wait until you added the SRDF tests. I'm fine either way, but if you do decide to wait with the other PR I think we should cherrypick this one.

Oh, sorry I meant I will add the unit tests later and not as part of the current PR.

mpowelson:
@Levi-Armstrong Can this be used with templates? I want to do something like

```

template <typename T>
class CloneCache
{
public:
    CREATE_MEMBER_FUNC_SIGNATURE_CHECK(update, bool, const std::shared_ptr<const typename T>&);

```

Levi-Armstrong:
It should work.

mpowelson:
I'm gettimg
```
In file included from /home/matthew/code/robot/src/tesseract/tesseract/tesseract_common/include/tesseract_common/clone_cache.h:39:0,
                 from /home/matthew/code/robot/src/tesseract/tesseract/tesseract_common/test/clone_cache_unit.cpp:6:
/home/matthew/code/robot/src/tesseract/tesseract/tesseract_common/include/tesseract_common/sfinae_utils.h:166:13: error: declaration of template parameter 'T' shadows template parameter
   template <typename T, typename = std::true_type>  

```

mpowelson:
Nevermind. I just noticed you are using T as well. I had to change my template to be something else

Issue #421, opened 2020-10-21
Consider making tesseract_planning its own repo.


mpowelson:
We already split tesseract_ros/ros2 into their own repos. tesseract_planning has a lot of dependencies that core tesseract does not need. Further, I would expect that tesseract_planning will continue to be under significant development for a while, but hopefully core tesseract will be more stable. I see this as particularly useful if we can address #419. Then developers could build just tesseract_planning.

PR #454, opened 2020-11-24
Add sfinae function signature checks

Levi-Armstrong:
This add several sfinae function signature checks. In order to accomplish without a lot of work it required upgrading to c++17.

Also includes a fix to tesseract collision example.


Levi-Armstrong:
This PR should be ready for review. I forgot to add the function signature check to the command language.

README.md
```diff
@@ -6,7 +6,6 @@ Platform             | CI Status
 ---------------------|:---------
 Linux (Focal)        | [![Build Status](https://github.com/ros-industrial-consortium/tesseract/workflows/Focal-Build/badge.svg)](https://github.com/ros-industrial-consortium/tesseract/actions)
 Linux (Bionic)       | [![Build Status](https://github.com/ros-industrial-consortium/tesseract/workflows/Bionic-Build/badge.svg)](https://github.com/ros-industrial-consortium/tesseract/actions)
-Linux (Xenial)       | [![Build Status](https://github.com/ros-industrial-consortium/tesseract/workflows/Xenial-Build/badge.svg)](https://github.com/ros-industrial-consortium/tesseract/actions)
```
mpowelson, line 9: Gasp!

mpowelson (commented):
These are pretty confusing, but I think there are correct. Not something the average user will touch anyway.

mpowelson:
We could also consider moving this into tesseract_planning if #421 ever happened. That might allow us to keep tesseract_core as a higher quality level than tesseract_planning since I think C++17 probably hurts us in that regard.

README.md
Levi-Armstrong: > These are pretty confusing, but I think there are correct. Not something the average user will touch anyway.

They definitely are. It was a lot of trial and error.

Levi-Armstrong:
@mpowelson Do you approve of the changes?

mpowelson (approved):

Issue #459, opened 2020-12-01
OFKT state solver does not update joint limits


mpowelson:
Using `Tesseract->getEnvironment()->changeJointPositionLimits()` the limits get changed in the scene graph and it looks like in the limits_ MatrixX2d in the solver. However, the nodes_ store their own limits and those do not get updated.

This is the command in ofkt
```
      case tesseract_environment::CommandType::CHANGE_JOINT_POSITION_LIMITS:
      {
        const auto& cmd = static_cast<const tesseract_environment::ChangeJointPositionLimitsCommand&>(*command);
        const std::unordered_map<std::string, std::pair<double, double>>& limits = cmd.getLimits();
        // Loop through all names until we find the one we need
        for (std::size_t i = 0; i < joint_names_.size(); i++)
        {
          auto it = limits.find(joint_names_[i]);
          // Assign the lower/upper. Velocity, acceleration, and effort are ignored
          if (it != limits.end())
          {
            limits_(static_cast<Eigen::Index>(i), 0) = it->second.first;
            limits_(static_cast<Eigen::Index>(i), 1) = it->second.second;
          }
        }

        break;
      }
```

mpowelson:
@Levi-Armstrong How do you think we should fix this? I kind of think we should just remove limits_ from the revolute node. None of the other joint types have them, and I kind of feel that joint limits should be handled higher up in the environment and the solver should just solve whatever it is given.

Levi-Armstrong:
I do not have issue with the change, though not sure what the impact is with the change since it is at a low level.

Levi-Armstrong:
Wait are you talking about removing it from the OFKT Revolute Node? If so that should not be an issue.

mpowelson:
Yes. That's what I was wanting to do. All it is used for is clamping the inputs to joint limits.

Levi-Armstrong:
I am good with you removing it.

PR #460, opened 2020-12-01
Remove limits from revolute nodes in OFKT state solver

mpowelson:
Addresses #459


Levi-Armstrong (approved):

mpowelson:
Note that the kdl state solver also doesn't do these checks

PR #462, opened 2020-12-04
Make getShortestPath const

haudren:
Indeed, it does not (and should not) modify the underlying graph.

I noticed this while trying to compute link paths on an existing environment, where the SceneGraph is `const`. I'm not sure why it wasn't const in the first place, so my suggestion might not be valid. In any case don't hesitate to modify this branch as you see fit.

### DIFF
--- a/tesseract/tesseract_scene_graph/include/tesseract_scene_graph/graph.h
+++ b/tesseract/tesseract_scene_graph/include/tesseract_scene_graph/graph.h
@@ -369,7 +369,7 @@ public:
    */
   void saveDOT(const std::string& path) const;
 
-  Path getShortestPath(const std::string& root, const std::string& tip);
+  Path getShortestPath(const std::string& root, const std::string& tip) const;
 
   // static inline Graph copyGraph(const Graph& graph)
   //{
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
Thank you for the contribution, it most likely was an oversight on my part. All tests pass so it should be good. I will port this change to the feature/CommandLanguage branch.

Just heads up, we will be merging the feature/CommandLanguage branch into master sometime late December or early January. I will create a tag for the current state of master prior to merging.

haudren:
Thank you for coming back to me! If I understand well you don't intend to merge it to master? This commandLanguage feature looks cool but wouldn't it be better to add that change directly in master for now as it's independent of any other change?

Oh and btw, I'm guessing there will be some API breakage when that branch lands, but what about compatibility with ROS1 / C++14 ?

Levi-Armstrong:
> Thank you for coming back to me! If I understand well you don't intend to merge it to master? This commandLanguage feature looks cool but wouldn't it be better to add that change directly in master for now as it's independent of any other change?

I will merge into master.

> Oh and btw, I'm guessing there will be some API breakage when that branch lands, but what about compatibility with ROS1 / C++14 ?

It will have ROS1 and ROS2 support and C++17.

PR #463, opened 2020-12-08
Add virtual destructor to ProcessInfo

mpowelson:
Need to make it polymorphic so we can cast it back


tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_input.h
```diff
@@ -47,6 +47,8 @@ class ProcessInfo
 
   ProcessInfo(std::size_t unique_id, std::string name = "") : unique_id(unique_id), message(std::move(name)) {}
 
+  virtual ~ProcessInfo() {}
```
Levi-Armstrong, line 50: Recommend '= default'

Levi-Armstrong (approved):
Good to merge after CI passes

Levi-Armstrong (approved):

PR #467, opened 2020-12-09
Add forgotten include in opw_inv_kin.h

haudren:
Indeed, the header uses CONSOLE_BRIDGE_logError but does not include the
relevant header, which means that including only that header causes
compilation errors.

I encountered this while trying to write a simple program that would only load some OPW kinematics parameters and check them... Again, please feel free to modify that branch to suit your liking!


Levi-Armstrong (approved):

Levi-Armstrong:
Note to self: This has been fixed in feature/CommandLanguage

PR #472, opened 2020-12-14
Clean up warnings related to setContactDistanceThreshold

Levi-Armstrong:
This PR also add two additional setters to the Collision Base Class `setDefaultCollisionMarginData` and `setPairCollisionMarginData` making it a little easier to set collision margin data. 

Also fixes the handling of collision margin data in environment.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/problem_generators/default_problem_generator.cpp
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/problem_generators/default_problem_generator.cpp
@@ -27,6 +27,7 @@
 #include <tesseract_motion_planners/trajopt/problem_generators/default_problem_generator.h>
 #include <tesseract_motion_planners/trajopt/profile/trajopt_default_composite_profile.h>
 #include <tesseract_motion_planners/trajopt/profile/trajopt_default_plan_profile.h>
+#include <tesseract_motion_planners/trajopt/profile/trajopt_default_solver_profile.h>
 #include <tesseract_motion_planners/core/utils.h>
 #include <tesseract_motion_planners/planner_utils.h>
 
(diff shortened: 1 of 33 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/problem_generators/default_problem_generator.cpp
```diff
@@ -58,6 +60,16 @@ trajopt::TrajOptProb::Ptr DefaultTrajoptProblemGenerator(const std::string& name
     throw std::runtime_error(error_msg);
   }
 
+  // Apply Solver parameters
+  std::string profile = request.instructions.getProfile();
+  profile = getProfileString(profile, name, PlannerProfileRemapping());
+  TrajOptSolverProfile::ConstPtr solver_profile =
+      getProfile<TrajOptSolverProfile>(profile, solver_profiles, std::make_shared<TrajOptDefaultSolverProfile>());
+  if (!solver_profile)
+    throw std::runtime_error("TrajOptSolverConfig: Invalid profile");
```
mpowelson, line 69: Just so I understand, the solver profile is coming from the composite instruction, right? So why not just put the solver in the composite profile?

mpowelson (commented):
I gave this a quick look. It looks fine to me. However I didn't check all of the functions in detail like the onCollisionMarginDataChanged functions for instance, if you are concerned about the specifics of those functions you might want to get someone else to look over them.

tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/problem_generators/default_problem_generator.cpp
Levi-Armstrong: > Just so I understand, the solver profile is coming from the composite instruction, right? So why not just put the solver in the composite profile?

Yea it uses the highest level composite profile to determine the solver profile. I was not sure if it should go in the composite profile or not. Do you see any reason not to include it in the composite profile?

Levi-Armstrong:
> onCollisionMarginDataChanged functions for instance, if you are concerned about the specifics of those functions you might want to get someone else to look over them.

Yea this was a copy paste so it should still be good. Just reducing code duplication.

tesseract/tesseract_planning/tesseract_motion_planners/src/trajopt/problem_generators/default_problem_generator.cpp
Levi-Armstrong: I will merge when CI passes, but let me know if your thought on where to locate the parameters. I can create another PR to move them.

PR #474, opened 2020-12-15
Add external tool center point support

Levi-Armstrong:
This PR depends on #472


gavanderhoorn:
Really nice to see this.

High-level comment: there is a lot of `if (isExternal) do_this(); else doThat();`.

Would a polymorphism based approach perhaps result in cleaner code? From a quick glance it looks like most of the differences between the execution paths are behaviour-related, which could be good candidates for overloading.

Levi-Armstrong:
> Really nice to see this.
> 
> High-level comment: there is a lot of `if (isExternal) do_this(); else doThat();`.
> 
> Would a polymorphism based approach perhaps result in cleaner code? From a quick glance it looks like most of the differences between the execution paths are behaviour-related, which could be good candidates for overloading.

@gavanderhoorn Thank you for the feedback. Do you mind providing more detail on what you think might be possible and I will see what I can do?

Levi-Armstrong:
@gavanderhoorn I am going to merge, but would like to make the changes you have recommend so please do comment when free. I will create a PR new PR with the changes.

PR #483, opened 2020-12-19
Improve cpp code for python

Levi-Armstrong:
This move several header implementation into a cpp file along with remove lambda definition in headers to provided better support for python wrapping. @johnwason 

@mpowelson I found a bug in tesseract_collision processResult function where it was checking against a pairs contact distance but was not returning if it was not a valid contact given the pairs contact distance. Do you agree with these changes?


Levi-Armstrong:
This will also enable the ability for process generator to leverage taskflow subtaskflow which is the next step for the motion planners. The motion planners like descartes and ompl which are also multi-threaded would generate a subtaskflow in the graph so all resources are managed by the same thread pool.

Levi-Armstrong:
This has been tested on a system and no issues have been found.

Levi-Armstrong:
Merging, but please still review and request changes.

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
```diff
@@ -119,24 +119,21 @@ inline ContactResult* processResult(ContactTestData& cdata,
   if (cdata.req.is_valid && !cdata.req.is_valid(contact))
     return nullptr;
 
+  if (!(contact.distance < cdata.collision_margin_data.getPairCollisionMarginData(key.first, key.second)))
+    return nullptr;
+
```
mpowelson, line 124: Should this go inside the if (!found) to avoid the lookup?

mpowelson:
@Levi-Armstrong I don't see the collision checking bug. There is a return at the bottom that should have caught all of those cases. That said, this is more concise.

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
```diff
@@ -119,24 +119,21 @@ inline ContactResult* processResult(ContactTestData& cdata,
...
-      if (contact.distance < cdata.collision_margin_data.getPairCollisionMarginData(key.first, key.second))
-      {
-        data.reserve(100);  // TODO: Need better way to initialize this
-        data.emplace_back(contact);
-      }
+      data.reserve(100);  // TODO: Need better way to initialize this
+      data.emplace_back(contact);
     }
 
     return &(cdata.res->insert(std::make_pair(key, data)).first->second.back());
```
Levi-Armstrong, line 139: The bug was due to this line which would access the data object back element where for the condition that `contact.distance` is not less than the margin` the vector would be empty causing a segfault. Moving the check to the top and return null if not valid prevented this from occurring.

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
Levi-Armstrong: I don't believe so because every if branch seems to be performing this check or am I missing one?

mpowelson: With the current layout I think you would need 2. One inside `if(!found)` and one afterward. The thought was that if it is a FIRST type and found = true, then we don't want to call getPairCollisionMarginData because it doesn't matter.

PR #485, opened 2020-12-21
Test building windows in debug

Levi-Armstrong:
This is just a test to see if the build fails when building in debug.


johnwason:
Windows debug builds you different standard library DLLs than release builds. This means that you can't mix DLLs built in release and debug build configurations. The solution is to use "Release with Debug Mode". (CMAKE_BUILD_TYPE=RelWithDebInfo). This builds with debug symbols enabled, but uses the release DLL. I think that the ROS Windows tool build with this mode by default.

https://docs.microsoft.com/en-us/cpp/build/how-to-debug-a-release-build?view=msvc-160

Levi-Armstrong:
Thank you @johnwason.

Levi-Armstrong:
This was just a test closing.

PR #486, opened 2020-12-21
Get Robot Redundancy

DavidMerzJr:
@Levi-Armstrong This is the feature implemented on the project machine.

### DIFF
--- a/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/robot_config.h
+++ b/tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/robot_config.h
@@ -128,6 +128,21 @@ inline RobotConfig getRobotConfig(const tesseract_kinematics::ForwardKinematics:
 
   return RobotConfig::NDB;
 }
+template <typename FloatType>
+inline Eigen::Vector3i getRobotRedundancy(const Eigen::Ref<const Eigen::Matrix<FloatType, Eigen::Dynamic, 1>>& joint_values,
+                                     const Eigen::Ref<const Eigen::Vector2i>& sign_correction = Eigen::Vector2i::Ones())
+{
+
+  Eigen::Vector3i redundancy;
+
+  int red_j_4 = int(joint_values(3) / M_PI);
+  int red_j_5 = int(sign_correction[1] * joint_values(4) / M_PI);
+  int red_j_6 = int(joint_values(5) / M_PI);
+
+  redundancy << red_j_4, red_j_5, red_j_6;
+  return redundancy;
+}
+
 }  // namespace tesseract_planning
 
 #endif  // TESSERACT_COMMAND_LANGUAGE_ROBOT_CONFIG_H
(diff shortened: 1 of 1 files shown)


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/robot_config.h
```diff
@@ -128,6 +128,22 @@ inline RobotConfig getRobotConfig(const tesseract_kinematics::ForwardKinematics:
 
   return RobotConfig::NDB;
 }
+
+template <typename FloatType>
+inline Eigen::Vector3i
+getRobotRedundancy(const Eigen::Ref<const Eigen::Matrix<FloatType, Eigen::Dynamic, 1>>& joint_values,
```
mpowelson, line 134: I'm not sure who is owning this code, but I'd request doxygen. This looks a bit robot specific. Why are Vector3i and Vector2i for instance? I would have expected the function to be 
```
inline  Eigen::Matrix<FloatType, Eigen::Dynamic, 1>
getRobotRedundancy(const Eigen::Ref<const Eigen::Matrix<FloatType, Eigen::Dynamic, 1>>& joint_values,
const Eigen::Ref<const Eigen::Matrix<FloatType, Eigen::Dynamic, 1>>& sign_correction = ...)
```

Levi-Armstrong: I agree need documentation. This is to go with the robotConfig which is for a 6DOF industrial manipulator. We will get the documentation added explaining the different inputs and outputs.

mpowelson: It just seems odd that it only outputs joints 4,5,6 and the sign correction is a vector2i but only the second term is used. I'd expect a sign correction for each joint and an output for each joint. They could be Vector6i if it is only for 6 dof. However you should be able to vectorized this with Eigen for any number of joints

Levi-Armstrong: @colin-lewis-19 Can you add API documentation?

Levi-Armstrong: [This](https://www.onerobotics.com/posts/2013/the-perils-of-six-axis-robots/) is a good read. There are only 3, one for each config variable where that configuration can be defined multiple redundant positions. Though not sure why the return variable are labeled 4, 5 and 6 which is miss leading as they are not related to joint 4, 5 and 6 directly.

colin-lewis-19: I've added documentation and clarified naming from joints -> redundancies. I also fixed a misunderstanding on my part and now check joint 1 for redundancy 1, instead of joint 4, based on the article linked above.

However, this PR was opened by David and I am unable to push my changes to his fork.

Levi-Armstrong (approved):

PR #490, opened 2020-12-22
Improved mesh parsing

johnwason:
This pull request contains initial work on extracting additional information from mesh files. It is not ready for merge, but I am opening a pull request so we can discuss the improvements.

Additional information includes:

* Vertex normals
* Vertex colors
* Material properties, focusing on PBR metallic model
* Diffuse textures. The PBR models supports textures for metalness, roughness, and emissivity, but these aren't important for the CAD level rendering necessary for robotics applications. The diffuse textures are intended to allow for use of decals and fiducial marks.

The improved parser will allow including this information in ROS and Robot Raconteur messages for viewers, eliminating the need to have the logic to locate mesh files on the filesystem as is currently necessary for displaying these features in a viewer or simulation.

More information on PBR Metallic workflow: https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_010_Materials.md

The


johnwason:
I finished updating the tesseract_viewer_python package to use the mesh material and texture data. An example:

https://johnwason.github.io/tesseract_viewer_python_example6/

It looks great in VR if you have a headset.

This pull request should be ready to merge. I can do a squash once the review is complete.

johnwason:
Looks like we need a newer version of assimp on bionic for this to work. Can we include a newer version in tesseract_ext?

Levi-Armstrong:
Do you know what version of assimp you need?

johnwason:
Looks like I need version 5.0.0 or greater.

Levi-Armstrong:
Would it be possible to disable pbr materials for versions less than 5.0? I do a similar thing in tesseract_kinematics where I check the version and set a compile definition. The concern I have on bionic if we use a custom version it is likely to cause runtime issues when linking against ogre and rviz because I believe they also link against assimp I believe.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -44,16 +47,20 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <assimp/scene.h>
 #include <assimp/Importer.hpp>
 #include <assimp/postprocess.h>
+#include <assimp/pbrmaterial.h>
```
Levi-Armstrong, line 50: I think we could do something similar to what is done for tesseract_kinematics [here](https://github.com/ros-industrial-consortium/tesseract/blob/e9d6289a864997eb21bea8c0c93343e1404fb0c9/tesseract/tesseract_kinematics/CMakeLists.txt#L84) where we check the version and if less than 5.0.0 set a compiler definition and use a default pbr settings.

johnwason:
Mesh loading on Windows is now fixed. I remember fixing the problem I found, but I guess it never made it to upstream.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: I added conditional building to disable PBR with older versions of ASSIMP.

tesseract/tesseract_common/include/tesseract_common/types.h
```diff
@@ -54,6 +54,7 @@ using AlignedUnorderedMap = std::unordered_map<Key,
 using VectorIsometry3d = AlignedVector<Eigen::Isometry3d>;
 using VectorVector4d = AlignedVector<Eigen::Vector4d>;
 using VectorVector3d = std::vector<Eigen::Vector3d>;
+using VectorVector2d = std::vector<Eigen::Vector2d>;
```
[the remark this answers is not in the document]
johnwason: Done

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
```diff
@@ -117,9 +119,12 @@ class SimpleLocatedResource : public tesseract_common::Resource
 
   std::shared_ptr<std::istream> getResourceContentStream() override;
 
+  tesseract_common::Resource::Ptr locateSubResource(const std::string& relative_path) override;
+
 protected:
   std::string url_;
   std::string filename_;
+  std::weak_ptr<SimpleResourceLocator> parent_;
```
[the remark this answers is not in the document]
johnwason: If you have two shared_ptr objects that point to each other, they will never be deleted because they created a mutual reference to each other. This is called a "reference loop" or "circular reference". weak_ptr is a weak reference, that stores a reference to the object, but does not increase the reference count. If the reference count goes to zero, weak_ptr will return nullptr. I am using it here because there is potentially a circular reference between the resource locator and the located resource.

johnwason:
I implemented loading normals from the mesh files. This required some changes to the unit test, since it created more vertices to handle the normal vectors. @Levi-Armstrong Can you use the normal vectors with the collision libraries? Do they have parameters to accept the normal information?

tesseract/tesseract_scene_graph/include/tesseract_scene_graph/resource_locator.h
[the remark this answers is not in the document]
Levi-Armstrong: Ah, good to know. Thank you for the information.

Levi-Armstrong:
> Can you use the normal vectors with the collision libraries? Do they have parameters to accept the normal information?

I do not believe so. Bullet only requires vertices and FCL requires vertices and faces. Though I am not sure why FCL requires faces and Bullet does not, I just assumed they are using a slightly different algorithm.

johnwason:
Assuming the unit tests pass this should be ready to merge. I'd like to get it in before the release.

I think you can just select "squash and merge" instead of me force pushing a squash.

Levi-Armstrong:
@johnwason Can you rebase on the latest? It should be ready to merge after that.

johnwason:
Ok, I merged in latest. The merge commit will not show if you do a squash merge of the pull request.

tesseract/tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
```diff
@@ -22,7 +22,7 @@ TEST(TesseractURDFUnit, parse_mesh)  // NOLINT
     EXPECT_FALSE(status->message().empty());
     EXPECT_TRUE(geom.size() == 1);
     EXPECT_TRUE(geom[0]->getTriangleCount() == 80);
-    EXPECT_TRUE(geom[0]->getVerticeCount() == 42);
+    EXPECT_TRUE(geom[0]->getVerticeCount() == 240);
```
Levi-Armstrong, line 25: Why the vertices count change?

johnwason: I made changes to the assimp loader to include normals. I think that this is stopping vertices from being merged. Some vertices may have the same location, different faces may have different normals. This prevents the vertices from being merged. STL files include normals so I had to modify the unit tests to accept the higher vertices count.

I noticed a significant improvement in the quality of rendering when normals were included. Before, when normals were not specified, the rendered edges were very blurry and hard to see. With the normals, the edges are very clear. This video has an overview of normal map textures that shows how the normals are used to simulate depth: https://www.youtube.com/watch?v=6_-NNKc4lrk (I am not loading normal maps for our purposes, but the video shows how the shaders use the normals). I think that the renderer was doing an interpolation to generate normal vectors leading to bad edges.

Levi-Armstrong: I assume you are referring to the edge of the triangles now are sharper? We need to make sure this only occurs for certain mesh types. This would be bad if it did this for convex_mesh types, if use the urdf version 2 it would cause problems because it does not automatically convert to convex hull. The default version does which is most likely why the tests are unaffected.

tesseract/tesseract_geometry/include/tesseract_geometry/impl/mesh_material.h
```diff
@@ -0,0 +1,185 @@
...
+   * @return The emissive factor in RGBA
+   */
+  Eigen::Vector4d getEmissiveFactor() const { return emissiveFactor_; }
+
+private:
+  // Mesh material based on simplified glTF 2.0 pbrMetallicRoughness parameters
+  Eigen::Vector4d baseColorFactor_;
+  double metallicFactor_ = 0;
+  double roughnessFactor_ = 0.5;
+  Eigen::Vector4d emissiveFactor_;
```
Levi-Armstrong, line 127: ROS style
```
  Eigen::Vector4d base_color_factor_;
  double metallic_factor_ = 0;
  double roughness_factor_ = 0.5;
  Eigen::Vector4d emissive_factor_;
```

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -169,11 +327,10 @@ std::vector<std::shared_ptr<T>> createMeshFromPath(const std::string& path,
 
   // Issue #38 fix: as part of the post-processing, we remove all other components in file but
   // the meshes, as anyway the resulting shapes:Mesh object just receives vertices and triangles.
+  // John Wason Dec 2020 - Load more information
```
Levi-Armstrong, line 330: Should we update to include vertices, triangles, normals colors and textures?

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -258,11 +417,10 @@ std::vector<std::shared_ptr<T>> createMeshFromResource(tesseract_common::Resourc
...
```
Levi-Armstrong, line 420: Same here.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
Levi-Armstrong: Should the commented out aiComponent_NORMALS be removed?

tesseract/tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
Levi-Armstrong: Maybe we pass in a flag which indicates if it is for visual or collision. If for visual we process it this way and for collision we process it the previous way?

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -45,15 +48,22 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
 #include <regex>
+#include <boost/filesystem/path.hpp>
 
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
+#include <tesseract_geometry/impl/mesh_material.h>
+
 namespace tesseract_geometry
 {
 /**
```
Levi-Armstrong, line 69: Recommend updating the documentation below.

tesseract/tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
Levi-Armstrong: As you suggested it looks like assimp does not perform this process `aiProcess_JoinIdenticalVertices` if normal's are requested to be parsed as outlined [here](`https://github.com/assimp/assimp/issues/407). I believe the previous approach should solve the issue. Then during processing at the urdf the parameters can be passed based on if it is a visual or collision geometry.

```
template <class T>
std::vector<std::shared_ptr<T>> createMeshFromPath(const std::string& path,
                                                   Eigen::Vector3d scale = Eigen::Vector3d(1, 1, 1),
                                                   bool triangulate = false,
                                                   bool flatten = false,
                                                   bool normals = false,
                                                   bool material_and_texture = false)
```

johnwason: Ok I will add the extra options

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: The property `AI_CONFIG_PP_RVC_FLAGS` is a mask flag, so if the flag is present, it disables that feature from being loaded. I am going to modify this section to set the flags based on the options you suggested in the other comment.

johnwason:
@Levi-Armstrong Do you have an example mesh with textures and material? I need to add another unit test loading a mesh file.

Levi-Armstrong:
> @Levi-Armstrong Do you have an example mesh with textures and material? I need to add another unit test loading a mesh file.

I do not have anything with texture, but one of the dae files I assume should have material.

tesseract/tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
johnwason: This has been changed back to the old vertex count for collision only, and the new count when visual enabled. Tesseract itself doesn't use the visual so it shouldn't have any effect on the collision checkers and planners.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: This section has been updated to use the additional `normals`, `vertex_colors`, and `material_and_texture` arguments.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: Also updated to use additional arguments

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: I have added argument documentation to the doxygen headers.

johnwason:
@Levi-Armstrong never mind I created a simple test mesh.

johnwason:
Here is a screenshot of the test mesh:

image

johnwason:
I haven't merged the unit test yet

PR #492, opened 2020-12-24
Add clang-tidy and python badge

Levi-Armstrong:


Levi-Armstrong:
@johnwason What versions of python are supported? This PR adds a badge for showing python support but was not sure what version/versions to add.

johnwason:
It should work with Python 2.7 and Python 3 versions 3.6 and greater

PR #493, opened 2020-12-28
Remove deprecated collision class methods and utility functions

Levi-Armstrong:


tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_profile.h
```diff
@@ -52,10 +52,10 @@ class DescartesPlanProfile
...
   DescartesPlanProfile() = default;
   virtual ~DescartesPlanProfile() = default;
-  DescartesPlanProfile(const DescartesPlanProfile&) = default;
-  DescartesPlanProfile& operator=(const DescartesPlanProfile&) = default;
-  DescartesPlanProfile(DescartesPlanProfile&&) noexcept = default;
-  DescartesPlanProfile& operator=(DescartesPlanProfile&&) noexcept = default;
+  DescartesPlanProfile(const DescartesPlanProfile&) = delete;
+  DescartesPlanProfile& operator=(const DescartesPlanProfile&) = delete;
+  DescartesPlanProfile(DescartesPlanProfile&&) noexcept = delete;
+  DescartesPlanProfile& operator=(DescartesPlanProfile&&) noexcept = delete;
```
mpowelson, line 58: Why this change? Should we add a clone?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/deserialize.h
```diff
@@ -34,25 +34,25 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_planning
 {
-TrajOptDefaultPlanProfile trajOptPlanParser(const tinyxml2::XMLElement& xml_element);
+TrajOptDefaultPlanProfile::Ptr trajOptPlanParser(const tinyxml2::XMLElement& xml_element);
```
mpowelson, line 37: Why this change?

tesseract/tesseract_planning/tesseract_motion_planners/test/serialize_test.cpp
```diff
@@ -45,164 +45,175 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+                                                                                                                  "ult_"
+                                                                                                                  "plan"
+                                                                                                                  "_"
+                                                                                                                  "exam"
+                                                                                                                  "ple_"
+                                                                                                                  "inpu"
+                                                                                                                  "t."
+                                                                                                                  "xm"
+                                                                                                                  "l");
```
mpowelson, line 163: What's up with this?

tesseract/tesseract_planning/tesseract_motion_planners/test/serialize_test.cpp
```diff
@@ -45,164 +45,175 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+                                                                 "s_"
+                                                                 "defa"
+                                                                 "ult_"
+                                                                 "plan"
+                                                                 "_exa"
+                                                                 "mple"
+                                                                 "_inp"
+                                                                 "ut."
+                                                                 "xm"
+                                                                 "l");
```
mpowelson, line 211: Same here. Is this the new clang format version?

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/simple_process_manager.h
```diff
@@ -1,92 +0,0 @@
...
- * execute them efficiently in a parallel based on those dependencies.
- *
- * The required format is below.
- *
- * Composite
- * {
- *   ... Must not contain any child composite instructions
- * }
- */
-class SimpleProcessManager : public ProcessManager
```
mpowelson, line 55: I guess you're thinking we are ready to remove this? I have not used the planning server yet. That provides the same functionality?

README.md
```diff
@@ -121,8 +122,8 @@ Tesseract currently leverages Compiler Warnigs, Clang Tidy and Code Coverage. Al
   - cppcoreguidelines-no-malloc
   - cppcoreguidelines-slicing
   - cppcoreguidelines-special-member-functions
-  - misc-*
-  - modernize-*
+  - misc-*,-misc-non-private-member-variables-in-classes
```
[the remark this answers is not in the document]
Levi-Armstrong: It does the `-` before `misc-non-private-member-variables-in-classes` indicates to not perform this check. It was added in the new version. This configures the new version to the same as the previous version being used. We will need to decide if we want these checks or not.

tesseract/tesseract_collision/src/fcl/fcl_discrete_managers.cpp
```diff
@@ -63,7 +63,7 @@ DiscreteContactManager::Ptr FCLDiscreteBVHManager::clone() const
     manager->addCollisionObject(cow.second->clone());
 
   manager->setActiveCollisionObjects(active_);
-  manager->setContactDistanceThreshold(collision_margin_data_.getMaxCollisionMargin());
+  manager->setCollisionMarginData(collision_margin_data_);
```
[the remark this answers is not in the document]
Levi-Armstrong: It called the same function under the hood which is way it was still working along with we do not have any unit tests that leverage the contact distance for a given pair which would have failed for this case.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_profile.h
Levi-Armstrong: This is to prevent slicing. At the moment we are not making copies so the clone was not needed, but may need to added it in the future.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/deserialize.h
Levi-Armstrong: For python, because the copy and move constructor is now deleted it required it to be a shared pointer.

tesseract/tesseract_planning/tesseract_motion_planners/test/serialize_test.cpp
Levi-Armstrong: No idea. I think clang format may have done this.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/simple_process_manager.h
Levi-Armstrong: I thought it would be good to remove before merging into master, since the planner server now provides the same functionality. Mainly to prevent other from using it. Do you want to keep it?

README.md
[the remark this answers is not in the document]
mpowelson: Ah. Gotcha

tesseract/tesseract_collision/src/fcl/fcl_discrete_managers.cpp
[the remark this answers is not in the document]
mpowelson: Oh right. Yeah I seem to remember something about that. FCL only uses the max anyway and the pairs are handled in the process results

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_profile.h
mpowelson: So this is because we are storing the base class in the new planning server? Do you have to do this since you changed them to be Ptrs below? It just seems like copying and modifying them might be a pretty common thing you might want to do. For example if you want to make a couple of profiles that are all the same except for one thing.

Levi-Armstrong: I could enable them for the implementation classes to allow this but mark them final. What do you think.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/simple_process_manager.h
mpowelson: You can remove it.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_profile.h
mpowelson: I think easily copying them is important whether we add a clone or mark them final. I'm a bit confused as to how marking them final helps though.

Levi-Armstrong: It effects others when inheriting from this class where they cannot overload, but it does not remove the possibility of slicing. I think it is low risk to that an intermediate type would be copied so I will add back the constructors for the implementation classes.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/simple_process_manager.h
Levi-Armstrong: The only thing I have noticed with the planning server is that having a single executor it does not release memory for some cases. I have created a test case and posted an issues [here](https://github.com/taskflow/taskflow/issues/264).

mpowelson: Interesting. That does seem like a pretty big deal actually - though probably a taskflow issue. So if we call make_shared in the taskflow anywhere that is essentially a memory leak?

mpowelson: Actually, that is really bizarre. Have you tried running valgrind on it?

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/profile/descartes_profile.h
Levi-Armstrong: Also forgot that having pure virtual methods prevents slicing at the base class but could still occur for intermediate types but that would be on the developer. I will go through and clean up similar classes which a have pure virtual methods in the base class and allow derived classes to be copied.

Levi-Armstrong: Finally came back across these. I will need to implement custom copy and assignment.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c67-a-polymorphic-class-should-suppress-copying

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c130-for-making-deep-copies-of-polymorphic-classes-prefer-a-virtual-clone-function-instead-of-copy-constructionassignment

Levi-Armstrong: One exception to deleting the copy and assignment
* Only delete the copy and assignment constructor if the base class does not have any pure virtual methods. If you have a pure virtual set these to default and the compiler handles it.

Levi-Armstrong: You can now copy derived classes for the profiles.

Levi-Armstrong: Reverted. The derived classes can be copied.

tesseract/tesseract_planning/tesseract_motion_planners/test/serialize_test.cpp
Levi-Armstrong: Clang format is doing this, but it is not as bad as before.

tesseract/tesseract_planning/tesseract_motion_planners/test/serialize_test.cpp
Levi-Armstrong: Clang format is doing this, but it is not as bad as before.

tesseract/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/trajopt/deserialize.h
Levi-Armstrong: reverted.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/process_managers/simple_process_manager.h
Levi-Armstrong: > Interesting. That does seem like a pretty big deal actually - though probably a taskflow issue. So if we call make_shared in the task anywhere that is essentially a memory leak?

It does not appear to be a memory leak for this case, but something related libc based on his reply. I have asked  for more details.

Levi-Armstrong: I have found that if I submit another request it does not appear to increase it memory but use what it has already taken ownership of. I do have concerns that this may starve other process but may not be the case per the authors comments related to modern OS handling this.

Levi-Armstrong:
@mpowelson All comments should be address now in the latest commit. I will squash the last commit into the last commit after approval and CI passes.

mpowelson (approved):

PR #495, opened 2020-12-29
Several tesseract collision fixes

Levi-Armstrong:


tesseract/tesseract_collision/include/tesseract_collision/fcl/fcl_utils.h
```diff
@@ -149,7 +149,9 @@ class CollisionObjectWrapper
     clone_cow->collision_objects_raw_.reserve(collision_objects_.size());
     for (const auto& co : collision_objects_)
     {
-      auto collObj = std::make_shared<FCLCollisionObjectWrapper>(*co);
+      assert(std::dynamic_pointer_cast<FCLCollisionObjectWrapper>(co) != nullptr);
+      auto collObj =
+          std::make_shared<FCLCollisionObjectWrapper>(*std::static_pointer_cast<FCLCollisionObjectWrapper>(co));
```
Levi-Armstrong, line 154: This was slicing because it was coping an intermediate type.

tesseract/tesseract_collision/src/bullet/tesseract_gjk_pair_detector.cpp
```diff
@@ -821,10 +821,10 @@ void TesseractGjkPairDetector::getClosestPointsNonVirtual(const ClosestPointInpu
     }
 
     m_simplexSolver->reset();
-    if (status == 0 && !m_cdata->req.calculate_distance)
+    if (status == 0 && !m_cdata->req.calculate_penetration)
     {
       // The shapes intersect but penetration was not request so return empty contact
-      output.addContactPoint(normalInB, pointOnB + positionOffset, distance);
+      output.addContactPoint(normalInB, pointOnB + positionOffset, -1);
```
Levi-Armstrong, line 827: the variable `distance` defaults to zero but if you set the contact threshold to zero then this was failing to get added as a contact because in the `processResults` function it checks if the distance is less than the contact threshold. This set the distance to negative one, though it may be better to set -1e-30.

tesseract/tesseract_collision/src/bullet/tesseract_gjk_pair_detector.cpp
```diff
@@ -821,10 +821,10 @@ void TesseractGjkPairDetector::getClosestPointsNonVirtual(const ClosestPointInpu
     }
 
     m_simplexSolver->reset();
-    if (status == 0 && !m_cdata->req.calculate_distance)
+    if (status == 0 && !m_cdata->req.calculate_penetration)
```
Levi-Armstrong, line 824: This was a bug. There is a following check that handles if distance is disabled.

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
```diff
@@ -119,7 +119,8 @@ inline ContactResult* processResult(ContactTestData& cdata,
   if (cdata.req.is_valid && !cdata.req.is_valid(contact))
     return nullptr;
 
-  if (!(contact.distance < cdata.collision_margin_data.getPairCollisionMarginData(key.first, key.second)))
+  if ((cdata.req.calculate_distance || cdata.req.calculate_penetration) &&
```
Levi-Armstrong, line 122: Only look up pair distance if distance or penetration data was requested.

mpowelson: This looks fine. I think what I was thinking before is that we need to add this before these lines (at 121) to keep from doing the lookup when the type is FIRST and we already found one.

```
  if (found && cdata.req.type == ContactTestType::FIRST)
    return nullptr;
```

tesseract/tesseract_collision/src/bullet/tesseract_gjk_pair_detector.cpp
mpowelson: I'll defer to your judgment here.

mpowelson (approved):

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
Levi-Armstrong: It should not enter this function a second time if the test type is `ContactTestType::FIRST`. It sets the `done` flag which prevents any further contact checks. I have an assert below which checks to make sure this does not happen `assert(cdata.req.type != ContactTestType::FIRST);`.

tesseract/tesseract_collision/src/bullet/tesseract_gjk_pair_detector.cpp
Levi-Armstrong: I will try and see if it works using `-std::numeric_limits<btScalar>::min()`.

Levi-Armstrong: This worked.

tesseract/tesseract_collision/include/tesseract_collision/core/common.h
mpowelson: This is fine then.

PR #496, opened 2020-12-31
Update planners to use CollisionCheckConfig

### LINKED ISSUE #427: Add generic collision config that all collision processes use.
These should all use the same config type for defining contact distance, special collision pairs, etc.

- [x] ContinuousContactCheckProcessGenerator
- [x] DiscreteContactCheckProcessGenerator
- [x] FixStateCollisionProcessGenerator
- [x] Motion planners

mpowelson:
I did not update trajopt because it has its own data structure. While we could go through and update it, I'd rather just focus on getting the ifopt planner working.

Similarly I added the ContactMarginData to OMPL but did not use the CollisionCheckConfig since OMPL handles LVS itself. I thought it would be confusing to have 2 longest valid segment terms where only one gets used.

Closes #427

### DIFF
--- a/tesseract/tesseract_common/cmake/tesseract_macros.cmake
+++ b/tesseract/tesseract_common/cmake/tesseract_macros.cmake
@@ -49,7 +49,41 @@ macro(tesseract_variables)
   set(TESSERACT_COMPILE_OPTIONS_PUBLIC "")
   set(TESSERACT_COMPILE_OPTIONS_PRIVATE "")
   if (NOT TESSERACT_ENABLE_TESTING AND NOT TESSERACT_ENABLE_TESTING_ALL)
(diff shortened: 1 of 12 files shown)


Levi-Armstrong (commented):
Overall looks good. With the addition of the CollisionCheckConfig what about for descartes remove the distiction between vertex and edge collision checking config because it contains the collision evaluator type? If discrete it would be applied as a vertex and if any of the others it would be applied either an edge or a combination.

The second questions, how do you think we can propagate a single CollisionCheckConfig  throughout for a single planning request? Thought I am not sure we would what this, but at least we would want to use the same parameters for the post collision check as what was used for the final motion planner.

mpowelson:
> Overall looks good. With the addition of the CollisionCheckConfig what about for descartes remove the distiction between vertex and edge collision checking config because it contains the collision evaluator type? If discrete it would be applied as a vertex and if any of the others it would be applied either an edge or a combination.

So would you have the option to use both?

> 
> The second questions, how do you think we can propagate a single CollisionCheckConfig throughout for a single planning request? Thought I am not sure we would what this, but at least we would want to use the same parameters for the post collision check as what was used for the final motion planner.

I was imagining that being a manual process where things like that are up to the user. Using the same config just makes it easier to fill out the profiles.

Levi-Armstrong:
> So would you have the option to use both?

It would behind the scene do the following based on the evaluator type. It is good have the vertex to filter out vertexs early prune down edges which take the longest during descartes planning. Though the current approach allows you have different settings for the vertices versus the path which may be nice feature for some cases.

Evaluator Types:
- DISCRETE: Add a vertex collision evaluator
- LVS DISCRETE: Add a vertex collision evaluator and edge collision evaluator. 
- CONTINUOUS: Add a vertex collision evaluator and edge collision evaluator.

Levi-Armstrong:
Based on the last comment it may be nice to leave it the way it is providing the ability to have different settings for vertices versus the path. What do you think?

mpowelson:
> Based on the last comment it may be nice to leave it the way it is providing the ability to have different settings for vertices versus the path. What do you think?

That's what I would be inclined to do. I think more flexibility is probably better in this case.

mpowelson:
@Levi-Armstrong So are there things I should change here or is this good to go?

Levi-Armstrong:
I think it is good to go.

Levi-Armstrong (approved):

PR #503, opened 2021-01-01
Utilize parameter in TrajOpt planner

Levi-Armstrong:


Levi-Armstrong:
@johnwason I am not sure why the python tests are failing for motion planning and the ones for c++ are passing due to the changes. Do you have any ideas?

Levi-Armstrong:
Oh, @johnwason How can I run the python tests locally?

johnwason:
You can run:

    python -m pytest

in the python test directory. You'll need to source the workspace and set TESSERACT_SUPPORT_DIR env variable before running. For Ubuntu Focal, run

    python3 -m pytest

johnwason:
The freespace planner is using OMPL, so it may just be having random failures during that phase. Maybe try using a different workflow for that test.

Levi-Armstrong:
Thanks, I also found an issues in the trajopt problem generator handling of fixed_states.

Issue #507, opened 2021-01-05
Does tesseract_python require yaml-cpp and jsoncpp?


Levi-Armstrong:
@johnwason Does tesseract_python required yaml-cpp and jsoncpp? It is listed in the tesseract_python_module.cmake but not listed in the package.xml so not sure if they should be added or if they should be removed from tesseract_python_module.cmake.

johnwason:
jsoncpp is required. I think yaml-cpp is there to fix some build issues I have run in to. These are dependencies of the packages already included in package.xml so I don't think they need to be explicitly included. They should be implicitly included from the other dependencies.

Levi-Armstrong:
Addressed by #508

PR #508, opened 2021-01-05
Remove jsoncpp and yaml-cpp from the tesseract_python_module.cmake

Levi-Armstrong:


Levi-Armstrong:
@johnwason Do you mind reviewing?

Levi-Armstrong:
This is to address issue #507

johnwason:
tesseract_python modules do use json-cpp. https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract_python/tesseract_python/swig/json_typemaps.i

Levi-Armstrong:
> tesseract_python modules do use json-cpp. https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract_python/tesseract_python/swig/json_typemaps.i

Ah, I added it back.

Levi-Armstrong:
@johnwason This is ready now. I was getting build issues in another CI because some included header files which included symbols that were not being linked.

Levi-Armstrong:
@johnwason Any objections with merging this?

PR #509, opened 2021-01-06
Add unit tests for improved mesh parser and fix normals loading bug

johnwason:
This pull request contains unit tests in C++ and Python to verify the improved mesh parsers. It also includes a fix for a normal loading bug found when running the tests.


Levi-Armstrong:
I assume this needs to be rebased on the latest?

johnwason:
I did a merge from latest so I think if you do a squash merge it should be fine. I will basically do the same thing locally anyway.

johnwason:
Screenshot of the test mesh for reference:

image

The test mesh contains material colors, PBR properties, and an image texture. It is loaded from collada and glb file formats. The collada loader does not support the PBR properties, so the default is used for all materials. It turns out that Ubuntu Bionic assimp does not support glb file formats, so the test is disabled when using the older assimp version.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
```diff
@@ -90,7 +90,7 @@ std::vector<std::shared_ptr<T>> extractMeshData(const aiScene* scene,
   {
     auto vertices = std::make_shared<tesseract_common::VectorVector3d>();
     auto triangles = std::make_shared<Eigen::VectorXi>();
-    std::shared_ptr<tesseract_common::VectorVector3d> normals = nullptr;
+    std::shared_ptr<tesseract_common::VectorVector3d> normals_ = nullptr;
```
Levi-Armstrong, line 93: Maybe rename to `vertex_normals` to align with ROS Style guide?

Levi-Armstrong (changes requested):
One minor change.

tesseract/tesseract_geometry/include/tesseract_geometry/mesh_parser.h
johnwason: I've made the change locally but am waiting for the CI to finish before pushing.

johnwason:
That small change was enough to throw off clang-format. Do you want me to do another pull request to fix it?

Levi-Armstrong:
Next PR should clean it up. Thank you. I assume in tesseract_ros we should add this information to the messages?

johnwason:
Yeah, I have these improvements in the Robot Raconteur messages, but I made some mistakes so I will need to update them. I will let you know when they are corrected.

Issue #511, opened 2021-01-06
Address Naming in Tesseract Process Managers


mpowelson:
The naming of some classes don't make a lot of sense as they are fragments of legacy code.

"We should probably open another issue to figure out what everything should be called. ProcessGenerators should probably be TaskGenerators. Then we would have TaskGenerators, TaskflowGenerators, and the ProcessPlanningServer + accessories."

_Originally posted by @mpowelson in https://github.com/ros-industrial-consortium/tesseract/issues/510#issuecomment-755022459_

Levi-Armstrong:
Current Idea:

ProcessGenerator -> TaskGenerator
ProcessInterface -> TaskflowInterface

Are there any other to be renamed?

Levi-Armstrong:
Would we also rename?

ProcessInfo -> TaskInfo
ProcessInput -> TaskflowInput

mpowelson:
I'd call ProcessInput TaskInput (rather than taskflow input) since we divide it up into sub ProcessInputs inside the taskflow generator before sending it to each Task.

I think ProcessInteface, ProcessPlanningFuture, ProcessPlanningRequest, and ProcessPlanningServer all all fine since they refer to a robotic process.

So we have things that refer to

- Taskflow Tasks
- Taskflow Taskflows
- Robotic processes (ie applications)

Levi-Armstrong:
Sounds good

mpowelson:
@Levi-Armstrong Are you addressing this or would you like me to?

Levi-Armstrong:
If you have time that would be great, if not I may be able to get later this week. I am currently working on releasing the dependencies.

mpowelson:
Summary
- ProcessGenerator -> TaskGenerator
- ProcessInterface -> TaskflowInterface
- ProcessInfo -> TaskInfo
- ProcessInfoContainer -> TaskInfoContainer
- ProcessInput -> TaskInput

Levi-Armstrong:
This has been addressed in #514

PR #512, opened 2021-01-06
Merge feature/CommandLanguage branch into master

Levi-Armstrong:


mpowelson:
If we are going to address #511 we probably should before we merge this. Updating user code is already going to be painful.

Levi-Armstrong:
@mpowelson You good with merging this into master? A tag has been created for the current state of master.

mpowelson:
> @mpowelson You good with merging this into master? A tag has been created for the current state of master.

In concept yes. I'm curious why you had to rebase it. Shouldn't we merge it in from the feature branch on this repo? I would think we would merge master into feature/CommandLanguage, add any fixes we need to, merge feature/CommandLanguage into master.

mpowelson:
That said, this is already done and passing CI. So on second thought, go for it.

PR #514, opened 2021-01-06
Move ProcessInfo into ProcessInterface for outside access

mpowelson:
Currently ProcessInfos are not accessibly outside of the PlanningServer. This moves them into the ProcessInterface such that they can be accessed by the user.


mpowelson:
Clang format failed because of the issue in mesh_parser. I can add that in a another commit if you want or leave it since I know there will be merge conflict fixes that need to be done too.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/core/task_generator.h
```diff
@@ -23,8 +23,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef TESSERACT_PROCESS_MANAGERS_PROCESS_GENERATOR_H
-#define TESSERACT_PROCESS_MANAGERS_PROCESS_GENERATOR_H
+#ifndef TESSERACT_PROCESS_MANAGERS_task_generator_H
```
Levi-Armstrong, line 26: I assume this happened during copy replace. Should be able to do a replace _task_generator_H with _TASK_GENERATOR_H.

Levi-Armstrong:
> Clang format failed because of the issue in mesh_parser. I can add that in a another commit if you want or leave it since I know there will be merge conflict fixes that need to be done too.

What ever is easiest. I can do it in the big PR for merging into master if that works best.

tesseract/tesseract_planning/tesseract_process_managers/include/tesseract_process_managers/core/task_generator.h
Levi-Armstrong: I will be pushing a commit to address the remaining items.

PR #515, opened 2021-01-06
Update processResults in tesseract_collision distance check

Levi-Armstrong:
The current approach does consider contact if the threshold and the distance are equal. So if the contact threshold is set to zero and the distance is zero this is not considered in contact. This switches the logic so if the threshold and distance are equal then it is considered in contact which is he behavior in the master branch.


DavidMerzJr:
As per our discussion, Benchmarking is currently broken.  If the others pass, this looks good to me.

PR #528, opened 2021-01-16
Add Findtcmalloc.cmake file

Levi-Armstrong:


mpowelson:
Looks like we need to add gperftools to the package.xml

https://github.com/ros/rosdistro/blob/f5a1e860b602a2e91356386f9b492460dc0c482a/rosdep/base.yaml#L1239

Levi-Armstrong:
At first I had that, but I left it out because this is more of just a cmake utility which is not used by this package and left it up to the package which actually calls find_package on tcmalloc to include it in there package.xml. I mainly put Findtcmalloc.cmake in tesseract_common because it was the lowest level and then the higher packages could then use this if needed. It may make sense to put this in cmake_common_scripts at some point.

PR #529, opened 2021-01-18
Add marker support and remove dependency on command language

Levi-Armstrong:
I did not have much time over the weekend to work on this but this is a start to breaking the tesseract_visualization dependency on tesseract_command_language. @marip8 and @mpowelson Do you have time to take a look and provided feedback?


tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
```diff
@@ -0,0 +1,25 @@
...
+{
+/** @brief An arrow defined by two points */
+class ToolpathMarker : public Marker
+{
+public:
+  ToolpathMarker() { type_ = static_cast<int>(MarkerType::TOOLPATH); }
+
+  bool show_path{ true };
+  bool show_axis{ true };
+  std::vector<std::vector<Eigen::Isometry3d>> toolpath;
```
marip8, line 21: Need special Eigen allocator for STL vectors

tesseract_visualization/include/tesseract_visualization/visualization.h
```diff
@@ -83,86 +79,46 @@ class Visualization
 
   /**
    * @brief Plot environment
-   * @param env The environment. If provided a nullptr it should the current environent.
+   * @param env The environment.
    */
-  virtual void plotEnvironment(tesseract_environment::Environment::ConstPtr env = nullptr) = 0;
+  virtual void plotEnvironment(tesseract_environment::Environment::ConstPtr env, std::string ns = "") = 0;
```
marip8, line 84: Is the namespace argument to these functions something that will change from call to call, or is it unlikely to change once a visualization class is created? If it's the latter, then I would suggest allowing an implementation to set this up in it's own `init` function

tesseract_visualization/include/tesseract_visualization/visualization.h
```diff
@@ -64,10 +60,10 @@ class Visualization
 
   /**
    * @brief Initialize the visualization tool
-   * @param env The environment object
+   * @detail It is recommend to check if connected before plotting any information
    * @return True if successful, otherwise false
    */
-  virtual bool init(tesseract_environment::Environment::ConstPtr env) = 0;
+  virtual bool init() = 0;
```
marip8, line 66: Normally, I would say that having two-step initialization is not a great idea. In this case, I assume that visualizations might be plugins in the future, in which case the `init` function is useful for setting up the class. I would suggest providing some sort of configuration object to the `init` function (e.g. YAML node, XMLRPC value, XML document, etc.) so it's internal parameters can be set dynamically

marip8 (commented):
I think this looks okay overall. Some thoughts:
- I'm on the fence about whether making a `Marker` base class buys you much over having separate functions to display each marker type. It's convenient for reducing the size of the visualization interface API, but you pay for it by having to do a lot of checking and type casting in the implementations 
- In the interest of keeping only the "core" code in the main repository, I would suggest that the ignition implementation eventually be migrated to the `tesseract_ignition` repository

tesseract_visualization/include/tesseract_visualization/visualization.h
Levi-Armstrong: I believe it could change per call. We currently do this on the ROS side for the marker array message, and thought it would be nice to have this capability here. I could image plotting each output of the trajectory pipeline in different namespace for trouble shooting and debugging.

tesseract_visualization/include/tesseract_visualization/visualization.h
Levi-Armstrong: Yep, it is currently intended to support the plugin functionality.  Shouldn't the init be universal to all implementation for the purpose of plugins? The config object(YAML node, XMLRPC value, XML document, etc.) seems specific to the implementation which could be handled by the implementation class, would we want this in the base class?

Levi-Armstrong:
> I'm on the fence about whether making a `Marker` base class buys you much over having separate functions to display each marker type.
> 
>     * Pro: convenient for reducing the size of the visualization interface API
> 
>     * Pro: allows for visualization of custom markers in the future
> 
>     * Con: requires a lot of checking and type casting in the implementations

I agree it does not gain much over what is currently provided, but my main motivation is expansion without having to modify the base class. This will hopefully allow the base class to be more stable where we can add additional markers without having to change the interface. Also as you pointed out it allows other to implement there own markers and leverage the base the class interface which will I could see being beneficial for outside users. I see this most useful on the trajopt side when implementing custom cost and constraints.

Levi-Armstrong:
>     * In the interest of keeping only the "core" code in the main repository, I would suggest that the ignition implementation eventually be migrated to the `tesseract_ignition` repository

I agree, I am waiting on some changes in snapcraft to support moving the plugin to tesseract_ignition.

tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
Levi-Armstrong: The Isometry3d is not a [fixed size vectorizable type](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html). Have you ran into issues?

marip8: Are you sure? I'm not sure that list is comprehensive, and it includes both the `Affine3d` and `Matrix4d` which are the same size matrix as `Isometry3d`

tesseract_visualization/include/tesseract_visualization/visualization.h
marip8: > Shouldn't the init be universal to all implementation for the purpose of plugins? The config object(YAML node, XMLRPC value, XML document, etc.) seems specific to the implementation which could be handled by the implementation class, would we want this in the base class?

Yes, the configuration object type should be generic and the same for all implementations, but its contents can change at runtime depending on how you need to configure the visualization. 

Normally you might provide implementation-specific parameters to these visualization implementations in their constructor. If they are plugins, though, they must be default constructable, so this isn't possible. It's valuable to have an alternative means of passing in parameters to visualization implementations in a generic way. This is where the `init` function with generic configuration object comes in. The configuration object should be a generic data type (like YAML node, XMLRPC value, `void*`, `std::any`, etc.) that is capable of dynamically representing different data, and the implementation should know how to interpret that generic data type to set up it's internal members.

In ROS-land, I typically use XMLRPC values because I can load them directly from the ROS parameter server and give them to a plugin. I would give the edge to YAML, XMLRPC, and XML over `void*` and `std::any` types because they are more straightforward to create from what we consider normal configuration files (e.g. text file, YAML file, XML file)

tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
Levi-Armstrong: Your right, I believe the list is not up to date. Looking at the code is is still a 4x4 matrix. Also I looked at tesseract_common types and I am actually aligning it so I will use that type. Good catch.

tesseract_visualization/include/tesseract_visualization/visualization.h
Levi-Armstrong: We decided to go wit std::any.

Levi-Armstrong:
@marip8 and @mpowelson This remove the tesseract_command_langauge dependency from tesseract_visualization. I am almost finished with updating tesseract_planning, tesseract_ros and trajopt_ros, but this PR should be ready to review.

tesseract_visualization/include/tesseract_visualization/visualization.h
marip8: It would be beneficial if the `Marker` base didn't have two-step initialization; all setup should really be done in the constructor. I think the model below would be a better way of handling plugins such that they (the plugins) can be configured after instantiation, but the marker classes can be fully initialized using only their constructor

```c++
// Marker plugin base class (i.e. type returned by plugin loader)
class MarkerPlugin : public Marker
{
public:
  // Calls the internal marker base class virtual function
  virtual void display() override final
  {
    return marker_->display();
  }
  
  // Provides a means for creating a specific type of Marker class
  virtual void init(std::any params) = 0;

protected:
  // The actual marker that can create a display
  std::unique_ptr<Marker> marker_;
};

// Plugin implementation (only responsible for creating marker type from parameters)
class AxisMarkerPlugin : public MarkerPlugin
{
public:
  virtual void init(std::any params) override
  {
    AxisMarkerConfig axis_marker_config = std::any_cast<AxisMarkerConfig>(any);
    marker = std::make_unique<AxisMarkerType>(config_object);
  }
};
```
I understand we're not implementing plugins in this PR, but I don't think we should introduce the `init` function into any `Marker` classes in this PR

Levi-Armstrong: @marip8 I removed the init method.

Levi-Armstrong: I held off with creating the plugin class until we get around to discussing how to handle loading plugins with Tesseract.

marip8 (approved):

PR #531, opened 2021-01-22
Move tesseract_command_langauge package to tesseract_planning repo

Levi-Armstrong:


Levi-Armstrong:
@mpowelson and @marip8 You good with merging this?

mpowelson:
Fine with me

marip8:
My suggestion [here](https://github.com/ros-industrial-consortium/tesseract_planning/discussions/7) was to only migrate the planning-specific command language features into the planning repository. Would it be beneficial to leave some of the core types (`Instruction`, `CompositeInstruction`, etc.) in the main repository for other `tesseract` to build on? Or is the ultimate plan for the core command language code to be in its own repository?

Levi-Armstrong:
> My suggestion [here](https://github.com/ros-industrial-consortium/tesseract_planning/discussions/7) was to migrate only the planning-specific command language features into the planning repository. Would it be beneficial to leave some of the core types (`Instruction`, `CompositeInstruction`, etc.) in the main repository for other `tesseract` to build on? Or is the ultimate plan for the core command language code to be in its own repository? Or is its only use case for planning?

I think it would be good to move it to the tesseact_planning until things are more stable. I believe somethings will be changing with the core items. Then once things are more stable we can break things out into its own repository or move into tesseract core repository. What do you think?

PR #535, opened 2021-01-27
Fix harmonizeTowardZero

Levi-Armstrong:


tesseract_kinematics/include/tesseract_kinematics/core/utils.h
```diff
@@ -360,10 +360,10 @@ inline void harmonizeTowardZero(FloatType* qs, int dof)
...
   for (int i = 0; i < dof; i++)
   {
-    if (qs[i] > pi)
-      qs[i] -= two_pi;
-    else if (qs[i] < -pi)
-      qs[i] += two_pi;
+    FloatType dif = std::fmod(qs[i] + pi, two_pi);
+    if (dif < 0)
+      dif += two_pi;
+    qs[i] = dif - pi;
```
marip8, line 366: Was the issue with the previous implementation that it didn't account for values that were greater than 2*pi?

marip8: Also, I assume we don't need to fix this equivalent function in `opw_kinematics` because I think that it only returns solutions on [-2*pi, 2*pi], correct?

Levi-Armstrong: > Was the issue with the previous implementation that it didn't account for values that were greater than 2*pi?

That was the issue.

> Also, I assume we don't need to fix this equivalent function in opw_kinematics because I think that it only returns solutions on [-2pi, 2pi], correct?

Yea that would be a good.

mpowelson:
Probably closes https://github.com/swri-robotics/descartes_light/issues/8?

marip8:
> Edit: Looks like this function is there too. We'll probably want to copy the changes

I'm not sure we need that function any more since we recently removed the kinematics interface. It doesn't look like it's being used. I'll make an issue on the repository and we can track it more offline

Levi-Armstrong:
@mpowelson and @marip8 are there any changes needed for this PR?

mpowelson (approved):

tesseract_kinematics/include/tesseract_kinematics/core/utils.h
marip8: Does the addition of `pi` in L363 and subtraction of `pi` in L366 enable you to take a value `pi <= qs[i] < 2 * pi` and force it into the [-pi, pi] range? I didn't exactly follow that

Levi-Armstrong: Me neither. I came across different examples and this was the first one I tested which worked for the test cases.

marip8: We probably should add some documentation to this operation to explain it. If it's not intuitive now, I can only image what we'll think when we look at this again after a few weeks.

PR #536, opened 2021-01-29
Add almost equal relative and abs vectorized version

Levi-Armstrong:
This also includes an update to harmonize unit test in kinematics.

### DIFF
--- a/tesseract_common/include/tesseract_common/utils.h
+++ b/tesseract_common/include/tesseract_common/utils.h
@@ -343,6 +343,7 @@ inline tinyxml2::XMLError QueryDoubleAttributeRequired(const tinyxml2::XMLElemen
 
 /**
  * @brief Check if two double are relatively equal
+ * @details The max_diff is for handling when comparing numbers near zero
  * @param a Double
  * @param b Double
  * @param max_diff The max diff when comparing std::abs(a - b) <= max_diff, if true they are considered equal
@@ -355,8 +356,6 @@ inline bool almostEqualRelativeAndAbs(double a,
                                       double max_diff,
                                       double max_rel_diff = std::numeric_limits<double>::epsilon())
 {
-  // Check if the numbers are really close -- needed
-  // when comparing numbers near zero.
   double diff = std::fabs(a - b);
   if (diff <= max_diff)
     return true;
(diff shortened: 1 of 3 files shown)


tesseract_kinematics/test/kinematics_core_unit.cpp
```diff
@@ -73,6 +73,13 @@ TEST(TesseractKinematicsUnit, UtilsHarmonizeUnit)  // NOLINT
   tesseract_kinematics::harmonizeTowardZero(q.data(), 2);
   EXPECT_NEAR(q[0], M_PI_4, 1e-6);
   EXPECT_NEAR(q[1], -M_PI_4, 1e-6);
+
+  q[0] = M_PI_4;
+  q[1] = -M_PI_4;
+
+  tesseract_kinematics::harmonizeTowardZero(q.data(), 2);
+  EXPECT_NEAR(q[0], M_PI_4, 1e-6);
+  EXPECT_NEAR(q[1], -M_PI_4, 1e-6);
```
marip8, line 82: It's probably worth adding a test case for a number in the range [pi, 2 * pi] (like 5 * pi / 4) to verify correct behavior

tesseract_common/include/tesseract_common/utils.h
```diff
@@ -368,5 +368,30 @@ inline bool almostEqualRelativeAndAbs(double a,
...
+ * largest is the largest of the absolute values of a and b.
+ * @return True if they are relativily equal, otherwise false.
+ */
+inline bool almostEqualRelativeAndAbs(const Eigen::Ref<const Eigen::VectorXd>& a,
+                                      const Eigen::Ref<const Eigen::VectorXd>& b,
+                                      double max_diff,
+                                      double max_rel_diff = std::numeric_limits<double>::epsilon())
+{
+  // Check if the numbers are really close -- needed
+  // when comparing numbers near zero.
```
marip8, line 387: Move this comment to the `@details` block

tesseract_common/include/tesseract_common/utils.h
```diff
@@ -368,5 +368,30 @@ inline bool almostEqualRelativeAndAbs(double a,
   return (diff <= largest * max_rel_diff);
 }
 
+/**
+ * @brief Check if two Eigen VectorXd are relatively equal
+ * @details This is a vectorized implementation of the function above.
+ * @param a A vector of double
+ * @param b A vector of double
+ * @param max_diff The max diff when comparing std::abs(a - b) <= max_diff, if true they are considered equal
```
marip8, line 376:
```suggestion
 * @param max_diff The max diff when comparing max(abs(a - b)) <= max_diff, if true they are considered equal
```

tesseract_common/include/tesseract_common/utils.h
```diff
@@ -368,5 +368,30 @@ inline bool almostEqualRelativeAndAbs(double a,
   return (diff <= largest * max_rel_diff);
 }
 
+/**
+ * @brief Check if two Eigen VectorXd are relatively equal
```
marip8, line 372: This function is checking both absolute and relative equality, not just relative equality, right?

tesseract_common/include/tesseract_common/utils.h
```diff
@@ -368,5 +368,30 @@ inline bool almostEqualRelativeAndAbs(double a,
...
+                                      double max_rel_diff = std::numeric_limits<double>::epsilon())
+{
+  // Check if the numbers are really close -- needed
+  // when comparing numbers near zero.
+  Eigen::VectorXd diff_abs = (a - b).cwiseAbs();
+  double diff = diff_abs.maxCoeff();
+  if (diff <= max_diff)
+    return true;
+
+  return (diff_abs.array() <= (max_rel_diff * a.cwiseAbs().cwiseMax(b.cwiseAbs())).array()).all();
```
marip8, line 393: These element-wise operations would probably be a little cleaner if you worked with `Eigen::ArrayXd` instead of `Eigen::VectorXd`

tesseract_common/include/tesseract_common/utils.h
Levi-Armstrong: That is correct. I will update the description.

tesseract_common/include/tesseract_common/utils.h
Levi-Armstrong: I will see what I can come up with.

Levi-Armstrong:
@marip8 Take a look at the latest two commits which include the changes you requested.

PR #537, opened 2021-02-01
Improve tesseract_scene_graph test coverage and fix found issues

Levi-Armstrong:


tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
```diff
@@ -222,6 +222,69 @@ TEST(TesseractSceneGraphSRDFUnit, LoadSRDFUnit)  // NOLINT
...
+  tinyxml2::XMLElement* element = xml_doc.FirstChildElement("robot");
+  EXPECT_TRUE(element != nullptr);
+
+  AllowedCollisionMatrix acm = parseDisabledCollisions(*g, element, std::array<int, 3>({ 1, 0, 0 }));
+  EXPECT_TRUE(acm.isCollisionAllowed("base_link", "link_1"));
+  EXPECT_TRUE(acm.isCollisionAllowed("base_link", "link_2"));
+  EXPECT_TRUE(acm.isCollisionAllowed("base_link", "link_3"));
+
+  // Now test failures
+  auto failure_test = [g](const std::string& xml_string) {
```
mpowelson, line 248: For these, maybe rename to is_failure, or add some documentation that make it clear that returning true is a failure.

tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
```diff
@@ -222,6 +222,69 @@ TEST(TesseractSceneGraphSRDFUnit, LoadSRDFUnit)  // NOLINT
...
+  };
+
+  {  // missing link1
+    std::string xml_string =
+        R"(<robot name="abb_irb2400">
+             <disable_collisions link2="link_1" reason="Adjacent" />
+             <disable_collisions link1="base_link" link2="link_2" reason="Never" />
+             <disable_collisions link1="base_link" link2="link_3" reason="Never" />
+           </robot>)";
+    EXPECT_TRUE(failure_test(xml_string));
```
mpowelson, line 266: Perhaps consider adding a fully valid case with an EXPECT_FALSE both to test failure_test and also act as an example.

mpowelson (approved):

tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
Levi-Armstrong: It may not be visible in the diff but right above the failure code should be the full working example; is this what you are looking for?

tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
Levi-Armstrong: renamed.

Issue #540, opened 2021-02-03
addSceneGraph does not add links to collision


haudren:
I was recently in the process of upgrading to the command language version of Tesseract and hit a problem with it: merging another scene graph into my environment resulted in links being correctly added to the environment, but not to the collision checking, and my robot would happily go through walls...

My investigation shows that:
- `addSceneGraph` mostly calls `scene_graph_->insertSceneGraph`, which calls its own `addLink`
- `Environment::addLink` correctly inserts both a link and the collisions, but obviously `SceneGraph::addLink` doesn't, as it does not have access to the managers.
- Because `getDiscreteContactManager` returns a clone of the manager, you can't add externally the collisions

Workaround is thus to call `Environment::registerDefaultContactManagers` to re-generate the list of collision objects. But this feels like an awkward API:
- I call `Environment::addSceneGraph`, I expect it to register itself properly in all nooks and crannies
- As in #539 , it feels quite counter-intuitive that a function starting with `get` that returns a `shared_ptr` actually returns a *clone* of the manager. I would recommend returning a const reference in the getter, and then using an actual `clone` method or `getCopy` when we need a copy.

How do you feel about fixing the above points? I can draft a PR, but I'm not sure which way you want to go

Levi-Armstrong:
@haudren Thank you for the bug report. I have created PR #541 which should address this issue. Do you mind testing it?

Levi-Armstrong:
@haudren  I have created a new PR #544 which includes the fix along with cleaning up duplicated code and adds several unit tests with one that confirms this bug is fixed. Let me know if this does not address the issue.

haudren:
Issue is fixed, thanks :+1:

PR #541, opened 2021-02-03
Fix issues with environment addSceneGraph not updating contact managers

Levi-Armstrong:
This includes the following

- Add a simple timer class to support motion planning and examples
- Fixes bug in Environment `addSceneGraph` which was not updating the contact manager with newly added links.

### DIFF
--- a/tesseract_common/include/tesseract_common/timer.h
+++ b/tesseract_common/include/tesseract_common/timer.h
@@ -1,0 +1,88 @@
+/**
+ * @file timer.h
+ * @brief Simple timer class using chrono
+ *
+ * @author Levi Armstrong
+ * @date February 2, 2021
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2021, Southwest Research Institute
+ *
+ * @par License
+ * Software License Agreement (Apache License)
+ * @par
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_COMMON_TIMER_H
+#define TESSERACT_COMMON_TIMER_H
+
+#include <tesseract_common/macros.h>
+TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
+#include <chrono>
+TESSERACT_COMMON_IGNORE_WARNINGS_POP
+
+namespace tesseract_common
+{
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
This is to address issue #540

haudren:
Thanks for the fix, but it doesn't seem to address my issue. My test goes as follows:
```cpp
     tesseract_environment::Environment env;
     /* init env with a robot, load a URDF that contains a single box */
    ASSERT_TRUE(env.addSceneGraph(box)));

    auto link = env.getLink("box");

    ASSERT_NE(link, nullptr);

    auto manager = env.getDiscreteContactManager();

    EXPECT_TRUE(manager->hasCollisionObject(link->getName()));
``` 
This fails on the last assertion, meaning that it seems like the problem is still there. I debugged quickly, seems like the set_difference is not working correctly: I have 11 members in pre_links, 13 in post_links yet the difference is empty (size 0).

Just a couple questions though:
- Can we maybe add a unit test to cover this? I have one in my branch, I'd be happy to contribute if that's welcome
- There seems to be an awful lot of duplication between the two overloads of `addSceneGraph`. Is there any reason why we don't simply call one from the other as follows:

```cpp
bool Environment::addSceneGraph(const tesseract_scene_graph::SceneGraph& scene_graph, const std::string& prefix)
{
  // Connect root of subgraph to graph
  tesseract_scene_graph::Joint root_joint(prefix + scene_graph.getName() + "_joint");
  root_joint.type = tesseract_scene_graph::JointType::FIXED;
  root_joint.parent_link_name = getRootLinkName();
  root_joint.child_link_name = prefix + scene_graph.getRoot();
  root_joint.parent_to_joint_origin_transform = Eigen::Isometry3d::Identity();

  return addSceneGraph(scene_graph, std::move(root_joint), prefix);
}
```
I almost feel that we are adding twice the same subgraph when calling the joint-less version :thinking: As far as I remember, that's how I designed the feature when I contributed it but I didn't follow the newer developments closely. I guess it has to do with having an empty scene graph, but shouidn't we straight up return false in that case?

haudren:
As I suspected, using this naive loop instead of `std::set_difference` makes the test pass:
```cpp
    for (const auto post_link : post_links)
    {
      if (std::find_if(pre_links.begin(), pre_links.end(), [post_link](const auto& pre_link) {
            return post_link->getName() == pre_link->getName();
          }) == pre_links.end())
      {
        diff_links.push_back(post_link);
      }
    }
```

Levi-Armstrong:
> Can we maybe add a unit test to cover this? I have one in my branch, I'd be happy to contribute if that's welcome

Absolutely. I checked and there is a test `runAddSceneGraphTest` which would be good to expand and added additional if needed. I am working on getting the unit test coverage for everything in this repository over 90%, but I am currently working on tesseract_kinematics then finish with tesseract_environment so any help would greatly appreciated. 

>There seems to be an awful lot of duplication between the two overloads of addSceneGraph. Is there any reason why we don't simply call one from the other as follows:

I agree, I was going to wait until the unit tests were written before going through and cleaning things up, but if you write test which cover these two function and want to clean it up that is ok with me.

Levi-Armstrong:
replaced by PR #544 

@haudren I have started cleaning up the environment in the above PR. I will apply these changes there and begin adding unit tests.

PR #544, opened 2021-02-08
Refactor tesseract_environment to use applyCommands

Levi-Armstrong:
- This refactors the environment to reduce code duplication.
- Use tesseract namespace alias for filesystem namespace to allow switching to std::filesystem when ready.
- Switch addLink, addJoint, moveLink and insertSceneGraph to take `const&` to remove the need for `std::move` in scene graph and environment
- Switch environment to leverage shared mutex
- Rename AddCommand to AddLinkCommand


Levi-Armstrong:
@johnwason These changes should hopefully simplify the python added the shared pointer versions.

tesseract_environment/include/tesseract_environment/core/command.h
```diff
@@ -0,0 +1,81 @@
...
+  using ConstPtr = std::shared_ptr<const Command>;
+
+  Command() = default;
+  virtual ~Command() = default;
+  Command(const Command&) = default;
+  Command& operator=(const Command&) = default;
+  Command(Command&&) = default;
+  Command& operator=(Command&&) = default;
+
+  virtual CommandType getType() const = 0;
```
mpowelson, line 75: Is it intended that users could add commands to this list and implement their own Environment? If so, we might want to make this return an int like we do in tcl for WaypointType, etc.

tesseract_environment/src/core/environment.cpp
```diff
@@ -161,227 +162,159 @@ bool Environment::isInitialized() const { return initialized_; }
 
 int Environment::getRevision() const { return revision_; }
 
-const Commands& Environment::getCommandHistory() const { return commands_; }
+Commands Environment::getCommandHistory() const
+{
+  std::lock_guard<std::mutex> lock(mutex_);
```
mpowelson, line 167: Should we make this a shared mutex?

tesseract_environment/src/core/environment.cpp
```diff
@@ -161,227 +162,159 @@ bool Environment::isInitialized() const { return initialized_; }
...
       }
-      else
+      case tesseract_environment::CommandType::REMOVE_LINK:
       {
-        // This should only occur if the this graph is empty.
-        if (!addSceneGraph(*(cmd.getSceneGraph()), cmd.getPrefix()))
-          return false;
+        auto cmd = std::dynamic_pointer_cast<const RemoveLinkCommand>(command);
+        success &= applyRemoveLinkCommand(cmd);
+        break;
```
mpowelson, line 208: Would it be unsafe to make these static_pointer_casts? The only way it could not be that type is if we have a bug in the code.

tesseract_environment/src/core/environment.cpp
```diff
@@ -1316,4 +881,492 @@ Environment::Ptr Environment::clone() const
...
+
+bool Environment::moveLink(const tesseract_scene_graph::Joint::ConstPtr& joint)
+{
+  return applyCommand(std::make_shared<MoveLinkCommand>(joint));
+}
+
+bool Environment::addLink(tesseract_scene_graph::Link link)
+{
+  return applyCommand(std::make_shared<AddCommand>(std::make_shared<tesseract_scene_graph::Link>(std::move(link))));
+}
```
mpowelson, line 913: This naming disconnect seems odd. If we are going to break it, now seems like the time.

tesseract_environment/test/tesseract_environment_unit.cpp
```diff
@@ -475,20 +475,21 @@ void runApplyCommandsTest()
...
       EXPECT_TRUE(env->applyCommands(commands));
       std::vector<std::string> link_names = env->getLinkNames();
       std::vector<std::string> joint_names = env->getJointNames();
       EXPECT_TRUE(std::find(link_names.begin(), link_names.end(), link_name2) != link_names.end());
       EXPECT_TRUE(std::find(joint_names.begin(), joint_names.end(), "joint_" + link_name2) != joint_names.end());
     }
     {
-      Commands commands{ std::make_shared<AddCommand>(nullptr, joint_1) };
-      EXPECT_FALSE(env->applyCommands(commands));
+      EXPECT_DEATH(Commands commands{ std::make_shared<AddCommand>(nullptr, joint_1) }, "");
```
mpowelson, line 486: Should we fix this? This seems less than ideal.

mpowelson (commented):
Looks pretty good to me.

tesseract_environment/include/tesseract_environment/core/command.h
Levi-Armstrong: I have not given it much thought but I think for now it would not be intended for the user to have the ability to add custom commands. The change was mainly to make this an abstract class.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: I think that would be good. I will do this in a separate commit or PR after the clean up.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: I think this would be ok.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: I agree. I will make this change in a separate command after the clean up.

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: I will fix it such that an exception is thrown in the constructor of the AddCommand class.

tesseract_environment/include/tesseract_environment/core/environment.h
```diff
@@ -226,7 +228,8 @@ class Environment
 
     for (std::size_t i = 1; i < commands.size(); ++i)
     {
-      if (!applyCommand(*(commands.at(i))))
+      /** @todo Levi fix */
```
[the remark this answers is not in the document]
Levi-Armstrong: I think this should be updated to send a vector to avoid unnecessary calls to environment changed.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: done.

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: I switched command to take const& which solved solves the issue.

tesseract_environment/include/tesseract_environment/core/environment.h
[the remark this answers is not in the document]
Levi-Armstrong: This has been cleanup in this [commit](https://github.com/ros-industrial-consortium/tesseract/pull/544/commits/bacb71cd2a47cb081230d3949104d91ee5a45267). I will fixup when review is finished.

Levi-Armstrong:
@johnwason Here is the PR where I am making the change to hopefully simplify the python as well.

tesseract_environment/include/tesseract_environment/ofkt/ofkt_state_solver.h
```diff
@@ -53,7 +53,9 @@ namespace tesseract_environment
 class OFKTStateSolver : public StateSolver
 {
 public:
+  // LCOV_EXCL_START
```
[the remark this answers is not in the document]
Levi-Armstrong: Yea, I am not sure if we will need to add these very many places. I do have this documented in the readme for reference if anyone needs to do this in the future.

johnwason:
There are a lot of changes here. I am going to need some time to go through and update the Python side to make sure everything still works.

Levi-Armstrong:
The interface changes should be minimal, most of the changes implementation cleanup and additional unit tests. Also broke out the environment commands into separate files.

johnwason:
I don't see the SWIG `%shared_ptr` declarations for the commands. They will need to be added to the files for the Python wrappers to work properly. I'll have to add the new files to the Python wrappers, since SWIG doesn't follow includes automatically.

Levi-Armstrong:
> I don't see the SWIG `%shared_ptr` declarations for the commands. They will need to be added to the files for the Python wrappers to work properly. I'll have to add the new files to the Python wrappers, since SWIG doesn't follow includes automatically.

They are all in the commands.h file still. Will this work?

Levi-Armstrong:
Here is the [file](https://github.com/Levi-Armstrong/tesseract-1/blob/issue/RefactorEnvironment2/tesseract_environment/include/tesseract_environment/core/commands.h)

johnwason:
Ok I see the %include, %shared_ptr, and %shared_factory there. That should be fine. SWIG will pick up the extra header files from there.

johnwason:
This PR looks fine. I had to make one minor change to the SWIG wrappers: https://github.com/ros-industrial-consortium/tesseract_python/pull/7/

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: This has been added in a single commit.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: A new commit has been added making this change.

tesseract_environment/include/tesseract_environment/core/commands/add_command.h
```diff
@@ -0,0 +1,80 @@
...
+   *
+   *        This method should attach the link to the root link with a fixed joint
+   *        with a joint name of joint_{link name}".
+   *
+   * @param link The link to be added to the graph
+   */
+  AddCommand(const tesseract_scene_graph::Link::ConstPtr& link)
+    : link_(std::make_shared<tesseract_scene_graph::Link>(link->clone())), joint_(nullptr)
+  {
+  }
```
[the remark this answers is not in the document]
Levi-Armstrong: A new commit has been added making this change.

mpowelson (approved):
Looks like good changes to me.

tesseract_environment/src/ofkt/ofkt_state_solver.cpp
```diff
@@ -291,9 +304,11 @@ void OFKTStateSolver::onEnvironmentChanged(const Commands& commands)
 
     switch (command->getType())
     {
-      case tesseract_environment::CommandType::ADD:
+      case tesseract_environment::CommandType::ADD_LINK:
       {
-        const auto& cmd = static_cast<const tesseract_environment::AddCommand&>(*command);
+        const auto& cmd = static_cast<const tesseract_environment::AddLinkCommand&>(*command);
+        assert(cmd.getJoint() != nullptr);
+
```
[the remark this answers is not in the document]
Levi-Armstrong: What about adding a ReplaceLinkCommand?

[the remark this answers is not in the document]
mpowelson: We could add another command. I figured it would probably be easier to just add a flag to the existing one that defaults to the current behavior.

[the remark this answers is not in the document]
Levi-Armstrong: Actually, I just went through the code and I think your approach would be best. Adding the replace command would also have the issue with what do you do if the link does not exist. I will add a flag that enables the ability to replace link if it exist but it will use the existing joint if a joint was not provided. Sound good?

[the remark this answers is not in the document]
mpowelson: I think that sounds good. Is this what you were thinking?
- Link provided, link already exists -> replace link
- Link provided, link doesn't exist -> add link
- Joint provided, joint already exists -> replace joint
- Joint provided, joint doesn't exist -> add joint
- Joint not provided, joint already exists -> use existing joint
- Joint not provided, joint doesn't exist -> error?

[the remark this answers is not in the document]
Levi-Armstrong: This is what I have implemented so far. I currently writing unit tests to make sure things are working as expects. With these changes the `ADD_LINK` does not accurately represent the command not that it can be used to replace only a joint should it be change?

| Link (Exists) | Joint  (Exists) | Description |
| ---| --- | --- |
| Provided (Yes) | Not Provided | replace link use existing joint |
| Provided (No) | Not Provided | add link with joint attaching to root link |
| Not Provided | Provided (Yes) | replace joint which may move if the parents are not the same |
| Not Provided | Provided (No) | error |
| Provided (Yes) | Provided (Yes) | replace link and joint (These do not have to be associated in any way) |
| Provided (No) | Provided (No) | add link and joint (joint child must be the provided link) |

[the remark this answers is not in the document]
mpowelson: We could move the logic for the 3rd and 4th row into the move joint command. That might fit the description a bit better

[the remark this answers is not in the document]
Levi-Armstrong: Well during the process it was getting messy so I broke out the replace joint into its own command

AddLinkCommand:

| Link (Exists) | Joint  (Exists) | Description |
| ---| --- | --- |
| Provided (Yes) | Not Provided | replace link use existing joint |
| Provided (No) | Not Provided | add link with joint attaching to root link |
| Provided (No) | Provided (No) | add link and joint (if child link is not the name of the provided link it errors) |
| Provided (Yes) | Provided (No) | error |
| Provided (No) | Provided (Yes) | error |

ReplaceJointCommand:

| Joint  (Exists) | Description |
| ---| --- |
| Provided (Yes) | replace joint use existing joint (if child link is not the same it results in an error) |
| Provided (No) | error |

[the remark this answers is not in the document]
Levi-Armstrong: @mpowelson I just created a [PR](https://github.com/ros-industrial-consortium/tesseract/pull/547) add support for replacing links and joints. It was a little more work than I expected but I believe it appears to be working.

PR #547, opened 2021-02-12
Add support for replacing links and joints

Levi-Armstrong:


tesseract_environment/include/tesseract_environment/ofkt/ofkt_state_solver.h
```diff
@@ -164,6 +166,29 @@ class OFKTStateSolver : public StateSolver
...
+   * @param removed_joints The removed joint names container
+   * @param removed_joints_indices The removed joint names indices container
+   * @param joint The joint performing the move
+   */
+  void moveLinkHelper(std::vector<tesseract_scene_graph::Joint::ConstPtr>& new_kinematic_joints,
+                      std::vector<std::string>& removed_joints,
+                      std::vector<long>& removed_joints_indices,
+                      const tesseract_scene_graph::Joint::ConstPtr& joint);
+
+  void replaceJointHelper(std::vector<tesseract_scene_graph::Joint::ConstPtr>& new_kinematic_joints,
```
mpowelson, line 181: Doxygen

tesseract_environment/src/core/environment.cpp
```diff
@@ -1061,12 +1067,51 @@ bool Environment::addSceneGraph(const tesseract_scene_graph::SceneGraph& scene_g
...
+  }
 
-  std::string link_name = cmd->getLink()->getName();
-  if (!cmd->getJoint())
+  if ((link_exists && cmd->getJoint()))
   {
+    CONSOLE_BRIDGE_logWarn("Tried to add link (%s) which already exists with a joint provided. This is not supported.",
+                           link_name.c_str());
+    return false;
+  }
```
mpowelson, line 1100: Why isn't replaceing the link and joint at the same time supported? Can't we just call the same replaceJoint function in the scene graph?

tesseract_environment/src/core/environment.cpp
```diff
@@ -1158,6 +1216,34 @@ bool Environment::applyRemoveJointCommand(const RemoveJointCommand::ConstPtr& cm
...
+    CONSOLE_BRIDGE_logWarn("Tried to replace Joint (%s) where the child links are not the same",
+                           cmd->getJoint()->getName().c_str());
+    return false;
+  }
+
+  if (!scene_graph_->removeJoint(cmd->getJoint()->getName()))
+    return false;
+
+  if (!scene_graph_->addJoint(*cmd->getJoint()))
+    return false;
```
mpowelson, line 1239: If this fails, should we re-add the original joint (if that is possible)? At the least we should print an error letting the user know that the original joint was removed successfully since that won't be stored in the commands

tesseract_environment/src/ofkt/ofkt_state_solver.cpp
```diff
@@ -556,6 +548,111 @@ void OFKTStateSolver::update(EnvState& state,
...
+{
+  auto& n = nodes_[joint->getName()];
+
+  if (n->getType() == joint->type)
+  {
+    n->getParent()->removeChild(n.get());
+    n->setStaticTransformation(joint->parent_to_joint_origin_transform);
+    OFKTNode* new_parent = link_map_[joint->parent_link_name];
+    n->setParent(new_parent);
+    new_parent->addChild(n.get());
```
mpowelson, line 587: So the state solver allows the child to be different?

tesseract_environment/test/tesseract_environment_unit.cpp
```diff
@@ -151,6 +221,29 @@ Environment::Ptr getEnvironment(bool register_default_contact_managers = false)
   return env;
 }
 
+template <typename S>
+void runEnvInitFailuresTest()
+{
+  auto env = std::make_shared<Environment>();
+  EXPECT_TRUE(env != nullptr);
+  EXPECT_EQ(0, env->getRevision());
+  EXPECT_FALSE(env->reset());
```
mpowelson, line 230: I didn't realize an uninitialized env returned false if you called reset. That makes sense, but we might want to update the doxygen.

tesseract_environment/test/tesseract_environment_unit.cpp
```diff
@@ -151,6 +221,29 @@ Environment::Ptr getEnvironment(bool register_default_contact_managers = false)
...
+  Commands commands;
+  EXPECT_FALSE(env->init<S>(commands));
+  EXPECT_FALSE(env->isInitialized());
+
+  // Test scene graph is not first command
+  commands.clear();
+  auto cmd = std::make_shared<MoveJointCommand>("joint_name", "parent_link");
+  commands.push_back(cmd);
+  EXPECT_FALSE(env->init<S>(commands));
+  EXPECT_FALSE(env->isInitialized());
```
mpowelson, line 244: Perhaps check that CommandHistory is still size 0?

tesseract_scene_graph/include/tesseract_scene_graph/graph.h
```diff
@@ -181,9 +181,10 @@ class SceneGraph
    * The first link added to the graph is set as the root by default. Use setRoot to change the root link of the graph.
    *
    * @param link The link to be added to the graph
-   * @return Return False if a link with the same name allready exists, otherwise true
+   * @param replace_allowed If true and the link exist it will be replaced
+   * @return Return False if a link with the same name allready exists and replace is not allowed, otherwise true
```
mpowelson, line 185: Typo `already`

tesseract_environment/src/core/environment.cpp
```diff
@@ -1075,11 +1120,24 @@ bool Environment::applyAddCommand(AddLinkCommand::ConstPtr cmd)
...
-  if (!scene_graph_->addLink(*cmd->getLink(), *cmd->getJoint()))
-    return false;
+  // If Link existed remove it from collision before adding the replacing links geometry
+  if (link_exists)
+  {
+    if (discrete_manager_ != nullptr)
+      discrete_manager_->removeCollisionObject(link_name);
+    if (continuous_manager_ != nullptr)
+      continuous_manager_->removeCollisionObject(link_name);
+  }
```
mpowelson, line 1140: This is extremely important, and for some reason I didn't think about this being required. Is there a generic way to unit test this for all relevant commands? Maybe apply the command to an environment, initialize a seperate environment with its scene graph, and then compare the two environment's collision scenes?

mpowelson:
Thanks for doing this. It looks great! I think the biggest thing is that I want to make sure I understand how it is intended to work. Let me try a different chart format that makes a bit more sense to me. This is what I was thinking, but I think it is different from what you have. What were you thinking?

***AddLink***

| Exists  | Provided Link Only | Provided Joint Only | Provided Link + Joint |
| ------------- | ------------- | ------------- | ------------- |
| Link Only  | Not Possible?  | Error  | Replace Link + Add Joint  |
| Joint Only  | Not Possible? | Error | Add Link + Replace Joint  |
| Link + Joint  | Replace Link  | Error  | Replace Link + Replace Joint  |

***ReplaceJoint***

|  | Joint Provided | Joint Not Provided | 
| ------------- | ------------- |  ------------- | 
|  Joint Exists  | Replace Joint | Error |
| Joint Doesn't Exist | Error (or add, but I don't think that is possible) | Error  |

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: I was not sure on this one. I guess it should probably return true?

tesseract_environment/src/core/environment.cpp
mpowelson: See chart below

tesseract_environment/test/tesseract_environment_unit.cpp
mpowelson: The doxygen says it resets to the initialized state, so returning false if it isn't initialized makes sense to me. I was just saying we should specify in documentation.

tesseract_environment/include/tesseract_environment/ofkt/ofkt_state_solver.h
Levi-Armstrong: done.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: Good catch. Fixed.

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: Documented.

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: Done.

tesseract_scene_graph/include/tesseract_scene_graph/graph.h
Levi-Armstrong: Fixed.

tesseract_environment/src/core/environment.cpp
Levi-Armstrong: I agree. I have a few unit tests checking that the contact manager is getting updated.

Levi-Armstrong:
@mpowelson This should be ready for review. You can not replace just links and joints along with link and joint pair. I also wrote several more unit test for the tesseract_environment, tesseract_common, and tesseract_scene_graph. Now this repository and all packages are above 90% test coverage.

tesseract_scene_graph/include/tesseract_scene_graph/kinematics_information.h
```diff
@@ -55,6 +78,32 @@ struct ROPKinematicParameters
   std::string positioner_group;
   std::string positioner_fk_solver;
   std::unordered_map<std::string, double> positioner_sample_resolution;
+
+  bool operator==(const ROPKinematicParameters& rhs) const
+  {
+    bool success = true;
+    success &= (manipulator_group == rhs.manipulator_group);
+    success &= (manipulator_ik_solver == rhs.manipulator_ik_solver);
+    success &= (tesseract_common::almostEqualRelativeAndAbs(manipulator_reach, manipulator_reach, 1e-6));
```
mpowelson, line 87: Missing `rhs.`

tesseract_scene_graph/include/tesseract_scene_graph/kinematics_information.h
```diff
@@ -65,6 +114,32 @@ struct REPKinematicParameters
   std::string positioner_group;
   std::string positioner_fk_solver;
   std::unordered_map<std::string, double> positioner_sample_resolution;
+
+  bool operator==(const REPKinematicParameters& rhs) const
+  {
+    bool success = true;
+    success &= (manipulator_group == rhs.manipulator_group);
+    success &= (manipulator_ik_solver == rhs.manipulator_ik_solver);
+    success &= (tesseract_common::almostEqualRelativeAndAbs(manipulator_reach, manipulator_reach, 1e-6));
```
mpowelson, line 123: Same here. Missing rhs

tesseract_environment/test/tesseract_environment_unit.cpp
```diff
@@ -68,8 +75,132 @@ tesseract_scene_graph::SRDFModel::Ptr getSRDFModel(const SceneGraph::Ptr& scene_
...
+
+  for (int i = 0; i < 20; ++i)
+  {
+    EnvState::Ptr random_state = base_solver.getRandomState();
+    compare_solver.setState(random_state->joints);
+    runCompareEnvStates(base_joint_names, *random_state, *compare_solver.getCurrentState());
+  }
+}
+
+enum class EnvRegistarMethod
```
mpowelson, line 195: Is this intended to be register?

tesseract_common/include/tesseract_common/clone_cache.h
```diff
@@ -165,17 +194,33 @@ class CloneCache
...
+    try
+    {
+      clone = original_->clone();
+    }
+    catch (std::exception& e)
+    {
+      CONSOLE_BRIDGE_logError("Clone Cache failed to update cache with the following exception: %s", e.what());
+      return nullptr;
+    }
+    return clone;
```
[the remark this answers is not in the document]
Levi-Armstrong: I do not think so and if it did I would believe there is a bug somewhere that needs to be tracked down. I noticed that the original implementation was doing this in one place but not all when I wrote a unit test to test the throw an exception in clone. I assume this could be possible depending on what class is passed to it, but I believe this would be an unlikely case because if the original was good then the clone should not have an issues unless there is a bug I assume. What do you think we should do here? We could remove it  and let the class that uses this decide to catch if this is occurring, because I believe this would mean there is an issue somewhere in the class being passed.

tesseract_scene_graph/include/tesseract_scene_graph/kinematics_information.h
Levi-Armstrong: fixed.

tesseract_scene_graph/include/tesseract_scene_graph/kinematics_information.h
Levi-Armstrong: fixed.

tesseract_environment/test/tesseract_environment_unit.cpp
Levi-Armstrong: Yep. Fixed.

tesseract_environment/src/ofkt/ofkt_state_solver.cpp
Levi-Armstrong: This is being checked prior to reaching this function but it may be good to add asserts here because it should never happen if the upstream checking is correct.

mpowelson (approved):

PR #556, opened 2021-03-01
Add universal robot inverse kinematics

Levi-Armstrong:
This add universal robot inverse kinematics implementation taken from the ur_kinematics which is a ROS packages. Next step is to add support to the SRDF to support defining this for a manipulator.


gavanderhoorn:
What's the motivation for duplication here @Levi-Armstrong ?

> universal robot inverse kinematics implementation taken from the ur_kinematics 

You are aware of the many problems with that package?

Levi-Armstrong:
> You are aware of the many problems with that package?

I am not, is there something wrong with the inverse kinematics code?

gavanderhoorn:
I'm still not sure whether copying it all makes sense, but this:

> The other reason is this package does not depend ROS so did not what to depend on ROS package given it was a small amount of code to port.

seems like the only proper reason -- and even that is probably fixable (but who's going to do that).

> > You are aware of the many problems with that package?
> 
> I am not, is there something wrong with the inverse kinematics code?

there are quite some open PRs that attempt to fix it all.

See [ros-industrial/universal_robot/projects/1](https://github.com/ros-industrial/universal_robot/projects/1) for an overview.

Levi-Armstrong:
> and even that is probably fixable (but who's going to do that)

What did you have in mind?

Levi-Armstrong:
> I guess that's part of the intent of this PR?

I assume you referring to something along the lines of plug-in support. From the API side it is setup to support this just but I have not got around to setup how the user defines the plug-in in a config file or srdf. I believe the plan is to added support to the srdf so everything related to kinematics is in the srdf.

Levi-Armstrong:
@mpowelson You have any objections with me merging this?

mpowelson:
No. It's not much code if we end up removing it later.

PR #557, opened 2021-03-02
Fix method for changing bullet extern margin gDbvtMargin

Levi-Armstrong:


mpowelson (approved):
So this was inflating objects by 5cm in the broadphase?

Levi-Armstrong:
> So this was inflating objects by 5cm in the broadphase?

In the beginning it was being controlled by compiler definition and they changed it to an extern which was not correctly being set so I believe everything had the 5cm inflation which was impacting collision checking performance.

PR #561, opened 2021-03-04
Add scale to tool path marker

Levi-Armstrong:
See title.

### DIFF
--- a/tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
+++ b/tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
@@ -28,6 +28,7 @@ public:
   bool show_path{ true };
   bool show_axis{ true };
   tesseract_common::Toolpath toolpath;
+  Eigen::Vector3d scale{ Eigen::Vector3d::Constant(0.03) };
 };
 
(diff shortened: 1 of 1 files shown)


tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
```diff
@@ -28,6 +28,7 @@ class ToolpathMarker : public Marker
   bool show_path{ true };
   bool show_axis{ true };
   tesseract_common::Toolpath toolpath;
+  Eigen::Vector3d scale{ Eigen::Vector3d::Constant(0.03) };
```
mpowelson, line 31: Was 0.03 already being hardcoded somewhere?

mpowelson (approved):

tesseract_visualization/include/tesseract_visualization/markers/toolpath_marker.h
Levi-Armstrong: Kinda, that is usually what I set it to in rviz when using pose array. With a scale of 1 the marker is cylinders are 1m long. The defaults them to 3cm for the tool path.

PR #569, opened 2021-03-11
Add kinematics utility function isNearSingularity

Levi-Armstrong:


tesseract_kinematics/test/kinematics_core_unit.cpp
```diff
@@ -89,6 +91,25 @@ TEST(TesseractKinematicsUnit, UtilsHarmonizeUnit)  // NOLINT
...
+  EXPECT_TRUE(tesseract_kinematics::isNearSingularity(jacobian, 0.001));
+  jv[4] = 1 * M_PI / 180.0;
+  jacobian = fwd_kin.calcJacobian(jv);
+  EXPECT_TRUE(tesseract_kinematics::isNearSingularity(jacobian));
+  jv[4] = 2 * M_PI / 180.0;
+  jacobian = fwd_kin.calcJacobian(jv);
+  EXPECT_FALSE(tesseract_kinematics::isNearSingularity(jacobian));
+  EXPECT_TRUE(tesseract_kinematics::isNearSingularity(jacobian, 0.02));
+}
+
```
mpowelson, line 112: I'll trust you that that is near a singularity for that robot

mpowelson (approved):
Looks fine to me, though I'm not familiar with the intricacies of JacobiSVD

tesseract_kinematics/include/tesseract_kinematics/core/utils.h
```diff
@@ -218,6 +218,20 @@ inline static bool dampedPInv(const Eigen::Ref<const Eigen::MatrixXd>& A,
...
+ * @details This is keep separated from the forward kinematics because special consideration may need to be made
+ * based on the kinematics arrangement.
+ * @param jacobian The jacobian to check if near a singularity
+ * @param threshold The threshold that all singular values must be above to not be considered near a singularity
+ */
+inline bool isNearSingularity(const Eigen::Ref<const Eigen::MatrixXd>& jacobian, double threshold = 0.01)
+{
+  Eigen::JacobiSVD<Eigen::MatrixXd> svd(jacobian, Eigen::ComputeThinU | Eigen::ComputeThinV);
+  const Eigen::VectorXd& sv = svd.singularValues();
+  return !(sv.tail(1).value() > threshold);
```
marip8, line 232:
```suggestion
  return sv.tail(1).value() < threshold;
```

Inverting the logic here might make this a little more understandable and match your documentation description

tesseract_kinematics/test/kinematics_core_unit.cpp
```diff
@@ -89,6 +91,25 @@ TEST(TesseractKinematicsUnit, UtilsHarmonizeUnit)  // NOLINT
...
+  Eigen::VectorXd jv = Eigen::VectorXd::Zero(6);
+  Eigen::MatrixXd jacobian = fwd_kin.calcJacobian(jv);
+  EXPECT_TRUE(tesseract_kinematics::isNearSingularity(jacobian, 0.001));
+  jv[4] = 1 * M_PI / 180.0;
+  jacobian = fwd_kin.calcJacobian(jv);
+  EXPECT_TRUE(tesseract_kinematics::isNearSingularity(jacobian));
+  jv[4] = 2 * M_PI / 180.0;
+  jacobian = fwd_kin.calcJacobian(jv);
+  EXPECT_FALSE(tesseract_kinematics::isNearSingularity(jacobian));
```
marip8, line 109: What are you trying to show by rotating joint 5 here? I'm a little confused about what you're testing here

tesseract_kinematics/include/tesseract_kinematics/core/utils.h
Levi-Armstrong: The descriptions states `The singular values must be greater than the threshold` and the suggested change would mean `The singular values must be greater than or equal to the threshold`. I am good with the change just want to make sure we are on the same page and I will update the documentation.

tesseract_kinematics/test/kinematics_core_unit.cpp
Levi-Armstrong: For the robot being used, singularity occurs when joint 4,5,6 are all at zeros. Changing joint 5 value should move it out of singularity which is demonstrated by the unit test. As joint 5 increase so does the smallest singular value.

tesseract_kinematics/test/kinematics_core_unit.cpp
Levi-Armstrong: Yea for the abb robot when joint 4, 5 and 6 are all zeros it is at a singularity and by changing joint 5 causes it to move away from the singularity.

tesseract_kinematics/include/tesseract_kinematics/core/utils.h
Levi-Armstrong: @marip8 I made the change and updated the documentation to reflect the behavior.

tesseract_kinematics/test/kinematics_core_unit.cpp
marip8: That's what I don't understand. Axes 4 and 6 would still be aligned (i.e. still in singularity) when J5=0, J5=180, and J5=360. This unit test shows that there is no singularity when J5=360, which should be the same condition as J5=0.

Levi-Armstrong: Sorry, I don't see where J5 is being set to 360 deg. It currently is only setting it to 0, 1 deg and 2 deg.

marip8: My bad; I was reading this wrong. This makes sense now

Issue #521, opened 2021-01-11
ROS Buildfarm Release "Noetic"


Levi-Armstrong:
- [ ] cmake_common_scripts
- [ ] opw_kinematics
- [ ] descartes_light
- [ ] osqp
- [ ] osqp_eigen
- [ ] qpOASES
- [ ] tesseract
- [ ] tesseract_planning
- [ ] tesseract_ros
- [ ] tesseract_python

Levi-Armstrong:
@mpowelson, @marip8 and @johnwason I believe this repository is ready to be released. Any objections to me releasing this repository for Noetic with tag 1.0.0?

johnwason:
I am going to try a conda build tonight, let's make sure it will work there.

johnwason:
No, it isn't building using conda on windows. I can't look into this until the week after next most likely. We can add patches later to the conda package if needed, or wait until the conda build is working.

marip8:
I'm on board for the release. As far as I can tell it looks like the major discussions about de-coupling the planning/visualization code have been addressed and there aren't any outstanding PRs with major changes.

What is your tagging strategy? Are the 0.x.x tags supposed to represent pre-release states and 1+.x.x represent post-release states? Just wondering if the next tag should be 0.3.0 since there really hasn't been a "major" change since the last tag

gavanderhoorn:
> What is your tagging strategy? Are the 0.x.x tags supposed to represent pre-release states and 1+.x.x represent post-release states? Just wondering if the next tag should be 0.3.0 since there really hasn't been a "major" change since the last tag

Going to `1.x` immediately may indeed not be what you'd want.

If you're going to use [semver](https://semver.org), seeing a `1.x` release implies quite a few things.

mpowelson:
I think I agree that a 0.x release might be a better option right now. Then if that goes well we can plan on a 1.0 release sometime soon.

Levi-Armstrong:
@gavanderhoorn, @mpowelson and @marip8 Thank you all for the input. I will begin the release process and use tag 0.4.0.

Levi-Armstrong:
Just submitted the [release](https://github.com/ros/rosdistro/pull/29068) on the ROS buildfarm.

Levi-Armstrong:
@johnwason Everything has been upgraded to taskflow v3.0.0.

johnwason:
@Levi-Armstrong ok I'll give the conda build another try

[...]

PR #570, opened 2021-03-11
Fix the way in which Eigen is included

haudren:
Currently, Eigen is added via `target_include_directories` as a PUBLIC
or INTERFACE tag. Unfortunately, this makes a built tesseract
non-reusable on a system where Eigen is installed in a different folder.
For example, Tesseract built on a system where Eigen was downloaded in
the ROS workspace is not usable on a system where Eigen is installed in
the default system directory.

For a few more details:
- I have some binary builds coming from a Docker image, but those are not reusable on the host system :(
- I test by moving my install folder around on various machines

In both cases, CMake errors out with:
```
CMake Error in CMakeLists.txt:
  Imported target "tesseract::tesseract_environment_core" includes
  non-existent path

    "/home/herve/ros/ws/install/include/eigen3"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

```
As indeed, the user on machine B is not the same as on machine A, but Eigen is installed on both.

The root cause is that CMake generates the `tesseract_*-targets.cmake`, and they contain the dependency list. Without this fix, they look like:
```
set_target_properties(tesseract::tesseract_environment_core PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_COMPILE_OPTIONS "-mno-avx"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/home/herve/ros/ws/install/include/eigen3"
  INTERFACE_LINK_LIBRARIES "tesseract::tesseract_common;tesseract::tesseract_collision_core;tesseract::tesseract_collision_bullet;tesseract::tesseract_collision_fcl;tesseract::tesseract_scene_graph;tesseract::tesseract_urdf;tesseract::tesseract_kinematics_kdl;tesseract::tesseract_kinematics_opw"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/home/herve/ros/ws/install/include/eigen3"
)
```

Note the two absolute paths to the Eigen libraries. With this fix:
```
set_target_properties(tesseract::tesseract_environment_core PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_COMPILE_OPTIONS "-mno-avx"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Eigen3::Eigen;tesseract::tesseract_common;tesseract::tesseract_collision_core;tesseract::tesseract_collision_bullet;tesseract::tesseract_collision_fcl;tesseract::tesseract_scene_graph;tesseract::tesseract_urdf;tesseract::tesseract_kinematics_kdl;tesseract::tesseract_kinematics_opw"
)
```

Closing thoughts:
- This is useful for #521 
- The same should probably be done for the other libraries linked to in this fashion, but they are harder to wrangle out. Orocos is marked as having a bug, TinyXML2 doesn't provide a CMake integration file on Ubuntu18.04, Boost relies on components, and those [were fixed](https://github.com/Kitware/CMake/commit/37da6af17d0d2cc8b499dc58f74866351e06c288#diff-e003c7469605cc6b910e66c8baffa6fb1c3b565467723808b14f63d88fb751a5) in CMake 3.15 which is only available on 20.04.
- Finally, it's probably worth setting up a CMakeLists formatter and / or linter as they are getting quite large. We have internally set up [cmake-format](https://github.com/cheshirekow/cmake_format) and it's been quite helpful in harmonizing all of them.


gavanderhoorn:
I'm pretty sure @Levi-Armstrong did things like this because the target you're using now is not universally supported on all platforms which Tesseract wants to support.

Levi-Armstrong:
Thank you for the PR. I have been meaning to go through and do this for Eigen and Boost for a while since we dropped support for 16.04.

> The same should probably be done for the other libraries linked to in this fashion, but they are harder to wrangle out. Orocos is marked as having a bug, TinyXML2 doesn't provide a CMake integration file on Ubuntu18.04, Boost relies on components, and those were fixed in CMake 3.15 which is only available on 20.04.

This is the goal to get everything to using target over cmake variables. 

I believe the Orocos marked as a having a bug may not be valid any more. I believe the issue was related to catkin to correctly handling interface targets which I believe have been fixed so its worth testing again. 

What was the issues with Boost on 18.04?

> Finally, it's probably worth setting up a CMakeLists formatter and / or linter as they are getting quite large. We have internally set up cmake-format and it's been quite helpful in harmonizing all of them.

I agree, I will take a look.

haudren:
> What was the issues with Boost on 18.04?

Main issue is that different Tesseract libraries look for different Boost components. In CMake 3.10, doing:
```
find_package(Boost COMPONENTS thread)
find_pacakge(Boost COMPONENTS filesystem)
```
Correctly finds both `boost::thread` and `boost::filesystem`. However, if you do:
```
find_dependency(Boost COMPONENTS thread)
find_dependency(Boost COMPONENTS filesystem)
```
Only `boost::thread` is found, because `find_dependency` contains a shortcutting instruction in case the package (not the components...) has already been found. 

This means that:
- You can't `find_package(Boost COMPONENTS xxx)` before any package that itself needs Boost
- You can't depend on two packages that both need different Boost packages

This behavior fixed in 3.15, now `find_dependency` always calls `find_package`.

Levi-Armstrong:
@haudren Thank you for the information. I have updated both tesseract and tesseract_planning to leverage boost targets and update the cmake config to use find_package if the cmake version is less than 3.15 due to the issue that you outlined. 

Also the issues with kdl has be resolved so that has also been updated. 

Next I will work on TinyXml2 to leverage targets. Do you know if anything else that needs updating?

PR #573, opened 2021-03-16
Add support for defining collision margin data in SRDF

Levi-Armstrong:
This PR adds the ability to define collision margin data in the SRDF as shown below and gets loaded into the environment.

``` xml
<collision_margins default_margin="0.025">
  <pair_margin link1="link_6" link2="link_5" margin="0.01"/>
  <pair_margin link1="link_5" margin="0.015"/>
</collision_margins>
```

Additional change to accommodate this change:

- Move CollisionMarginData to tesseract_common
- Simplified SRDF removing setters and getters and exposed variables directly
- Add new environment command ChangeCollisionMarginsCommand and deprecated the other two command now that this one command encompasses both. They will be removed in about a week to give time for project to update.


tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
```diff
@@ -1974,6 +2000,125 @@ TEST(TesseractSceneGraphSRDFUnit, SRDFGroupTCPsUnit)  // NOLINT
...
+  {  // missing pair margin
+    std::string str = R"(<robot name="abb_irb2400">
+                           <collision_margins default_margin="0.025">
+                             <pair_margin link1="link_6" link2="link_5" margin="0.01"/>
+                             <pair_margin link1="link_5" link2="link_4"/>
+                           </collision_margins>
+                         </robot>)";
+    EXPECT_ANY_THROW(is_failure(str));
+  }
+}
```
mpowelson, line 2120: Possible additional tests

- Negative margin
- Fields exist but are empty

mpowelson (approved):
Looks reasonable to me. Do you want to add this to one of the SRDFs in tesseract_support just as an example/test?

tesseract_scene_graph/test/tesseract_scene_graph_srdf_unit.cpp
Levi-Armstrong: These are good additional tests. I will work on adding these.

Levi-Armstrong:
> Looks reasonable to me. Do you want to add this to one of the SRDFs in tesseract_support just as an example/test?

Sure, I will add it the sphere test with a larger margin between the wrist joints and the sphere.

PR #577, opened 2021-03-27
Remove dependency on class_loader and leverage Boost DLL

Levi-Armstrong:
This adds two classes to support loading plugins within Tesseract.

- Class ClassLoader: 
  - This is a wrapper around Boost DLL for simplifying loading symbols from libraries.
- Class PluginLoader
  - This leverages ClassLoader for loading plugins. This class supports adding search paths, multiple plugins, and Environment Variables to leverage for locating plugins. 
  - This is what will be used throughout for managing plugins.


Levi-Armstrong:
@johnwason This should solve the class_loader issue.

PR #578, opened 2021-03-30
Fixup enforceJointLimits

haudren:
While browsing recent changes, I noticed a bug in the new enforce_joint_limits function:
Up to now, it would incorrectly apply the upper limit to any position
that's outside the range. For example, a position that's slightly under
the lower limit would get assigned the upper limit. Fix this by using
Eigen's min and max functions, resulting in a proper clamp.

To test this:
```cpp
  Eigen::VectorXd test = Eigen::VectorXd::Constant(6, -2);
  Eigen::MatrixXd limits(6, 2);
  limits.col(0) = Eigen::VectorXd::Constant(6, -1);
  limits.col(1) = Eigen::VectorXd::Constant(6, 1);

  enforcePositionLimits(test, limits);

  std::cout << test.transpose() << std::endl;
```

Before:
```
1 1 1 1 1 1
```
After
```
-1 -1 -1 -1 -1 -1
```


Levi-Armstrong (approved):

Levi-Armstrong:
Thank you for the fix and improving the unit tests.

PR #579, opened 2021-03-30
Fix inv kinematics to only return solution within limits and add redundant solutions for kdl ik solvers

Levi-Armstrong:


tesseract_kinematics/src/opw/opw_inv_kin.cpp
```diff
@@ -66,7 +66,8 @@ IKSolutions OPWInvKin::calcInvKin(const Eigen::Isometry3d& pose,
       harmonizeTowardZero<double>(eigen_sol);  // Modifies 'sol' in place
 
       // Add solution
-      solution_set.push_back(eigen_sol);
+      if (tesseract_common::satisfiesPositionLimits(eigen_sol, limits_.joint_limits))
```
mpowelson, line 69: Off topic, but I notice at 58 we aren't reserving any room for redundant solutions.

tesseract_kinematics/src/kdl/kdl_inv_kin_chain_nr.cpp
```diff
@@ -93,7 +94,21 @@ IKSolutions KDLInvKinChainNR::calcInvKinHelper(const Eigen::Isometry3d& pose,
 
   KDLToEigen(kdl_solution, solution);
 
-  return { solution };
+  IKSolutions solution_set;
+  tesseract_kinematics::harmonizeTowardZero<double>(solution);
```
mpowelson, line 98: Does this mean that the first solution will be the one that is closest to 0 rather than the one that is closest to the seed?

Levi-Armstrong: That is correct. Should I change the logic so the original solution is first?

mpowelson: Yeah I would think so. I'm not sure if we state it in the documentation, but my perception was that the first solution should be the closest to the seed. If you seed IK with the current location and take the first solution, I wouldn't want it to do a joint flip.

tesseract_kinematics/src/opw/opw_inv_kin.cpp
Levi-Armstrong: I am not sure there is a good way to calculate the number of redundant solutions before hand given the position, though it should not be difficult to calculate the max number of redundant solutions based on the joint limits.

tesseract_kinematics/src/kdl/kdl_inv_kin_chain_nr.cpp
Levi-Armstrong: Yea, I don't think we want to guarantee that just because it can be computationally intensive for the other IK solvers like opw_kinematics and ikfast where we would need to run a sort operation. The seed is more to help out the numerical solvers though for these we can add it first. Though I would push this on the user to perform the sort if needed to avoid a sort operation.

Levi-Armstrong:
So in the process I noticed that the `getRedundantSolutions` did not have a unit test so I added one. In the process I found that it was not returning all of the possible permutation of the redundant solutions. Also in the process ran into more numerical issues so update the `satisfiesPositionLimits` to function the same way as `almostEqualRelativeAndAbs`. It should be good for another look.

mpowelson (approved):

PR #580, opened 2021-03-31
Update benchmarks after API change in discrete contact manager

deh0512:
The API for `tesseract_collision::DiscreteContactManager` changed, but benchmark code was not updated accordingly. This PR corrects that oversight and fixes a couple of warnings.

To test locally, simply build and run the benchmarks (e.g., with `--cmake-args -DTESSERACT_ENABLE_BENCHMARKING=Yes`). Presumably, this will allow the benchmarks to be run in CI as well.

On a side note, it would be nice to be able to build the benchmarks without running them immediately, but that's a separate issue.


Levi-Armstrong:
Thank you for the fix!

PR #583, opened 2021-04-02
Cleanup equal operator

Levi-Armstrong:


tesseract_common/src/manipulator_info.cpp
```diff
@@ -81,152 +83,30 @@ const Eigen::Isometry3d& ToolCenterPoint::getTransform() const
...
+}
+bool ToolCenterPoint::operator!=(const ToolCenterPoint& rhs) const { return !operator==(rhs); }
 
-      tcp.setExternal(external, external_frame);
-    }
-  }
+template <class Archive>
+void ToolCenterPoint::serialize(Archive& ar, const unsigned int /*version*/)  // NOLINT
+{
+  ar& boost::serialization::make_nvp("type", type_);
```
mpowelson, line 101: I notice about you used `BOOST_SERIALIZATION_NVP`. It looks like this probably doesn't matter, but is there any reason that we should be using the macro instead of this?

tesseract_common/src/serialization.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+template <class Archive>
+void serialize(Archive& ar, Eigen::VectorXd& g, const unsigned int version)
+{
+  split_free(ar, g, version);
+}
+
+/*****************************/
+/****** Eigen::VectorXd ******/
```
mpowelson, line 72: Comment wrong

tesseract_common/src/serialization.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+template <class Archive>
+void load(Archive& ar, Eigen::Isometry3d& g, const unsigned int /*version*/)
+{
+  g.setIdentity();
+  ar& boost::serialization::make_nvp("xyz", boost::serialization::make_array(g.translation().data(), 3));
+  Eigen::Quaterniond q;
+  ar& boost::serialization::make_nvp("xyzw", boost::serialization::make_array(q.vec().data(), 4));
+  q.normalize();
+  g.linear() = q.toRotationMatrix();
```
mpowelson, line 91: Is this correct?

mpowelson (approved):

tesseract_common/src/manipulator_info.cpp
marip8: I think the main difference is that you can specify the name in the NVP using the non-macro method. The macro uses the name of the variable as the name in the NVP. Not sure if the manually specified name is worth keeping or not

tesseract_common/src/serialization.cpp
marip8: This looks fine to me. `linear` is the function call for the `Transform` class to get the upper left 3x3 matrix. Is there something specific that looks incorrect?

tesseract_common/include/tesseract_common/serialization.h
```diff
@@ -48,82 +45,38 @@ namespace serialization
 /****** Eigen::VectorXd ******/
 /*****************************/
 template <class Archive>
-inline void save(Archive& ar, const Eigen::VectorXd& g, const unsigned int /*version*/)
-{
-  long rows = g.rows();
-  ar& BOOST_SERIALIZATION_NVP(rows);
-  ar& boost::serialization::make_nvp("data", boost::serialization::make_array(g.data(), rows));
-}
+void save(Archive& ar, const Eigen::VectorXd& g, const unsigned int version);  // NOLINT
```
marip8, line 48: It might be nice at some point (maybe not in the scope of this PR) to try to serialize the `Matrix` or `MatrixBase` class to be able to serialize vectors and matrices of any shape/number type using additional template parameters to specify these dimensions/types. I tried doing this quickly the other day, but wasn't all that successful. Something to think about for the future

tesseract_common/src/manipulator_info.cpp
Levi-Armstrong: Yea it was more for esthetics. I did not what the name of the variable in the xml to have underscore though it may be good to keep them the same for traceability. What do you all thing?

tesseract_common/include/tesseract_common/serialization.h
Levi-Armstrong: I agree, I had looked at this a while back and found examples but was unsure if this would cause issues with the custom Isometry3d. I will take another look in the near future.

tesseract_common/src/serialization.cpp
Levi-Armstrong: Fixed.

tesseract_common/include/tesseract_common/serialization.h
marip8: `Isometry3d` is a typedef of `Transform`, so I wouldn't think it would cause issues

Levi-Armstrong: That is good. I will see if I can find the code again and add it.

johnwason:
We are seeing a new build error after this PR:

```
error: ‘const ManipulatorInfo’ {aka ‘const struct tesseract_common::ManipulatorInfo’} has no member named ‘toXML’
  104 |     tinyxml2::XMLElement* xml_manip_info = getManipulatorInfo().toXML(doc);
```

Levi-Armstrong:
Is this coming from tesseract_planning? If so there is a open PR which should solve the issues. I should have it merged in tomorrow.

PR #601, opened 2021-05-05
IK Solver Redundant Solutions Update

marip8:
This PR:
- Removes the addition of redundant solutions to the output of the inverse kinematic solvers
  - Returning redundant IK solutions with nominal solutions can cause downstream applications to repeat operations because those applications cannot disambiguate the two types of solutions. For example, the Tesseract Descartes planner collision checks all generated IK solutions, resulting in a lot of unnecessary repeated work when the vast majority of IK solutions are just redundancies of a solution that has already been checked.
  - Applications requiring redundant solutions should use the provided utility function to produce them from a nominal solution
- Adds a function for generating a list of indices which indicate which joints in a input list are capable of generating redundant solutions. 
  - Linear joints, unlike revolute joints, are not capable of producing redundant solutions. 
- Updates the redundant solution calculation function to only generate redundant solutions for joints that are capable of producing them

### DIFF
--- a/tesseract_kinematics/include/tesseract_kinematics/core/utils.h
+++ b/tesseract_kinematics/include/tesseract_kinematics/core/utils.h
@@ -394,6 +394,28 @@ ForwardKinematicsConstPtrMap createKinematicsMap(const tesseract_scene_graph::Sc
 }
 
 /**
+ * @brief Creates a vector indicating which joints in the input list of joint names are capable of producing redundant solutions
+ */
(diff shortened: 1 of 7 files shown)


Levi-Armstrong (approved):

johnwason:
@Levi-Armstrong Can we get a release tag to catch up with these changes?

Levi-Armstrong:
I have created a new tag and submitted a release to the build farm.

PR #604, opened 2021-05-17
fix error so that initial state has dt=0

cbw36:


Levi-Armstrong:
@cbw36 If you run clang formatting this should be good to merge.

PR #609, opened 2021-06-03
Add convex decomposition support

Levi-Armstrong:
This adds a convex decomposition interface along with a VHACD and HACD implementation. The VHACD code has been added to extras because it does not appear to be available for install which may be removed in the future. It has been namespace under tesseract_collision to avoid the possible linking issue if linked against another library using another version of VHACD.

Files to review are located in these two directories:

https://github.com/Levi-Armstrong/tesseract-1/tree/feature/ConvexDecompositon/tesseract_collision/include/tesseract_collision/convex_decomposition

https://github.com/Levi-Armstrong/tesseract-1/tree/feature/ConvexDecompositon/tesseract_collision/src/convex_decomposition


Levi-Armstrong:
We should be able to get rid of the VHACD code once it is released as part of Bullet. I have created a discussion on Bullet to track the progress.

PR #616, opened 2021-07-06
Add redundancy methods to Forward and Inverse Kinematics interface

Levi-Armstrong:
Utility functions were previously added to support calculating redundant solutions. This updated the Forward and Inverse Kinematics interface to simplify using the function making them part of the interface.


tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
```diff
@@ -131,6 +131,12 @@ class InverseKinematics
    */
   virtual void setLimits(tesseract_common::KinematicLimits limits) = 0;
 
+  /**
+   * @brief Get vector indicating which joints are capable of producing redundant solutions
+   * @return A vector of joint indicies
+   */
+  virtual const std::vector<Eigen::Index>& getRedundancyCapableJointIndices() const = 0;
```
marip8, line 138: Does it make sense for a pure virtual function to return a const reference? It's one thing for a class to return a reference of a member that will always exist, but it's another thing to require the implementation of this interface to keep the `std::vector<Eigen::Index>&` in scope for as long as it will be used (which the implementation can't know). I would suggest just returning the object instead of a referenece. It will probably be small and this method probably won't be called frequently, so we won't be paying a noticeable performance penalty by doing this
```suggestion
  virtual const std::vector<Eigen::Index> getRedundancyCapableJointIndices() const = 0;
```

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
```diff
@@ -151,6 +157,13 @@ class InverseKinematics
 
   /** @brief Clone the forward kinematics object */
   virtual std::shared_ptr<InverseKinematics> clone() const = 0;
+
+  /**
+   * @brief Given a solution calculate its redundant solutions
+   * @param sol The solution to calculate redundant solutions about
+   * @return A vector of redundant solutions
+   */
+  IKSolutions getRedundantSolutions(const Eigen::Ref<const Eigen::VectorXd>& sol) const;
```
marip8, line 166: I assume this function just returns the redundant solutions, not including the original solution. If this is the case, it would good to make note of this behavior

tesseract_kinematics/src/core/inverse_kinematics.cpp
```diff
@@ -0,0 +1,36 @@
...
+ */
+#include <tesseract_kinematics/core/utils.h>
+#include <tesseract_kinematics/core/inverse_kinematics.h>
+
+namespace tesseract_kinematics
+{
+IKSolutions InverseKinematics::getRedundantSolutions(const Eigen::Ref<const Eigen::VectorXd>& sol) const
+{
+  return tesseract_kinematics::getRedundantSolutions<double>(
+      sol, getLimits().joint_limits, getRedundancyCapableJointIndices());
```
marip8, line 34: Seems a bit odd to have such a simple wrapper around a helper function, especially if `getRedundancyCapableJointIndices()` is a public member of this interface. Is the benefit of adding this to the interface just convenience?

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
Levi-Armstrong: I do not see an issue with a pure virtual returning a const reference in this case because the information should be known at construction. This just forces that they are stored as member variables so they can be returned as a const reference.

marip8:
Have we considered creating a "motion group" or "manipulator group" class that contains information relative to how named sets of joints can move (i.e. joint names, joint limits, forward and inverse kinematic solvers, redundant joint indices)? I recall there being a "manipulator manager" that was similar to this concept, but didn't include much of this information. It seems like having a concept like this would significantly reduce the amount of duplicated code in the forward and inverse kinematics solvers and in other places in `tesseract`.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
Levi-Armstrong: I did not look but is this the case for the utility function you added?

marip8: I believe so. I made this comment before I realized this function just wraps that function

Levi-Armstrong:
> Have we considered creating a "motion group" or "manipulator group" class that contains information relative to how named sets of joints can move (e.g. joint names, joint limits, forward and inverse kinematic solvers, redundant joint indices, etc.)? I recall there being a "manipulator manager" that was similar to this concept, but didn't include much of this information. It seems like having a concept like this would significantly reduce the amount of duplicated code in the forward and inverse kinematics solvers and in other places in tesseract.

This is something to be developed over the next few months. I will setup a meeting to start the discussion.

marip8:
If we make these changes, we should probably also remove the function below and make the kinematics interfaces solely responsible for determining which joints can produce redundant solutions

https://github.com/ros-industrial-consortium/tesseract/blob/b653eb3d81a14d5d8da0b37354d11badad901fce/tesseract_kinematics/include/tesseract_kinematics/core/utils.h#L400-L419

tesseract_kinematics/src/core/inverse_kinematics.cpp
Levi-Armstrong: I added it just for convenience.

marip8:
> This is something to be developed over the next few months. I will setup a meeting to start the discussion.

I bring this up because it may not be worth breaking API multiple times if we think that this functionality will be moved to a different class in the near future

Levi-Armstrong:
> If we make these changes, we should probably also remove the function below and make the kinematics interfaces solely responsible for determining which joints can produce redundant solutions

I think having it in the utility reduces duplicating this code in all of the kinematics for now but may change after adding the concepts of models.

tesseract_kinematics/src/core/inverse_kinematics.cpp
Levi-Armstrong: I was on the fence about adding this. It is not used very many places so I would be ok with removing it. Having here though provided more visibility than as a separate utility function.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
marip8: > the information should be known at construction

For the majority of cases, yes. My issue with this is mainly conceptual. The restrictions an interface imposes should be as minimal as possible to keep the interface flexible. It's possible that a use-case might want to compute this dynamically so redundant solutions don't get generated for a particular joint. If you restrict the output type to being a reference, it makes things more difficult for that case. There's almost no performance penalty for returning the actual object, so it seems like we're making the interface more restrictive for no reason

marip8:
> I think having it in the utility reduces duplicating this code in all of the kinematics for now but may change after adding the concepts of models.

To me it seems more confusing to have this code in two places. Should you use this publicly visible function to find the redundant indices or should you ask the kinematics solver (which actually knows how that motion group moves) to tell you what its redundant joints are? The benefit of having the `getRedundantJointIndices` function in the kinematics solvers is so it can be customized to the way the kinematics are being performed. If they all just use the helper function under the hood, why add a helper to the kinematics solvers?

tesseract_kinematics/src/core/inverse_kinematics.cpp
marip8: I'm personally opposed to code duplication in the interest of visibility. If the helper function we have right now is not easy to find we should move it to a better place or document it better. I would suggest putting it as a static member of the IK class, but the FK class also needs to use it. I think that's why I put it in the utils file to begin with

Levi-Armstrong:
> > I think having it in the utility reduces duplicating this code in all of the kinematics for now but may change after adding the concepts of models.
> 
> To me it seems more confusing to have this code in two places. Should you use this publicly visible function to find the redundant indices or should you ask the kinematics solver (which actually knows how that motion group moves) to tell you what its redundant joints are? The benefit of having the `getRedundantJointIndices` function in the kinematics solvers is so it can be customized to the way the kinematics are being performed. 

Makes sense, I will remove the utility function.


> If they all just use the helper function under the hood, why add it to the kinematics solvers?

This usually for visibility and convience.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
Levi-Armstrong: > It's possible that a use-case might want to compute this dynamically so redundant solutions don't get generated for a particular joint.

I do see your point related to flexibility versus performance on the interface. In this case, I selected performance because I don't see how it is possible given the current interface where dynamically calculating the redundant indices would make sense because the function does not take any inputs?

tesseract_kinematics/src/core/inverse_kinematics.cpp
Levi-Armstrong: I am good with that. I will remove it and I will add documentation to the `calcInvKin` methods.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
Levi-Armstrong: It does not include the original solution. I have updated the documentation.

Levi-Armstrong:
@marip8 I have made most of the requested changes except for the one currently being discussed related to returning const reference.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
Levi-Armstrong: I made the change to the interface.

marip8 (approved):

PR #619, opened 2021-07-13
Add ability to synchronize inverse kinematics object to forward kinematics

Levi-Armstrong:
Currently, it is difficult and sometimes impossible through normal construction methods to create an inverse kinematics object where the joints_names, link_names, etc align. This adds a synchronize method only to the inverse kinematics class where it takes a forward kinematics object to align the internal data of the inverse kinematics object with. All inverse kinematics objects have been updated, but need to write unit tests to confirm things are working as expected.


marip8:
For what it's worth, this is another thing that should probably be handled at the model/motion group level instead of at the kinematics solver level. I think a model/motion group should have an ordered set of joints and the kinematics solvers associated with the model should have to accept those joints in the specified order

Levi-Armstrong:
> For what it's worth, this is another thing that should probably be handled at the model/motion group level instead of at the kinematics solver level. I think a model/motion group should have an ordered set of joints and the kinematics solvers associated with the model should have to accept those joints in the specified order

I agree, this is more of an immediate fix for @mohamedsayed18 because it is currently not possible to motion plan for a robot with an external positioner because it is impossible to configure the inverse kinematics object to align with the forward kinematics.

mohamedsayed18:
There is a problem building tesseract_environment
```
In member function ‘bool tesseract_environment::ManipulatorManager::addInvKinematicSolver(const Ptr&)’:
/media/mohamed/C03CCDB43CCDA62E/mx_space/src/motion-planner/tesseract/tesseract_environment/src/core/manipulator_manager.cpp:835:14: error: ‘const Ptr’ {aka ‘const class std::shared_ptr<tesseract_kinematics::InverseKinematics>’} has no member named ‘isSynchronized’
   if (solver.isSynchronized())
```

tesseract_environment/src/core/manipulator_manager.cpp
```diff
@@ -832,6 +832,10 @@ ManipulatorManager::getInvKinematicFactory(const std::string& name) const
 
 bool ManipulatorManager::addInvKinematicSolver(const tesseract_kinematics::InverseKinematics::Ptr& solver)
 {
+  if (solver.isSynchronized())
```
mohamedsayed18, line 835:
```suggestion
  if (solver->isSynchronized())
```

Levi-Armstrong: Yep. I just pushed the changes.

Levi-Armstrong:
Fix a segfault issue and now kinematics unit tests are passing.

tesseract_kinematics/include/tesseract_kinematics/core/inverse_kinematics.h
```diff
@@ -71,6 +72,18 @@ class InverseKinematics
    */
   virtual bool update() = 0;
 
+  /**
+   * @brief Synchronize inverse data to align with forward kinematics object
+   * @param fwd_kin The forward kinematics object to synchronize with
+   */
+  virtual void synchronize(ForwardKinematics::ConstPtr fwd_kin) = 0;
```
mohamedsayed18, line 79:
```suggestion
  virtual void synchronize(ForwardKinematics::ConstPtr fwd_kin) const = 0;
```

tesseract_kinematics/include/tesseract_kinematics/core/rep_inverse_kinematics.h
```diff
@@ -69,6 +70,9 @@ class RobotWithExternalPositionerInvKin : public InverseKinematics
 
   bool update() override;
 
+  void synchronize(ForwardKinematics::ConstPtr fwd_kin) override;
```
mohamedsayed18, line 73:
```suggestion
  void synchronize(ForwardKinematics::ConstPtr fwd_kin) const override;
```

tesseract_kinematics/include/tesseract_kinematics/core/rop_inverse_kinematics.h
```diff
@@ -67,6 +68,9 @@ class RobotOnPositionerInvKin : public InverseKinematics
 
   bool update() override;
 
+  void synchronize(ForwardKinematics::ConstPtr fwd_kin) override;
```
mohamedsayed18, line 71:
```suggestion
  void synchronize(ForwardKinematics::ConstPtr fwd_kin) const override;
```

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+  if (!init(scene_graph_, manip_inv_kin_, manip_reach_, positioner_fwd_kin_, positioner_sample_resolution_, name_))
+    return false;
+
+  if (sync_fwd_kin_ != nullptr)
+    synchronize(sync_fwd_kin_);
+
+  return true;
+}
+
+void RobotWithExternalPositionerInvKin::synchronize(ForwardKinematics::ConstPtr fwd_kin)
```
mohamedsayed18, line 58:
```suggestion
void RobotWithExternalPositionerInvKin::synchronize(ForwardKinematics::ConstPtr fwd_kin) const
```

tesseract_kinematics/src/core/rop_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotOnPositionerInvKin::update()
...
+  if (!init(scene_graph_, manip_inv_kin_, manip_reach_, positioner_fwd_kin_, positioner_sample_resolution_, name_))
+    return false;
+
+  if (sync_fwd_kin_ != nullptr)
+    synchronize(sync_fwd_kin_);
+
+  return true;
+}
+
+void RobotOnPositionerInvKin::synchronize(ForwardKinematics::ConstPtr fwd_kin)
```
mohamedsayed18, line 58:
```suggestion
void RobotOnPositionerInvKin::synchronize(ForwardKinematics::ConstPtr fwd_kin) const
```

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+  local_data.tip_link_name = fwd_kin->getTipLinkName();
+  local_data.joint_names = fwd_kin->getJointNames();
+  local_data.link_names = fwd_kin->getLinkNames();
+  local_data.active_link_names = fwd_kin->getActiveLinkNames();
+  local_data.redundancy_indices = fwd_kin->getRedundancyCapableJointIndices();
+  local_data.limits = fwd_kin->getLimits();
+  if (data_ == local_data)
+    return;
+
+  sync_joint_map_.clear();
```
mohamedsayed18, line 83:
```
error: passing ‘const std::vector<long int>’ as ‘this’ argument discards qualifiers [-fpermissive]
   sync_joint_map_.clear();
```

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+  const std::vector<std::string>& joint_names = fwd_kin->getJointNames();
+  if (orig_data_.joint_names != joint_names)
+  {
+    for (std::size_t i = 0; i < joint_names.size(); ++i)
+    {
+      auto it = std::find(orig_data_.joint_names.begin(), orig_data_.joint_names.end(), joint_names[i]);
+      Eigen::Index idx = std::distance(orig_data_.joint_names.begin(), it);
+      sync_joint_map_.push_back(idx);
```
mohamedsayed18, line 91:
```
error: passing ‘const std::vector<long int>’ as ‘this’ argument discards qualifiers [-fpermissive]
       sync_joint_map_.push_back(idx);
```

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+    }
+  }
+
+  sync_fwd_kin_ = std::move(fwd_kin);
```
mohamedsayed18, line 95:
```
error: no match for ‘operator=’ (operand types are ‘const ConstPtr’ {aka ‘const std::shared_ptr<const tesseract_kinematics::ForwardKinematics>’} and ‘std::remove_reference<std::shared_ptr<const tesseract_kinematics::ForwardKinematics>&>::type’ {aka ‘std::shared_ptr<const tesseract_kinematics::ForwardKinematics>’})
   sync_fwd_kin_ = std::move(fwd_kin);
```

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+  data_ = local_data;
```
mohamedsayed18, line 96:
```
error: passing ‘const tesseract_kinematics::SynchronizableData’ as ‘this’ argument discards qualifiers [-fpermissive]
   data_ = local_data;
```

tesseract_kinematics/src/core/rop_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotOnPositionerInvKin::update()
...
+  local_data.tip_link_name = fwd_kin->getTipLinkName();
+  local_data.joint_names = fwd_kin->getJointNames();
+  local_data.link_names = fwd_kin->getLinkNames();
+  local_data.active_link_names = fwd_kin->getActiveLinkNames();
+  local_data.redundancy_indices = fwd_kin->getRedundancyCapableJointIndices();
+  local_data.limits = fwd_kin->getLimits();
+  if (data_ == local_data)
+    return;
+
+  sync_joint_map_.clear();
```
mohamedsayed18, line 83:
```
error: passing ‘const std::vector<long int>’ as ‘this’ argument discards qualifiers [-fpermissive]
   sync_joint_map_.clear();
```

tesseract_kinematics/src/core/rop_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotOnPositionerInvKin::update()
...
+  const std::vector<std::string>& joint_names = fwd_kin->getJointNames();
+  if (orig_data_.joint_names != joint_names)
+  {
+    for (std::size_t i = 0; i < joint_names.size(); ++i)
+    {
+      auto it = std::find(orig_data_.joint_names.begin(), orig_data_.joint_names.end(), joint_names[i]);
+      Eigen::Index idx = std::distance(orig_data_.joint_names.begin(), it);
+      sync_joint_map_.push_back(idx);
```
mohamedsayed18, line 91:
```
error: passing ‘const std::vector<long int>’ as ‘this’ argument discards qualifiers [-fpermissive]
       sync_joint_map_.push_back(idx);
```

tesseract_kinematics/src/core/rop_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotOnPositionerInvKin::update()
...
+    }
+  }
+
+  sync_fwd_kin_ = std::move(fwd_kin);
```
mohamedsayed18, line 95:
```
error: no match for ‘operator=’ (operand types are ‘const ConstPtr’ {aka ‘const std::shared_ptr<const tesseract_kinematics::ForwardKinematics>’} and ‘std::remove_reference<std::shared_ptr<const tesseract_kinematics::ForwardKinematics>&>::type’ {aka ‘std::shared_ptr<const tesseract_kinematics::ForwardKinematics>’})
   sync_fwd_kin_ = std::move(fwd_kin);
```

tesseract_kinematics/src/core/rop_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotOnPositionerInvKin::update()
...
+  data_ = local_data;
```
mohamedsayed18, line 96:
```
error: passing ‘const tesseract_kinematics::SynchronizableData’ as ‘this’ argument discards qualifiers [-fpermissive]
   data_ = local_data;
```

mohamedsayed18 (commented):
I'm not if is right to changes these methods to const or it is a wrong fix

Levi-Armstrong:
Yea, that method cannot be const since it is supposed to modify the internal data of the class. Were you having an issue prior to making the method const?

mohamedsayed18:
> Yea, that method cannot be const since it is supposed to modify the internal data of the class. Were you having an issue prior to making the method const?

Error in tesseract planning repo I added it in the PR review there
```
/tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/problem_generators/default_problem_generator.h:75:3: error: passing ‘std::__shared_ptr_access<const tesseract_kinematics::InverseKinematics, (__gnu_cxx::_Lock_policy)2, false, false>::element_type’ {aka ‘const tesseract_kinematics::InverseKinematics’} as ‘this’ argument discards qualifiers [-fpermissive]
   prob->manip_inv_kin->synchronize(prob->manip_fwd_kin);
```

Levi-Armstrong:
> > Yea, that method cannot be const since it is supposed to modify the internal data of the class. Were you having an issue before making the method const?
> 
> Error in tesseract planning repo I added it in the PR review there
> 
> ```
> /tesseract_planning/tesseract_motion_planners/include/tesseract_motion_planners/descartes/problem_generators/default_problem_generator.h:75:3: error: passing ‘std::__shared_ptr_access<const tesseract_kinematics::InverseKinematics, (__gnu_cxx::_Lock_policy)2, false, false>::element_type’ {aka ‘const tesseract_kinematics::InverseKinematics’} as ‘this’ argument discards qualifiers [-fpermissive]
>    prob->manip_inv_kin->synchronize(prob->manip_fwd_kin);
> ```

I have pushed a commit to the planning repo fixing the issue.

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```diff
@@ -46,9 +46,58 @@ bool RobotWithExternalPositionerInvKin::update()
...
+  if (numJoints() != fwd_kin->numJoints())
+    throw std::runtime_error("Tried to synchronize kinematics objects with different number of joints!");
+
+  if (!tesseract_common::isIdentical(orig_data_.joint_names, fwd_kin->getJointNames(), false))
+    throw std::runtime_error("Tried to synchronize kinematics objects with different joint names!");
+
+  if (!tesseract_common::isIdentical(orig_data_.link_names, fwd_kin->getLinkNames(), false))
+    throw std::runtime_error("Tried to synchronize kinematics objects with different link names!");
+
+  if (!tesseract_common::isIdentical(orig_data_.active_link_names, fwd_kin->getActiveLinkNames(), false))
```
mohamedsayed18, line 69: It passes the link names, but fails in the active link names. This is how the active links look like
Forward: 
```
positioner_link_1, positioner_base_link, world, base_link, link_1, link_2, link_3, link_4, link_5, link_6, flange, tool0
```
Inverse:
```
flange, link_1, link_2, link_3, link_4, link_5, link_6, positioner_link_1, positioner_link_2, positioner_tool0, tool0
```

Levi-Armstrong: It looks like the forward kinematics list is incorrect. I now see why it is wrong because of how the kinematics chain is defined. I need to improve the way we are generating active link names. You can comment on this check which should not affect the motion planners. It will put a few static links in the active tree but they will never move during collision checking. I will update the unit tests to add this kinematic group to develop a better way for getting active links.

Levi-Armstrong: I just pushed a commit that should fix things.

mohamedsayed18: I Commented the active link check as it still not working. If it may be helpful here is how the active links look like
Forward:
```
positioner_link_1, link_1, link_2, link_3, link_4, link_5, link_6, flange, tool0
```
Inverse:
```
flange, link_1, link_2, link_3, link_4, link_5, link_6, positioner_link_1, positioner_link_2, positioner_tool0, tool0
```

Levi-Armstrong: Thank you. I will keep working on things. I should be able to write some unit tests this weekend to find the edge cases. I

Is the Inverse kinematics the REP inverse kinematics object?

mohamedsayed18: it is the ` orig_data_.active_link_names`. I noticed that you also log these data in the terminal

Levi-Armstrong: Alright, I have updated the generation of the active links within ROP and REP which I think should resolve the issue. Also fixed another issue in kdl generation of active links when the base is actually a child of a joint. Let me know if this changes anything on your end.

mohamedsayed18: It still fails on the active link condition
Forward (fwd_kin->getActiveLinkNames()):
```
positioner_tool0, positioner_link_1, link_1, link_2, link_3, link_4, link_5, link_6, flange, tool0,
```
Inverse (orig_data_.active_link_names):
```
flange, link_1, link_2, link_3, link_4, link_5, link_6, positioner_link_1, positioner_link_2, positioner_tool0, tool0,
```

Levi-Armstrong: Now the forward looks correct, but there is an extra link in the inverse. Is the urdf public where I can take a look?

mohamedsayed18: > Now the forward looks correct, but there is an extra link in the inverse. Is the urdf public where I can take a look?

Yes, I'm using the positioner from [tesseract_ros_workcell](https://github.com/mohamedsayed18/motion_planner/blob/main/tesseract_ros_workcell/twc_support/CAD/robot/abb_irbpa250_d1000/urdf/irbpa250_d1000_macro.xacro)
and the abb_robot from [here](https://github.com/mohamedsayed18/motion_planner/blob/main/abb_irb2600_support/urdf/irb2600_12_165_macro.xacro)

Levi-Armstrong: Ok I will look into it.

mohamedsayed18:
I'm not sure if this is related to the changes proposed here or not.
When I'm trying to make a program that consists of one way point using the `CARTESIAN_PLANNER_NAME` It generates the path but it gives the following error
https://github.com/ros-industrial-consortium/tesseract/blob/b777fa2425c01b82cec7b41de6b6e9f6ec903750/tesseract_kinematics/include/tesseract_kinematics/core/validate.h#L149-L153
When I execute the same program but with only the manipulator it works fine.

Levi-Armstrong:
> I'm not sure if this is related to the changes proposed here or not.
> When I'm trying to make a program that consists of one way point using the `CARTESIAN_PLANNER_NAME` It generates the path but it gives the following error
> 
> https://github.com/ros-industrial-consortium/tesseract/blob/b777fa2425c01b82cec7b41de6b6e9f6ec903750/tesseract_kinematics/include/tesseract_kinematics/core/validate.h#L149-L153
> 
> When I execute the same program but with only the manipulator it works fine.

Do you know how far off it is? I should update the message to provide this information.

mohamedsayed18:
>Do you know how far off it is? I should update the message to provide this information.

I gives 3.05 as I remember, but I'd like to understand what this value present when having multiple points in the path

Levi-Armstrong:
> > Do you know how far off it is? I should update the message to provide this information.
> 
> I gives 3.05 as I remember, but I'd like to understand what this value present when having multiple points in the path

This is just a check that the motion planners perform up front to make sure the forward and kinematics are correct. It performs a forward kinematics given for a random joint solution and then provides that to the IK solver and then check the solution form each of the solution to make sure produce the same forward kinematics results. The issues is that the forward and inverse may have different base links which means the forward kinematics results are in a different frame than what the inverse kinematics solver expects. The motion planners handle this function is not where it should take the output and transform it into the base link of the inverse kinematic solver before solving for solutions.

Levi-Armstrong:
This function should be update to do what is being done [here](https://github.com/Levi-Armstrong/tesseract_planning/blob/7ddb57c77d4ef23901711170aae9419464373a1a/tesseract_motion_planners/include/tesseract_motion_planners/descartes/impl/profile/descartes_default_plan_profile.hpp#L200-L201). Does everything motion plan OK?

mohamedsayed18:
> This function should be update to do what is being done [here](https://github.com/Levi-Armstrong/tesseract_planning/blob/7ddb57c77d4ef23901711170aae9419464373a1a/tesseract_motion_planners/include/tesseract_motion_planners/descartes/impl/profile/descartes_default_plan_profile.hpp#L200-L201). Does everything motion plan OK?

I have both the manipulator and external axis with the same base link `world` otherwise it will give the following error
```
Provided positioner and manipulator base link are not the same
at line 356 in src/motion-planner/tesseract/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
```
The motion plan is ok when I'm working with `ManipulatorInfo` as the manipulator.
Note: the translation norm error doesn't appear when using the trajopt planner

mohamedsayed18:
when  `ManipulatorInfo` is set to the manipulator and the external axis and I try to execute two consicutive cartesian waypoints
It didn't generate a path even if I keep it  running for a very long time (30 min). It shows no error or warning messages just running,whether I'm using Cartesian planner or Trajopt planner.
I tried to execute a program consits of the following four points in order: State Wp1, Cart WP2, State Wp1(return to the first point), Cart WP3. It generates the path fine.

Levi-Armstrong:
@mohamedsayed18 Is your code somewhere I can pull down and test?

mohamedsayed18:
> @mohamedsayed18 Is your code somewhere I can pull down and test?

https://github.com/mohamedsayed18/motion_planner/tree/positioner

Levi-Armstrong:
@mohamedsayed18 I think the issue was that the inv kinematics was synchronizing the base and tip link which it should not have been. I have also update the checkKinematics and updated the tesseract_planners. Let me know if this solve the issue. 

I also noticed that in your example you are copying in the source code from various repos and it may be worth using gitmodules for each of the repositories to make it easy to update and switching between different remotes and branches.

tesseract_kinematics/src/core/rep_inverse_kinematics.cpp
mohamedsayed18: it still fails in the active links
fwd_kin->getActiveLinkNames():
```
positioner_link_1, link_1, link_2, link_3, link_4, link_5, link_6, flange, tool0`
```
orig_data_.active_link_names
```
flange, link_1, link_2, link_3, link_4, link_5, link_6, positioner_link_1, positioner_link_2, positioner_tool0, tool0,
```
I commented it for now. 
Does this error affect the synchronization?

mohamedsayed18:
It still takes a long time without showing any solution. I turned the debug on
```
Info:    Tesseract Planning Server Received Request!
Info:    Tesseract Planning Server: Input program required formatting!
Debug:   worker 4 ready to run Has Seed Check
Debug:   Seed is empty!
Debug:   worker 4 finished running Has Seed Check
Debug:   worker 3 ready to run Interpolator
Debug:   Profile FREESPACE was not found. Using default if available. Available profiles:
Debug:   DEFAULT
Debug:   Profile FREESPACE was not found. Using default if available. Available profiles:
Debug:   DEFAULT
```
The `Debug:   Motion Planner process succeeded` never shows.
When running the same program only changing the ManipulatorInfo  to `manipulator` the path is generated

tesseract_kinematics/include/tesseract_kinematics/kdl/kdl_utils.h
```diff
@@ -185,68 +182,88 @@ inline bool parseSceneGraph(KDLChainData& results,
...
 
     if (jnt.getType() == KDL::Joint::None)
       continue;
 
     if (!found)
     {
       found = true;
-      results.active_link_list.push_back(seg.getName());
+      results.data.active_link_names.push_back(seg.getName());
     }
```
mohamedsayed18, line 221: is `active_link_names` set here? 
If you want to make active_link_names in the right order, should I do changes to this file?

Levi-Armstrong:
I was running this yesterday and noticed that the interpolate planner is generating odd results. I went back to the master branch and everything works as expected with just the manipulator but does not on this branch. I am working on tracking down the issue.

mohamedsayed18:
> I was running this yesterday and noticed that the interpolate planner is generating odd results. I went back to the master branch and everything works as expected with just the manipulator but does not on this branch. I am working on tracking down the issue.

I checkout the master branch in tesseract and tesseract_planning and it gives the following error
`Error:   SimplePlanner failed to generate problem: Forward and Inverse Kinematic objects joints are not ordered the same!.`
This is how the joints looks like
inv_kin->getJointNames()
```
positioner_base_joint
positioner_joint_1
joint_1
joint_2
joint_3
joint_4
joint_5
joint_6
```
fwd_kin->getJointNames()
```
positioner_joint_1
positioner_base_joint
joint_1
joint_2
joint_3
joint_4
joint_5
joint_6
```

mohamedsayed18:
As I understand the active links are child links whose joints are moving(active joints). but when checking the output of the following operation
https://github.com/ros-industrial-consortium/tesseract/blob/59f84c7827d862b48af4f7fe1f3677e6fa4c3ff4/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L472-L479
contain links like `base_link`, `flange`, `tool0` and `world`

mohamedsayed18:
Something I also noticed is that when I have only one waypoint in the program, it generated the path but when having two cartesian waypoints consecutive in the program it takes a very long time to find a solution and didn't return the solution at the end.

Levi-Armstrong:
Yea with the recent changes the simple planner is generating incorrect results causing issues with the other motion planners. I am trying to track down what was changes that is causing the issue.

mohamedsayed18:
I checked the size of the joint solutions in the following method
https://github.com/ros-industrial-consortium/tesseract_planning/blob/6907026937ed60b0344e5844ddb97e21c959fab3/tesseract_motion_planners/src/simple/profile/simple_planner_utils.cpp#L221-L244
When having only the manipulator the size of j1 = 12 and j2= 12
When having the manipulator and the external axis j1 = 1636212 and j2 = 1582972

tesseract_kinematics/include/tesseract_kinematics/kdl/kdl_utils.h
Levi-Armstrong: I don't think the active links should need to be in the same order. If the check is requiring this then it should be updated.

Levi-Armstrong:
> As I understand the active links are child links whose joints are moving(active joints). but when checking the output of the following operation
> 
> https://github.com/ros-industrial-consortium/tesseract/blob/59f84c7827d862b48af4f7fe1f3677e6fa4c3ff4/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L472-L479
> 
> contain links like `flange`, `tool0` and `positioner_tool0` as it is already contained in the `full_active_link_names`
> 
> Should we in the following loop ignore fixed joints
> 
> https://github.com/ros-industrial-consortium/tesseract/blob/59f84c7827d862b48af4f7fe1f3677e6fa4c3ff4/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L453-L468

They are fixed joint types, but they are added to the list because they can move if one of the joints in the kinematic object changes. This is primarily used to update the contact checker during motion planning.

Levi-Armstrong:
> Something I also noticed is that when I have only one waypoint in the program, it generated the path but when having two cartesian waypoints consecutive in the program it takes a very long time to find a solution and didn't return the solution at the end.

I assume this is when trying to plan with an external positioned? If so, what process planner are you using?

Levi-Armstrong:
> I checked the size of the joint solutions in the following method
> https://github.com/ros-industrial-consortium/tesseract_planning/blob/6907026937ed60b0344e5844ddb97e21c959fab3/tesseract_motion_planners/src/simple/profile/simple_planner_utils.cpp#L221-L244
> When having only the manipulator the size of j1 = 12 and j2= 12
> When having the manipulator and the external axis j1 = 1636212 and j2 = 1582972

Yea that looks about right. At every sampled positioner state it solves IK and adds it to the list. I just pushed a few more commits which resolved a few issues. Though there is more work to be done to get things working with an external positioner. I could quickly get it working with just the TrajOpt planner but if you need to use descartes then it and the simple planner need to be updated to handle when the working frame is a active link. This has already been done with TrajOpt.

mohamedsayed18:
> I assume this is when trying to plan with an external positioned? If so, what process planner are you using?

Yes with the positioner, the planner is TRAJOPT_PLANNER_NAME

Levi-Armstrong:
> > I assume this is when trying to plan with an external positioned? If so, what process planner are you using?
> 
> Yes with the positioner, the planner is TRAJOPT_PLANNER_NAME

OK, I will get a working example using just the trajopt planner and then work on updating the simple and descartes planner to support the working frame being an active link.

mohamedsayed18:
I found that there is a problem here
https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_environment/src/core/manipulator_manager.cpp#L1180-L1186
When the  manipulator and the positioner have the same base link ('world' in my case) the transform is left as identity
and this causes the problem
When I hardcoded added the transformation in and assign it in this line
https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L417
I was able to generate a path and the Trajopt solves the problem

Levi-Armstrong:
@mohamedsayed18 If they both have the same base link shouldn't the transform between the two be identity, or am I missing something?

mohamedsayed18:
> @mohamedsayed18 If they both have the same base link shouldn't the transform between the two be identity, or am I missing something?

True, The base link of the manipulator and the positioner are both connected by the `world` by fixed joints.
I thought it could be related to the real base link(active link) of the positioner and the manipulator.

Another note the following check is not showing
https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L333-L337
I think this method is no longer being invoked and the following is the one being used
https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L349

I tried with having different base links but it is giving the deadlock error above

mohamedsayed18:
A question I have is what `positioner_tf` represents
https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L133
is the transformation between the base link of the manipulator(full_manipulator) to the end effector of the positioner?

Levi-Armstrong:
> A question I have is what `positioner_tf` represents
> 
> https://github.com/ros-industrial-consortium/tesseract/blob/a26836664c4dc0177627845c7dedfc0d4fc6c99c/tesseract_kinematics/src/core/rep_inverse_kinematics.cpp#L133
> 
> Is it the transformation between the base link of the manipulator(full_manipulator) to the end effector of the positioner?

This should be the tip_link of the positioner in the base frame of the positioner.

Also, I am currently reworking the kinematics to better support multi kinematic planning. Should have things wrapped up this week and will use the robot with an external positioner as the test case.

Levi-Armstrong:
@mohamedsayed18 I just wanted to give you an update. I am working on a different branch which builds from this one feature/AddKinematicGroup and hope to have things wrapped up early next week.

Levi-Armstrong:
@mohamedsayed18 I have the kinematic group implemented and unit tests passing. I will have an internal review on Monday and then go through and update the rest of Tesseract to leverage the new KinematicGroup which will solve the issues with the external positioner.

mohamedsayed18:
Hi Levi, I really appreciate your work.
Are there any updates about the new KinematicGroup? When is it expected to be ready?

Levi-Armstrong:
Sorry, it is taking longer than expected. Since we are breaking the interface we decided to make some additional changes to fully support plugins and a few other things. So far in this repository, everything has been updated except the tesseract_environment. I am currently working on it now and hope to have it wrapped up this week. Next, I will start on updating tesseract_planning and tesseract_ros which should not be a long task and should have that wrapped up the following week.

Levi-Armstrong:
This has been replaced by PR #630

PR #633, opened 2021-09-14
Feature/write urdf

DavidMerzJr:
Write functions written and implemented to correspond to all tesseract_urdf::parseX calls, with the exception of writePointCloud.  Point clouds are always used to construct octomaps on load, so the original cloud is not around to write back out.  Feedback and suggestions on how to test would be welcome.  Some changes in tesseract_geometry were required to get meshes and octomaps writing out correctly.

@Levi-Armstrong let me know what more you would like done here.

### DIFF
--- a/tesseract_geometry/include/tesseract_geometry/impl/mesh.h
+++ b/tesseract_geometry/include/tesseract_geometry/impl/mesh.h
@@ -33,9 +33,10 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <memory>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
-#include <tesseract_geometry/geometry.h>
 #include <tesseract_common/types.h>
+#include <tesseract_geometry/geometry.h>
 #include <tesseract_geometry/impl/mesh_material.h>
+#include <tesseract_geometry/impl/polygon_mesh.h>
 
 #ifdef SWIG
 %shared_ptr(tesseract_geometry::Mesh)
@@ -44,7 +45,7 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_geometry
 {
-class Mesh : public Geometry
+class Mesh : public PolygonMesh
 {
 public:
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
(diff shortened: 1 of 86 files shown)


marip8:
Did you consider using the `boost` serialization library instead of manually defining the serialization?

DavidMerzJr:
No, I did not.  We specifically need this to be written out in the URDF XML format.  Can boost serialization do that?

marip8:
I don't think it can write attributes to elements, so it probably won't work here actually

Levi-Armstrong:
@DavidMerzJr It looks like it needs to rebase on master and most of the CI is failing.

tesseract_geometry/include/tesseract_geometry/impl/mesh.h
```diff
@@ -156,26 +156,32 @@ class Mesh : public PolygonMesh
...
-                                    getFaces(),
-                                    getFaceCount(),
-                                    getResource(),
-                                    getScale(),
-                                    getNormals(),
-                                    getVertexColors(),
-                                    std::make_shared<MeshMaterial>(*getMaterial()),
-                                    getTextures());
+    {
+      ptr = std::make_shared<Mesh>(getVertices(),
```
Levi-Armstrong, line 162: I would do what you originally had just remove the else.

```
    if (getMaterial() != nullptr)
      return std::make_shared<Mesh>(getVertices(),
                                    getFaces(),
                                    getFaceCount(),
                                    getResource(),
                                    getScale(),
                                    getNormals(),
                                    getVertexColors(),
                                    std::make_shared<MeshMaterial>(*getMaterial()),
                                    getTextures());

    return std::make_shared<Mesh>(getVertices(),
                                    getFaces(),
                                    getFaceCount(),
                                    getResource(),
                                    getScale(),
                                    getNormals(),
                                    getVertexColors(),
                                    nullptr,
                                    getTextures());
```

Levi-Armstrong:
Also will need full unit test coverage for what was added. Take a look at the current unit tests for the urdf. You could probably reuse most of it.

DavidMerzJr:
I pushed a small change to add the XML declaration to the output files.  Before that change, everything but the Windows build was passing.  The Windows build was still running after more than 1:45:00, so I'm not sure whether it was going to succeed.  The error in the CI log was:
```
Error C1083 Cannot open include file: 'pcl/io/vtk_lib_io.h': No such file or directory
```
As suggested [here](https://stackoverflow.com/questions/64374233/vcpkg-pcl-cannot-open-include-file-pcl-io-vtk-lib-io-h-no-such-file-or-dire) and [here](https://stackoverflow.com/questions/67320722/pcl-library-setup-using-vcpkg-seems-to-be-missing-include-files-pcl-visualizati?rq=1), this seems to be an issue with the 'normal' pcl distribution on vcpkg / Windows not including the PCL Visualization module / VTK-oriented code.  Presumably, this is a problem now and not previously because of my inclusion of `pcl/io/lib_vtk_io.h`.  At the advice of @Levi-Armstrong , I tried adding `BEFORE` commands to inclusions of `PCL_INCLUDE_DIRS` and `PCL_LIBRARY_DIRS`.  When that failed, I tried adding an explicit installation of pcl to the ci file, which resulted in the above-mentioned very long CI runtime.  Still waiting to see if it finishes.

Levi-Armstrong:
I looks like what you found is correct that it does not have it so it may be best to remove it and use the PLY writer [here](https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract_collision/include/tesseract_collision/core/common.h) to avoid the need for PLC/VTK IO.

PR #637, opened 2021-10-04
Added ikfast compatibility with 7 DOF robots

marrts:
Added support for 7 DOF robots for ikfast. 7 DOF robots require an additional input to the ComputeIk method that was previously hard coded as a nullptr, which works for robots with fewer degrees of freedom. This change doesn't break any of the interfaces, but allows users the option of inputting a vector of doubles that represent the values the "free" 7th degree of freedom should be sampled at when computing the IK.


marrts:
This was tested with an ikfast genrated from an iiwa7. The implementation can be found [here](https://github.com/marrts/iiwa_ikfast_kinematics/tree/tesseract_ikfast).

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
```diff
@@ -160,6 +162,7 @@ class IKFastInvKin : public InverseKinematics
   std::vector<Eigen::Index> redundancy_indices_; /**< @brief Joint indicies that have redundancy (ex. revolute) */
   std::string name_;                             /**< @brief Name of the kinematic chain */
   std::string solver_name_;                      /**< @brief Name of this solver */
+  std::vector<double> free_joint_values_;        /**< @brief values to sample the free joint in case of a 7 DOF robot */
```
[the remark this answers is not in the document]
marrts: I thought about this as well. In the MoveIt docs I [found](url) it says: 

> While it works in theory, currently the IKFast plugin generator tool does not work with >7 degree of freedom arms.

This could be outdated, but I am unsure. I am not sure exactly what the interface would look like if we are giving multiple values to the ComputeIk method, so that is why I restrained from doing what you suggested

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
```diff
@@ -122,6 +122,7 @@ class IKFastInvKin : public InverseKinematics
   const std::string& getTipLinkName() const override;
   const std::string& getName() const override;
   const std::string& getSolverName() const override;
+  std::vector<double> getFreeJointValues() const;
```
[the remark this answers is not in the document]
marrts: No, I only included it because there was a method to get every other protected class member, but I can remove it

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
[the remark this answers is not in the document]
marrts: This has now been implemented to support >7 DOF

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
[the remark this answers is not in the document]
marrts: There function has been added back now because there is a new method for when using >7 DOF IKfast implementations that allows for the creation of all possible joint combinations. The user may want to extract this data, so I think it makes sense to provide this method now.

tesseract_kinematics/include/tesseract_kinematics/ikfast/impl/ikfast_inv_kin.hpp
```diff
@@ -212,6 +236,29 @@ bool IKFastInvKin::checkInitialized() const
...
+    {
+      if (curr_joint_indices[i] == free_joint_samples[i].size())
+      {
+        curr_joint_indices[i] = 0;
+        curr_joint_indices[i - 1]++;
+      }
+    }
+  }
+  return;
+}
```
marip8, line 260: I would make this a static function that takes the possible values for each joint as input and returns a vector of "free-joint states". Using this function to generate the input to the constructor seems a little more straightforward than initializing it with an empty set of free-joint states and overwriting that in a non-const way

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
```diff
@@ -160,6 +173,9 @@ class IKFastInvKin : public InverseKinematics
   std::vector<Eigen::Index> redundancy_indices_; /**< @brief Joint indicies that have redundancy (ex. revolute) */
   std::string name_;                             /**< @brief Name of the kinematic chain */
   std::string solver_name_;                      /**< @brief Name of this solver */
+  /**< @brief combinations of free joints to sample when computing IK
+   * Example: Given 3 free joints, a valid input would be [[0,0,0][0,0,1][-1,0,1][0,2,0]] */
+  std::vector<std::vector<double>> free_joint_combos_;
```
marip8, line 178: Being a bit pedantic, but I think the name `free_joint_states` is more clear than "free_joint_combos" since a "state" in the sense of `tesseract` implies values for various joints in a single robot pose

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
```diff
@@ -141,14 +143,25 @@ class IKFastInvKin : public InverseKinematics
...
 
+  /**
+   * @brief Generates all possible combinations of joint states and stores it to the free_joint_combos_ class member
+   * Example: Given 2 free joints, wanting to sample the first joint at 0, 1, and 2 and the second joint at 3 and 4
+   * the input would be [[0, 1, 2][3,4]] and it would generate [[0,3][0,4][1,3][1,4][2,3][2,4]]
+   * @param free_joint_samples A vector of vectors in which the lower level vectors each represent all of a single
+   * joint's possible positions to be sampled
+   * @return
+   */
+  void generateAllPossibleCombinations(const std::vector<std::vector<double>> free_joint_samples);
```
marip8, line 163:
```suggestion
  void generateAllFreeJointStateCombinations(const std::vector<std::vector<double>> free_joint_samples);
```

Similarly pedantic, but adds more clarity. The documentation is helpful though

tesseract_kinematics/include/tesseract_kinematics/ikfast/impl/ikfast_inv_kin.hpp
```diff
@@ -201,6 +224,7 @@ const std::string& IKFastInvKin::getBaseLinkName() const { return base_link_name
 const std::string& IKFastInvKin::getTipLinkName() const { return tip_link_name_; }
 const std::string& IKFastInvKin::getName() const { return name_; }
 const std::string& IKFastInvKin::getSolverName() const { return solver_name_; }
+std::vector<std::vector<double>> IKFastInvKin::getFreeJointCombos() const { return free_joint_combos_; };
```
marip8, line 227: If you change `generateAllPossibleCombinations` to be a static function, I don't think there would be a need for this getter function anymore since all combinations would have already been passed to the constructor

tesseract_kinematics/include/tesseract_kinematics/ikfast/impl/ikfast_inv_kin.hpp
marrts: That's a much better way of doing it. Updated to match your suggestion.

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
marrts: That works, I was just trying to make the language as clear as possible, especially considering we will be providing a different function that takes exactly the same variable type but it represents something else.

tesseract_kinematics/include/tesseract_kinematics/ikfast/ikfast_inv_kin.h
marrts: Agreed, this matches the other renaming you requested

tesseract_kinematics/include/tesseract_kinematics/ikfast/impl/ikfast_inv_kin.hpp
marrts: Removed again.

marip8 (approved):
Looks good. Thanks for making the changes. @Levi-Armstrong are there any conflicts between this branch and your IK refactor branch, and if so how do you want to proceed?

Levi-Armstrong:
replaced by #657

PR #638, opened 2021-10-05
WIP: Fix/write urdf

DavidMerzJr:
Addressing various bugs and requests.  Also expanded unit tests and added more comments / documentation.

Still not really sure about how we should be passing in the package path / name for relative paths - the interface looks a lot like before, although the output now fits the `package://mypackage/collision/link1.py` style.

Also, how should material textures be handled?  Are those ordinarily stored in packages, with the same styling?


Levi-Armstrong:
@DavidMerzJr What is the status of this PR?

Levi-Armstrong:
merged in by #689

PR #641, opened 2021-10-13
Add sub-directories for external dependencies of tesseract_collision

marip8:
This PR divides `tesseract_collision` into sub-directories based on the optional external dependencies each sub-directory introduces (`bullet`, `FCL`, `VHACD`).

Note: the function `createConvexHull` in the library `tesseract_collision_core` currently depends on `bullet` API, so I left the function signature in `core` but moved the implementation to the `tesseract_collision_bullet` library. This means that any code that wants to use `createConvexHull` (i.e. functions in `tesseract_urdf`) needs to link against the `tesseract_collision_bullet` library until we find a better solution for uncoupling the creation of convex hulls from `bullet`.


tesseract_collision/CMakeLists.txt
```diff
@@ -82,292 +59,54 @@ set(COVERAGE_EXCLUDE
...
+# Examples
+if(TESSERACT_BUILD_BULLET AND TESSERACT_BUILD_FCL)
+  add_subdirectory(examples)
+endif()
+
+# Package configuration
+install(EXPORT ${PROJECT_NAME}-targets NAMESPACE "tesseract::" DESTINATION lib/cmake/${PROJECT_NAME})
+export(EXPORT ${PROJECT_NAME}-targets NAMESPACE "tesseract::"
+       FILE ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-targets.cmake)
+configure_package()
```
Levi-Armstrong, line 108: This can now be replaced with `configure_pacakge(NAMESPACE tesseract)`.

marip8: Does this require `ros-industrial-cmake-boilerplate` > 0.2.11? On my machine the `tesseract_collision-targets` file does not get installed without explicitly doing it

Levi-Armstrong: It requires master branch.

marip8: Can you push a tag 2.12.0 for that change? I'll update the `dependencies.rosinstall` once you do that and change this line. It would also be nice if you could start another debian release for the new tag

marip8:
`qhull` seems relatively difficult to integrate into `tesseract_collision` at the moment, so we decided to look into using `boost` to create convex hulls. It has an algorithm, but according to [this issue](https://github.com/boostorg/geometry/issues/683) it only works in 2D and not in 3D. I replicated the test case from this issue and was able to confirm that it does not in fact work correctly in 3D.

If we decide to go the `qhull` route, I would like to do it in a separate PR for more clarity on the changes. @Levi-Armstrong what do you think about this?

Levi-Armstrong:
How about just moving the createConvexHull to bullet directory. It should only require a few headers update and then the linking is not straight forward.

PR #653, opened 2021-10-25
Overload checkTrajectory to find points of collision

ben-greenberg:
`checkTrajectory` overload added to populate a vector of trajectory points in collision


Levi-Armstrong:
@ben-greenberg It turns out this should have been possible with the currently implementation but it was not correctly populating the results. The  following PR #698 should enable the you to determine the first segment with a collision and then you could also inspect the results for a segment to determine where in the segment it had a collision using the cc_time and cc_type. The cc_time returns a value between 0-1 indicating the interpolation length so cc_time=0.5 indicates the collision occurred halfway between the two states.

PR #672, opened 2021-12-02
Adds adds an additional AllowedCollisionMatrix to the CollisionCheckConfig

mpowelson:
This allows you to do things like disable collisions between certain links in binary checks like OMPL samplers and post checks without globally disabling them for other Tasks like Trajopt.

Some related changes

- Moves the ACM into tesseract_common to avoid tesseract_collision depending on tesseract_scene_graph. 
- Adds applyCollisionCheckConfig to the contact managers
- Correct some places where the collision margin data in the CollisionCheckConfig was not being applied


tesseract_collision/core/include/tesseract_collision/core/continuous_contact_manager.h
```diff
@@ -262,6 +262,12 @@ class ContinuousContactManager
    * @param type The type of contact test
    */
   virtual void contactTest(ContactResultMap& collisions, const ContactRequest& request) = 0;
+
+  /**
+   * @brief Applies settings in the config
+   * @param config Settings to be applies
+   */
+  virtual void applyCollisionCheckConfig(const CollisionCheckConfig& config);
```
Levi-Armstrong, line 270: It think it may be best to avoid using the CollisionCheckConfig directly because it contains other information which is more to be used for executing a contact test. This could be confusing from a user perspective. What about changing it to something along the line of what is shown bellow?

``` c++
virtual void setIsContactAllowedFn(IsContactAllowedFn fn, ContactAllowedFnOverrideType override_type = ACMOverrideType::REPLACE);

virtual void setIsContactAllowedFn(tesseract_common::AllowedCollisionMatrix acm, ContactAllowedFnOverrideType override_type = ContactAllowedFnOverrideType::REPLACE);
```

tesseract_collision/core/include/tesseract_collision/core/types.h
```diff
@@ -239,6 +240,20 @@ enum class CollisionEvaluatorType
...
+/** @brief Identifies how the provided AllowedCollisionMatrix should be applied relative to the isAllowedFn in the
+ * contact manager */
+enum class ACMOverrideType
+{
+  /** @brief Do not apply AllowedCollisionMatrix */
+  NONE,
+  /** @brief Replace the current IsContactAllowedFn with one generated from the ACM provided */
+  ASSIGN,
+  /** @brief New IsContactAllowedFn combines the contact manager fn and the ACM generated fn with and AND */
+  AND,
```
Levi-Armstrong, line 252: This is interesting. Did you have a use case in mind?

Levi-Armstrong (changes requested):
Overall looks good. One thing to discuss further.

tesseract_collision/core/include/tesseract_collision/core/types.h
mpowelson: Not specifically. Maybe you want to do a collision check of 2 arms and only ignore pairs in the acm for both manipulators?

mpowelson:
@Levi-Armstrong I can clean up the history if you like it. I'm not sure why CodeCov doesn't like those lines. They are tested in tesseract_environment_utils.cpp.  Maybe because that is a downstream package?

Levi-Armstrong:
I will review again. Sometimes I believe the codecov does not update due to ccache, because if you run it locally it is different and after a nightly build it will update.

Levi-Armstrong (approved):

Levi-Armstrong:
I would clean up the history before merging.

Levi-Armstrong:
I am also fine if you just squash merge.

PR #676, opened 2021-12-07
Add redundancy capable joints to the harmonizeTowardZero function

Levi-Armstrong:
This function cannot be used to harmonize non-revolute joints. The redundancy cable joints identifies which joint has the possibility of redundant solutions. For example this should not be applied to prismatic joints.

### DIFF
--- a/tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
+++ b/tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
@@ -336,67 +336,6 @@ inline Manipulability calcManipulability(const Eigen::Ref<const Eigen::MatrixXd>
   return manip;
 }
 
-///**
-// * @brief Create a kinematics map from the srdf model
-// * @param scene_graph Tesseract Scene Graph
-// * @param kinematics_information Tesseract Kinematics Information
-// * @return Kinematics map between group name and kinematics object
-// */
-// template <class Chain_T, class Tree_T>
-// ForwardKinematicsConstPtrMap createKinematicsMap(const tesseract_scene_graph::SceneGraph::ConstPtr& scene_graph,
-//                                                 const tesseract_srdf::KinematicsInformation& kinematics_information)
-//{
-//  ForwardKinematicsConstPtrMap manipulators;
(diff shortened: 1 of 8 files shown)


tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
```diff
@@ -535,12 +474,13 @@ inline bool isValid(const std::array<FloatType, 6>& qs)
  * @param dof The length of the float array
  */
 template <typename FloatType>
-inline void harmonizeTowardZero(Eigen::Ref<VectorX<FloatType>> qs)
+inline void harmonizeTowardZero(Eigen::Ref<VectorX<FloatType>> qs,
+                                const std::vector<Eigen::Index>& redundancy_capable_joints)
```
mpowelson, line 478: This is a breaking change. Should we add a default value or a deprecated version that assigns it to all joints?

marip8: It would be nice to minimize API breakage when possible. I don't think a default value would work; an empty vector would result in a no-op function and you can't programmatically generate a range of indices to match the size of the input `qs` with a default variable. You could make the previous version of the function a wrapper for the new version though:

```suggestion
inline void harmonizeTowardZero(Eigen::Ref<VectorX<FloatType>> qs)
{
  // Programmatically make a vector of the default indices (i.e. a range of all indices with std::iota, etc.)
  VectorX<FloatType> idx = ...;
  
  // Call the new version of the function
  return harmonizeTowardZero(qs, idx);
}

```

marip8:
Can you provide a description of what this PR is and why it's necessary? The motivation is not obvious from the title and it's beneficial to have good documentation of the changes being made to this repository

tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
Levi-Armstrong: I made the function deprecated to be removed at a later time.

mpowelson (approved):

PR #682, opened 2021-12-13
Move checkKinematics to getKinematicGroup and add support for clang-tidy-12

Levi-Armstrong:
Previously this was called in the motion planners resulting in multiple calls impacting planning time. This has been moved here to only run when a new instance of the kinematics object is created.

### DIFF
--- a/tesseract_environment/src/environment.cpp
+++ b/tesseract_environment/src/environment.cpp
@@ -29,6 +29,7 @@
 #include <tesseract_collision/core/common.h>
 #include <tesseract_srdf/utils.h>
 #include <tesseract_state_solver/ofkt/ofkt_state_solver.h>
+#include <tesseract_kinematics/core/validate.h>
 
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <queue>
@@ -397,6 +398,14 @@ tesseract_kinematics::KinematicGroup::UPtr Environment::getKinematicGroup(const 
       group_name, joint_names, std::move(inv_kin), *scene_graph_const_, current_state_);
 
   kinematic_group_cache_[group_name] = std::make_unique<tesseract_kinematics::KinematicGroup>(*kg);
+
+#ifndef NDEBUG
+  if (!tesseract_kinematics::checkKinematics(*kg))
+  {
+    CONSOLE_BRIDGE_logError("Check Kinematics failed. This means that inverse kinematics solution for a pose do not "
+                            "match forward kinematics solution. Did you change the URDF recently?");
+  }
+#endif
 
   return kg;
 }
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,6 +18,7 @@ Checks: >
     -cppcoreguidelines-non-private-member-variables-in-classes,
     misc-*,
     -misc-non-private-member-variables-in-classes,
+    -misc-no-recursion,
     modernize-*,
     -modernize-use-trailing-return-type,
     -modernize-use-nodiscard,
@@ -26,7 +27,8 @@ Checks: >
     -readability-braces-around-statements,
     -readability-named-parameter,
     -readability-magic-numbers,
-    -readability-isolate-declaration
+    -readability-isolate-declaration,
+    -readability-function-cognitive-complexity
 WarningsAsErrors: >
     -*,
     clang-diagnostic-*,
@@ -46,6 +48,7 @@ WarningsAsErrors: >
     -cppcoreguidelines-non-private-member-variables-in-classes,
     misc-*,
     -misc-non-private-member-variables-in-classes,
+    -misc-no-recursion,
     modernize-*,
     -modernize-use-trailing-return-type,
     -modernize-use-nodiscard,
@@ -54,7 +57,8 @@ WarningsAsErrors: >
     -readability-braces-around-statements,
     -readability-named-parameter,
     -readability-magic-numbers,
-    -readability-isolate-declaration
+    -readability-isolate-declaration,
+    -readability-function-cognitive-complexity
 HeaderFilterRegex: '.*'
 AnalyzeTemporaryDtors: false
 FormatStyle:     none
--- a/.github/workflows/api_docs.yml
+++ b/.github/workflows/api_docs.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - master
 
+  # allow manually starting this workflow
+  workflow_dispatch:
+
 jobs:
   deploy_documentation:
     runs-on: ubuntu-18.04
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -3,6 +3,9 @@ name: Benchmarking
 on:
   push:
 
+  # allow manually starting this workflow
+  workflow_dispatch:
+
 jobs:
   industrial_ci:
     name: ${{ matrix.env.CI_NAME }}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -6,6 +6,9 @@ on:
   schedule:
     - cron: '0 5 * * *'
 
+  # allow manually starting this workflow
+  workflow_dispatch:
+
 jobs:
   industrial_ci:
     if: contains(github.event.pull_request.labels.*.name, 'check-tesseract-ros') || github.event.schedule == true
--- a/tesseract_collision/bullet/include/tesseract_collision/bullet/bullet_utils.h
+++ b/tesseract_collision/bullet/include/tesseract_collision/bullet/bullet_utils.h
@@ -584,7 +584,7 @@ inline btScalar addCastSingleResult(btManifoldPoint& cp,
   }
   else
   {
-    bool castShapeIsFirst = (cd0->m_collisionFilterGroup == btBroadphaseProxy::KinematicFilter) ? true : false;
+    bool castShapeIsFirst = (cd0->m_collisionFilterGroup == btBroadphaseProxy::KinematicFilter);
     btVector3 normalWorldFromCast = -(castShapeIsFirst ? 1 : -1) * cp.m_normalWorldOnB;
     const btCollisionObjectWrapper* firstColObjWrap = (castShapeIsFirst ? colObj0Wrap : colObj1Wrap);
     const btTransform& first_tf_inv = (castShapeIsFirst ? tf0_inv : tf1_inv);
--- a/tesseract_collision/bullet/src/bullet_factories.cpp
+++ b/tesseract_collision/bullet/src/bullet_factories.cpp
@@ -58,11 +58,15 @@ ContinuousContactManager::UPtr BulletCastSimpleManagerFactory::create(const std:
 
 }  // namespace tesseract_collision::tesseract_collision_bullet
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_PLUGIN(tesseract_collision::tesseract_collision_bullet::BulletDiscreteBVHManagerFactory,
                      BulletDiscreteBVHManagerFactory);
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 TESSERACT_ADD_PLUGIN(tesseract_collision::tesseract_collision_bullet::BulletDiscreteSimpleManagerFactory,
                      BulletDiscreteSimpleManagerFactory);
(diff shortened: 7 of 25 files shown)


tesseract_environment/src/environment.cpp
```diff
@@ -398,6 +399,14 @@ tesseract_kinematics::KinematicGroup::UPtr Environment::getKinematicGroup(const
 
   kinematic_group_cache_[group_name] = std::make_unique<tesseract_kinematics::KinematicGroup>(*kg);
 
+#ifndef NDEBUG
+  if (!tesseract_kinematics::checkKinematics(*kg))
+  {
+    CONSOLE_BRIDGE_logError("Check Kinematics failed. This means that inverse kinematics solution for a pose do not "
+                            "match forward kinematics solution. Did you change the URDF recently?");
+  }
+#endif
```
marip8, line 408: A few things:
1. Groups should be cached after they've passed this check
1. This check should probably be a requirement for every kinematics group that's added regardless of build type since it is fairly critical. Groups won't be added frequently, and retrieving a cached group doesn't require this check to be re-run*, so speed performance shouldn't be affected significantly
1. This PR should probably also update the `checkKinematics` function to remove the redundant checking of all combinations of working frame and TCP frame per [this issue](https://github.com/tesseract-robotics/tesseract_planning/issues/148)

```diff
-  kinematic_group_cache_[group_name] = std::make_unique<tesseract_kinematics::KinematicGroup>(*kg);

+  if (!tesseract_kinematics::checkKinematics(*kg))
+    throw std::runtime_error("Check Kinematics failed. This means that inverse kinematics solution for a pose do not "
+                            "match forward kinematics solution. Did you change the URDF recently?");

+  // Cache the group only after it has passed the check                            
+  kinematic_group_cache_[group_name] = std::make_unique<tesseract_kinematics::KinematicGroup>(*kg);
```

\* I'm not sure what should happen if the structure of the environment changes after a kinematic group has been cached. Is the cached group still valid? Should the kinematics be checked every time the group is requested to make sure everything still works?

marip8 (changes requested):

tesseract_environment/src/environment.cpp
Levi-Armstrong: The cache is cleared after every environment change and state change. I want to avoid this check in release because it will get called at least once for a planning request and for gantry systems this could be a significant amount of time. I think it is valid to only run in debug because it is meant to catch things that should never occur.

Levi-Armstrong: > This PR should probably also update the checkKinematics function to remove the redundant checking of all combinations of working frame and TCP frame per this issue

I do not think this should be removed as it is meant to catch things that should not occur which there is logic in the JointGroup and KinematicGroup which this would catch if something changed along with other unit tests.

marip8: > I want to avoid this check in release because it will get called at least once for a planning request and for gantry systems this could be a significant amount of time

`checkKinematics` doesn't need to check every combination of working frame and TCP frame since those frames are just static offsets from the base and tip frame. At most it should call FK and IK once and check that the `n` IK solutions match the joint values used in the FK call.

In practice, the amount of time spent checking kinematics when the group gets created (if done correctly) will be a negligible amount of the total planning time because groups generally don't get created at high frequency and environments generally don't get modified at high frequency. Even for a gantry mounted robot, the amount of time required for this check is trivial (~1-2s max). Making the check is really valuable though because it guarantees that FK and IK work as expected. I think that's worth checking all the time, not just in debug

> I think it is valid to only run in debug because it is meant to catch things that should never occur.

My argument against asserts is that if the error should never occur, why should we even be checking it? I think the answer to that question is that the error does occur occasionally, and I would prefer to always know about it rather than letting the system ignore it and manifest in different errors or unexpected behavior later.

Levi-Armstrong: > My argument against asserts is that if the error should never occur, why should we even be checking it? I think the answer to that question is that the error does occur occasionally, and I would prefer to always know about it rather than letting the system ignore it and manifest in different errors or unexpected behavior later.

Yes they should not occur but developers make mistakes. This catches issues where logic has changed in the code which broke things. Also adds value when debugging issues because sometime it is much easier to add an assert than write a unit test that may capture all logical paths. One example is critical point so of the code a map is access using `[ ]` instead of `at` due to performance but asserts were added to check that the key exists.

Levi-Armstrong: > In practice, the amount of time spent checking kinematics when the group gets created (if done correctly) will be a negligible amount of the total planning time because groups generally don't get created at high frequency and environments generally don't get modified at high frequency. Even for a gantry mounted robot, the amount of time required for this check is trivial (~1-2s max). Making the check is really valuable though because it guarantees that FK and IK work as expected. I think that's worth checking all the time, not just in debug

This is intended to check that the logic remains correct for any possible kinematic configuration. The only why this can be done is to check all combination because the logic changes if you are using a robot on a positioner or a robot with and external positioner and if logic upstream changes. This was added because there were logical error in the KinematicGroup which were not caught by just check the tip link because the logic in how those static offsets were created.

Levi-Armstrong: This is to catch logical errors not user errors which is the main argument that is should be an assert and not checked as part of the normal flow.

marip8: > The only way this can be done is to check all combination because the logic changes if you are using a robot on a positioner or a robot with and external positioner and if logic upstream changes.

You mean that there are multiple checks you need to make if you are using a special kinematic setup, like a robot with external positioner, right? i.e:
  - Check that FK/IK still work if the target pose is transformed into the manipulator base frame (i.e. "normal" check)
  - Check that FK/IK still work if the target pose is transformed into the positioner tool frame (i.e. special check for robot with external positioner)

In either case, you still only need to make a check with a single working frame and TCP, but the current way to cover both cases generically is to check all combinations?

Levi-Armstrong: The kinematic check related to the KinematicGroup is more involved than just forward and inverse kinematics found in tesseract_kinematics. Those object can be checked using the tesseract_kinematics unit test frame work. This check was intended to exercises the KinematicGroup object for logical errors. For example the kinematics group creates an optimize forward kinematics object and if there is an error in its logic related to the generation of static frames attached to the Working Frame and TCP Frames this would catch those errors. It is not just enough to check one attached link because the logic is different if the child links contains non fix joints versus fixed joints which may be different for each user.

Levi-Armstrong: Now it probably does not need to check all permutation of tip links to working frames but it could be change to check all tip links to the first working_frame and then check the first tip link to all working frames which should capture all possible logical errors that I can think of.

marip8: To summarize:
- The `KinematicGroup` class uses a `tesseract_scene_graph::StateSolver` object to generate all possible working frames for the group [here](https://github.com/tesseract-robotics/tesseract/blob/06d77de7aa0d362b7c1628d8c6daa6283d15a4ce/tesseract_kinematics/core/src/kinematic_group.cpp#L64-L79)
- The `tesseract_scene_graph::StateSolver` has been tested thoroughly enough that we believe that it identifies all possible working frames correctly, regardless of whether the single working frame specified by the inverse kinematics interface is a moving link or static link
- Special kinematics (i.e. robot on positioner (ROP), robot with external positioner (REP)) require a user to specify the forward kinematics of the positioner in the definition of the inverse kinematics of the whole system (e.g. [here for REP](https://github.com/tesseract-robotics/tesseract/blob/06d77de7aa0d362b7c1628d8c6daa6283d15a4ce/tesseract_kinematics/core/include/tesseract_kinematics/core/rep_inv_kin.h#L138) and [here for ROP](https://github.com/tesseract-robotics/tesseract/blob/06d77de7aa0d362b7c1628d8c6daa6283d15a4ce/tesseract_kinematics/core/include/tesseract_kinematics/core/rop_inv_kin.h#L87))
    - **There is a possibility that a user who implements this forward kinematics interface does it incorrectly**
- One purpose of `checkKinematics` is to ensure that the internal "logic" of the solvers was implemented correctly (such as the positioner forward kinematics in the REP and ROP inverse kinematics solvers)

Am I following correctly?

Levi-Armstrong: Yea, there is a lot of different pieces that are involved which I have ran into issue where this function has help identify logical errors. I have been adding unit tests and different kinematics arrangements each time an issue was found. I think mostly the unit tests cover a lot and this check is not need in the normal flow but is good to have in the debug flow in case anything is ever refactored or a logical branch has not been tested in the even an issue is raised this would hopeful trigger in a debug build. 

> The tesseract_scene_graph::StateSolver has been tested thoroughly enough that we believe that it identifies all possible working frames correctly, regardless of whether the single working frame specified by the inverse kinematics interface is a moving link or static link

For the mos part this is true except the Kinematic group uses a different variant which will replace non-fixed joints with a fixed joint at the joint value when the kinematic object is created. I have also been adding unit tests as I have found issues but this check has also been useful in identifying logical errors. Again I think most are now covered by unit tests but still good to have in the debug flow in case the code ever goes through a refactor for example.

> Special kinematics (i.e. robot on positioner (ROP), robot with external positioner (REP)) require a user to specify the forward kinematics of the positioner in the definition of the inverse kinematics of the whole system (e.g. here for REP and here for ROP)

I believe most of this is covered in the unit tests now but the checkKinematics was useful in identify logical errors in these implementation. The implementation should not throw an exception due to user error.

I have been using this not to check for user errors as those should be capture by the forward and inverse kinematics objects them self during construction, but more to ensure to myself that logically the KinematicGroup and JointGroup logically function correctly. Also it will identify errors in kinematic implementation if they fail to handle something on there end.

marip8: Okay. I won't hold this change up any longer, but I have a few more thoughts:

I don't think that validating the logic of important classes should be done with half-measures like assertions. If you don't want to run this check at run-time (which still seems like the best solution, IMO), it seems like it would be better to replace the assertion with a unit test that checks as many known challenging/problematic kinematic setups as you can devise URDFs/SRDFs for. As you discover new edge cases, add the associated URDF/SRDF to the unit test to catch logic errors that might be caused by upstream changes.

I'm still also not convinced that the `checkKinematics` function needs to check more than one combination of working frame and TCP. It seems like the key aspect is to do the check with more than one joint state where all of the joints within the group have changed. I would think this should capture any incorrect behavior with special kinematic setups like REP and ROP

Levi-Armstrong: I updated it to reduce the number of tcp_frame to working_frame checks.

Levi-Armstrong: I believe you are misunderstanding the purpose of asserts. The purpose of asserts are a developers friend to allow them  diagnose issues faster related to pre and post condition through the code. They are not intended to be a half measure but to check logical assumption that  developers has made at points throughout the code. I also, do not believe you should replace asserts with unit test. That assert should always remain related to pre and post conditions and included while writing code because without these it adds significant amount of time tracking down issues. In parallel, unit tests should be written and updated as issues are found. 

I recommend checking out the several section of the Cpp Core Guidelines starting [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i5-state-preconditions-if-any) covering pre and post conditions. One of the examples below. 

``` c++
int area(int height, int width)
{
    Expects(height > 0 && width > 0);            // good
    if (height <= 0 || width <= 0) my_error();   // obscure
    // ...
}
```

This `checkKinematics` is a post condition check where I expect the solution return by inverse kinematics to match the provided pose. As you mention as developers we do a lot of testing in debug and I was avoiding including this directly in the KinematicGroup method that calculates inverse solution as a post condition check because I assume this could bog things down but I did not test it to confirm. Instead this function was used at critical points in the code as a pre conditional test where I expect the KinematicGroup to produce valid inverse solution.

Why this valuable to me as a developer is covered in this example. Say a user/developer posted an issue saying things are not functioning correctly related to motion planning and they had implemented a custom inverse kinematics solver. This has nothing to do with this code base but by quickly building in debug and running an example it would quickly identify where the issues is. Without pre and post conditional checks you could spend a significant amount of time tracking down if the issue is in your code base or somewhere else. 

Another example I like to use is if you access a map using the `[ ]` operator an assumption has been made that an entry for the provide key should exist and as a result there should always be a prepositional assert check that the key in fact does exist. There is only a upside when include pre and post conditional assert because they are optimized out of the release code.

PR #683, opened 2021-12-15
Add docker creation for major.minor

Levi-Armstrong:
Currently a docker is create for major.minor.patch but this also adds one with just major.minor which will be used for other CI pipelines to avoid having to frequently update other CI pipelines for patch versions.

### DIFF
--- a/.github/workflows/bionic_build.yml
+++ b/.github/workflows/bionic_build.yml
@@ -69,7 +69,8 @@ jobs:
             suffix=
           tags: |
             type=ref,event=branch,prefix=${{ env.ROS_DISTRO }}-
-            type=ref,event=tag,prefix=${{ env.ROS_DISTRO }}-
+            type=semver,pattern={{version}},prefix=${{ env.ROS_DISTRO }}-
+            type=semver,pattern={{major}}.{{minor}},prefix=${{ env.ROS_DISTRO }}-
 
       - name: Set build type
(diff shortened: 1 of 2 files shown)


marip8 (approved):
I think this is good and will make deployment and CI easier. We should be careful to follow semver rules closely though going forward; otherwise these tags and images won't have much meaning.

Levi-Armstrong:
@marip8 It looks like there is an issue with the push. I noticed in the documentation they using the following to push. Did you try this originally and it did not work?

```
-
        name: Build and push
        uses: docker/build-push-action@v2
        with:
          context: .
          push: ${{ github.event_name != 'pull_request' }}
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
```

marip8:
I was under the impression that the `build-push-action` actually builds a Docker image and then pushes that built image to some repository. I didn't think you could skip the build step, but I didn't look into it very deeply since the push aspect is a one-line call: `docker push ...`. What's the issue with the push?

Levi-Armstrong:
Yea you are correct.

marip8:
If you're wondering why release 0.7.3 didn't produce a docker image for the Bionic and Focal builds, it's because the docker meta-information extractor did not come up with any tag names based on the new pep440 rules

image

PR #685, opened 2021-12-15
Fix docker push

Levi-Armstrong:
@marip8


Levi-Armstrong:
@marip8 The issue I believe is that this is `${{ steps.meta.outputs.tags }}` is now a list and running docker push on this variable it fails. I made a change in this PR but not sure how best to test it without merging. How did you originally test this?

marip8:
I see. I've never had more than one tag be active at a time, so I hadn't run into this yet. I tested by opening a PR against my fork of the repository and merging it into my fork's master branch

PR #693, opened 2022-01-04
CPack Update

marip8:
This PR fixes a few issues with the CPack configurations:
## Debian dependencies
Currently the dependencies of a package are passed into the `tesseract_cpack` macro as a list of strings. `cmake` concatenates these entries into a semi-colon separated string. This is problematic for debians because they expect a comma-separated list of dependencies. I think NuGet packages have the same requirement, but I was not able to confirm. The solution was to replace the semi-colons in `${ARG_<OS>_DEPENDS}` with commas

## Debian architecture name
The debians I initially created would not install because the specified architecture `x86_64` did not match the system architecture `amd64`. The specified architecture currently comes from the `CMAKE_SYSTEM_PROCESSOR` variable, which returns `x86_64` in my case. Per [this document](https://wiki.debian.org/ArchitectureSpecificsMemo), `x86_64` is not actually a valid debian architecture; the corresponding architecture should actually be `amd64`. The solution was to change references from `x86_64` to `amd64` for Linux builds. I'm not sure if Windows NuGet packages are affected the same way, so I didn't change anything there


Levi-Armstrong (approved):

Levi-Armstrong (approved):

Levi-Armstrong:
@marip8 Is it okay to merge this is CI passes?

marip8:
Yes, I've run it locally and been able to produce debians

PR #694, opened 2022-01-05
Fixed environment benchmark

marip8:


Levi-Armstrong:
@marip8 Are you able to trigger the benchmark build on your repo to show it passes? If not and you tested locally feel free to merge.

Levi-Armstrong (approved):

marip8:
I built the benchmarks locally and ran the one that I changed, so I'm fairly confident this fixes the issue

marip8:
It would probably be a good idea to separate the building of benchmarks from the running of them, like we do for unit tests. Then we could build the benchmarks as a part of CI, but only run them when the master branch is changed. That way they will stay up-to-date

PR #695, opened 2022-01-10
Update Kinematics Cache To Include IK Solver

marrts:
Previously if you try and load multiple IK solvers for a manipulator the cache would automatically load in the first used IK solver for the given manipulator group


tesseract_environment/src/environment.cpp
```diff
@@ -373,14 +373,15 @@ tesseract_kinematics::KinematicGroup::UPtr Environment::getKinematicGroup(const
   std::shared_lock<std::shared_mutex> lock(mutex_);
 
   std::unique_lock<std::shared_mutex> cache_lock(kinematic_group_cache_mutex_);
-  auto it = kinematic_group_cache_.find(group_name);
+  std::string group_plus_ik_name = group_name + "_" + ik_solver_name;
```
Levi-Armstrong, line 376: I would update the cache to use a `std::pair<std::string, std::string>` as the key.
``` c++
auto key = std::make_pair(group_name, ik_solver_name);
```

tesseract_environment/src/environment.cpp
```diff
@@ -373,14 +373,15 @@ tesseract_kinematics::KinematicGroup::UPtr Environment::getKinematicGroup(const
...
+  auto it = kinematic_group_cache_.find(group_plus_ik_name);
   if (it != kinematic_group_cache_.end())
   {
-    CONSOLE_BRIDGE_logDebug("Environment, getKinematicGroup(%s) cache hit!", group_name.c_str());
+    CONSOLE_BRIDGE_logDebug("Environment, getKinematicGroup(%s) cache hit!", group_plus_ik_name.c_str());
     return std::make_unique<tesseract_kinematics::KinematicGroup>(*it->second);
   }
 
-  CONSOLE_BRIDGE_logDebug("Environment, getKinematicGroup(%s) cache miss!", group_name.c_str());
+  CONSOLE_BRIDGE_logDebug("Environment, getKinematicGroup(%s) cache miss!", group_plus_ik_name.c_str());
```
Levi-Armstrong, line 384: Update these to the following.

``` c++
CONSOLE_BRIDGE_logDebug("Environment, getKinematicGroup(%s, %s) cache miss!", group_name.c_str(), ik_solver_name.c_str());
```

Levi-Armstrong (changes requested):
Good catch!

tesseract_environment/src/environment.cpp
marrts: By default you can't use a std::pair<> as the key to an unordered map. You need to provide a hashing operation for the pair, so in the end I set it up so what's happening under the hood is functionally the same as what you see here.

tesseract_environment/include/tesseract_environment/environment.h
```diff
@@ -531,11 +531,20 @@ class Environment
...
   mutable std::shared_mutex joint_group_cache_mutex_;
 
+  struct kinematic_group_pair_hash
+  {
+      std::size_t operator () (const std::pair<std::string,std::string> &p) const
+      {
+          std::string combined = p.first + p.second;
+          return std::hash<std::string>{}(combined);
+      }
+  };
```
Levi-Armstrong, line 541: This should not be required since;  did it complain without it?

marrts: Yes, because there is no default hashing for a std::pair, which is required for an unordered_map

tesseract_environment/src/environment.cpp
Levi-Armstrong: I see, I would switch the cache form an `std::unordered_map`  to a `std::map` then you will not need the hash function.

tesseract_environment/include/tesseract_environment/environment.h
Levi-Armstrong: See comment above about switching cache type.

tesseract_environment/src/environment.cpp
marrts: Do you want me to switch all the caches to that for sake of consistency?

Levi-Armstrong: Nah, those should remain the same because the key is actually a string. From what I understand unordered_map is better performing when the key is a string. In this case we have switched to std::pair as a key so std::map is ok.

Levi-Armstrong (approved):

PR #698, opened 2022-01-15
Fix check trajectory which should return a vector same length as trajectory

Levi-Armstrong:
If using discrete checkTrajectory should return a vector the same length as the input trajectory. If using continuous it should return a vector with a length 1 minus the trajectory length.


tesseract_environment/include/tesseract_environment/utils.h
```diff
@@ -53,64 +53,75 @@ void getActiveLinkNamesRecursive(std::vector<std::string>& active_links,
 
 /**
  * @brief Should perform a continuous collision check between two states configuring the manager with the config
- * @param contacts A vector of vector of ContactMap where each index corresponds to a timestep
  * @param manager A continuous contact manager
  * @param state0 First environment state
  * @param state1 Second environment state
  * @param config CollisionCheckConfig used to specify collision check settings
- * @return True if collision was found, otherwise false.
+ * @return Return the contact results map. If empty not contacts were found
```
mpowelson, line 60: typo `no contacts`

mpowelson: Same below

mpowelson (approved):

PR #702, opened 2022-01-24
Remove unneeded boost bind include

jdlangs:
Not needed since C++11 and this header puts placeholder objects in the global namespace on system-installed Boost versions


Levi-Armstrong (approved):

Levi-Armstrong:
I will merge once CI passes.

PR #703, opened 2022-01-25
Add ability to provide calibration information in the SRDF

Levi-Armstrong:
The current calibration config support defining a joint origin in a yaml config file which gets loaded through the srdf. In the environment it will apply these as a `ChangeJointOriginCommand`.

``` yaml
calibration:
  joints:
    joint_1:
      position:
        x: 1
        y: 2
        z: 3
      orientation:
        x: 0
        y: 0
        z: 0
        w: 1
    joint_2:
      position:
        x: 4
        y: 5
        z: 6
      orientation:
        x: 0
        y: 0
        z: 0
        w: 1
```

srdf element:
``` xml
<calibration_config filename="calibration_config.yaml"/>
```

### DIFF
--- a/tesseract_common/include/tesseract_common/types.h
+++ b/tesseract_common/include/tesseract_common/types.h
@@ -160,6 +160,29 @@ struct ContactManagersPluginInfo
   // Yaml Config key
   static inline const std::string CONFIG_KEY{ "contact_manager_plugins" };
 };
+
+/** @brief The CalibrationInfo struct */
+struct CalibrationInfo
+{
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  CalibrationInfo() = default;
+
+  /** @brief The joint origin calibration information */
+  tesseract_common::TransformMap joints;
+
+  /** @brief Insert the content of an other CalibrationInfo */
+  void insert(const CalibrationInfo& other);
+
+  /** @brief Clear the contents */
+  void clear();
+
+  /** @brief Check if structure is empty */
+  bool empty() const;
+
+  // Yaml Config key
+  static inline const std::string CONFIG_KEY{ "calibration" };
+};
 }  // namespace tesseract_common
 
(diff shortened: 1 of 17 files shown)


tesseract_common/include/tesseract_common/types.h
```diff
@@ -160,6 +160,29 @@ struct ContactManagersPluginInfo
...
+
+/** @brief The CalibrationInfo struct */
+struct CalibrationInfo
+{
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  CalibrationInfo() = default;
+
+  /** @brief The joint origin calibration information */
+  tesseract_common::TransformMap joints;
```
mpowelson, line 172: Can we get more documentation on what this does? Does it replace the existing map or get applied on top of it?

mpowelson (commented):
Looks reasonable. Are you updating the docs too? It looks like you can specify either a file in the urdf or include the information directly in the SRDF? Is there an easy way to load this after the fact if you have to pull it from somewhere?

Levi-Armstrong:
> Are you updating the docs too?

Yep, see this [PR](https://github.com/tesseract-robotics/tesseract_docs/pull/13/commits/2f7b1f3baa1385e5e8af0073083c4d19c6133bd7)

> It looks like you can specify either a file in the urdf or include the information directly in the SRDF?

You can only provide it in the SRDF. The URDF does have a calibration tag but it is currently not supported. I never looked into how this information is used.

> Is there an easy way to load this after the fact if you have to pull it from somewhere?

Not exactly. I could create its own command. It currently just loops over the map and applies a ChangeJointOriginCommand for each entry.

tesseract_common/include/tesseract_common/types.h
Levi-Armstrong: For every entry it applies a `ChangeJointOriginCommand`, so it is replacing the origin of the joint with what was provided.

PR #708, opened 2022-02-04
A few fixes that were needed for Windows

jdlangs:
The three commits here:

- Bullet extras are not easily obtained on Windows so this makes building the HACD convex decomposition library optional
- Paths to bullet are specified as `${BULLET_ROOT_DIR}/${BULLET_INCLUDE_DIRS}`, e.g., but the include dirs already contains absolute paths. ~Probably this wasn't noticed on Ubuntu since it would already be looking in the system install dirs.~ This is a result of packaging for vcpkg and is different from the version packaged for Ubuntu.
- Had a compile error since a header used but didn't include `std::string`

### DIFF
--- a/tesseract_collision/bullet/CMakeLists.txt
+++ b/tesseract_collision/bullet/CMakeLists.txt
@@ -12,14 +12,30 @@ if(NOT
     WARNING "Bullet does not appear to be build with double precision, current definitions: ${BULLET_DEFINITIONS}")
 endif()
 
-if(WIN32) # This was required for windows to find things in tesseract_ext before the system which is an issue in ros-win
-  include_directories(BEFORE "${BULLET_ROOT_DIR}/${BULLET_INCLUDE_DIRS}")
-  link_directories(BEFORE "${BULLET_ROOT_DIR}/${BULLET_LIBRARY_DIRS}")
+# Some Bullet installations (vcpkg) use absolute paths instead of relative to BULLET_ROOT_DIR in the CMake vars
+set(BULLET_INCLUDE_DIRS_ABS "")
+set(BULLET_LIBRARY_DIRS_ABS "")
+if(NOT IS_ABSOLUTE "${BULLET_INCLUDE_DIR}")
+  foreach(dir IN LISTS BULLET_INCLUDE_DIRS)
+    list(APPEND BULLET_INCLUDE_DIRS_ABS "${BULLET_ROOT_DIR}/${dir}")
+  endforeach()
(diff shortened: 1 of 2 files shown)


jdlangs:
I'll work through what's going wrong with the build later tonight.

jdlangs:
Ok, I didn't realize the Ubuntu installed version *does* use a root dir prefix and relative paths for the other variables. After looking into, it seems the vcpkg version of Bullet [rewrites](https://github.com/microsoft/vcpkg/blob/master/ports/bullet3/cmake-fix.patch#L1-L36) the CMake config file, which I assume is where the absolute paths are coming from. I can redo the commit to take into account both situations, unless there's another suggestion.

Levi-Armstrong:
> Ok, I didn't realize the Ubuntu installed version _does_ use a root dir prefix and relative paths for the other variables. After looking into, it seems the vcpkg version of Bullet [rewrites](https://github.com/microsoft/vcpkg/blob/master/ports/bullet3/cmake-fix.patch#L1-L36) the CMake config file, which I assume is where the absolute paths are coming from. I can redo the commit to take into account both situations, unless there's another suggestion.

In the case where they are using absloute path is the BULLET_ROOT_DIR empty? If so you could just do a check at the top which defines a local variable for the path based on if it is empty or not.

jdlangs:
It's not empty but I can still use that strategy. I see CMake has an `IS_ABSOLUTE` check on paths that can be used first.

Levi-Armstrong:
It looks good to merge once the cmake format is resolved.

jdlangs:
I don't see anything in the log that shows what went wrong.

Levi-Armstrong:
> I don't see anything in the log that shows what went wrong.

I would just run `./.run-cmake-formate` script and it will take care of the formatting issue.

jdlangs:
> I would just run `./.run-cmake-formate` script and it will take care of the formatting issue.

Will do, but that probably shouldn't be blocking contributions if people can't see what's wrong.

Levi-Armstrong:
> > I would just run `./.run-cmake-formate` script and it will take care of the formatting issue.
> 
> Will do, but that probably shouldn't be blocking contributions if people can't see what's wrong.

I have updated the [CI](https://github.com/tesseract-robotics/tesseract/pull/709) to output and error message stating to run the script and also output the diff if you want to manually make the change.

jdlangs:
Awesome, thanks for adding that!

I've adjusted the CMake formatting so we should be good to go.

PR #710, opened 2022-02-09
Sort ACM alphabetically when writing

mpowelson:
Sorts the ACM alphabetically when writing it. This is motivated by [this](https://github.com/tesseract-robotics/tesseract_ignition/issues/8) issue. With this change the allowed collisions in the generated SRDF will be sorted alphabetically first by the first link and then by the second link. This does not address that I assume it is possible that which link is the first/second link in the pair could be flipped. We may want to address that in the future too if we want generated SRDFs to be the same run to run. That could be done here, but it might make more sense to do that when we create the pairs.

### DIFF
--- a/tesseract_srdf/include/tesseract_srdf/utils.h
+++ b/tesseract_srdf/include/tesseract_srdf/utils.h
@@ -26,6 +26,9 @@
 #ifndef TESSERACT_SRDF_UTILS_H
 #define TESSERACT_SRDF_UTILS_H
 
+#include <functional>
+#include <tesseract_common/allowed_collision_matrix.h>
+#include <tesseract_common/types.h>
 #include <tesseract_scene_graph/graph.h>
 #include <tesseract_srdf/srdf_model.h>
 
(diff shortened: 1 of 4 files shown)


Levi-Armstrong:
@mpowelson One question, were you not able to use `const &` instead of `const *` ?

mpowelson:
I didn't use references because I wanted to make a vector of them.

Levi-Armstrong:
> I didn't use references because I wanted to make a vector of them.

Can you not use `std::vector<std::reference_wrapper<tesseract_common::LinkNamesPair>>`?

mpowelson:
> > I didn't use references because I wanted to make a vector of them.
> 
> Can you not use `std::vector<std::reference_wrapper<tesseract_common::LinkNamesPair>>`?

I think either one should functionally work here. Is that preferred? Interestingly the c++ core guidelines don't mention reference_wrapper vs ptr usage.

Levi-Armstrong:
I think this is the relevant[ cpp guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es65-dont-dereference-an-invalid-pointer) which recommends using references if null pointers are not possible. I personally avoid raw pointers if it is possible in the public API.

mpowelson:
I'm fine with changing to reference wrappers, but they also seem to push [not_null](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f23-use-a-not_nullt-to-indicate-that-null-is-not-a-valid-value).

Levi-Armstrong:
> I'm fine with changing to reference wrappers, but they also seem to push [not_null](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f23-use-a-not_nullt-to-indicate-that-null-is-not-a-valid-value).

Interesting. Is the `not_null` part of the std or their GSI tool?

Levi-Armstrong (approved):

mpowelson:
It's just gsl. Reference wrapper is probably better in this case. It sounds like not_null is mostly there for historical reasons.

PR #714, opened 2022-02-18
Modifying Boost_VERSION check to use semver

KStaub:
This does not change the comment above the if() statement. I tested with BOOST_VERSION and Boost_VERSION and did not see any different. The if() block is not entered with boost-1.71.0 with the current version check.


Levi-Armstrong (approved):

johnwason:
This fix is not going to work with older versions of cmake. FindBoost returns Boost_VERSION in macro format up until cmake version 3.15, then changes to semver and adds Boost_VERSION_MACRO to return the old format. Ubuntu Bionic uses cmake 3.10, so it is going to cause a problem.

There is a policy to request the old format: https://cmake.org/cmake/help/v3.23/policy/CMP0093.html#policy:CMP0093

However I don't know if this will work since newer versions of Boost provide cmake package files so FindBoost isn't used. I don't know if the Boost cmake files use the policy.

Levi-Armstrong:
@KStaub What version of cmake were you using when you had issues and required the change?

KStaub:
I was using cmake-3.16.3. I see that you've merged a change that should resolve it so thank you and @johnwason!!

PR #724, opened 2022-03-18
Fix incorrect linking to console_bridge

jdlangs:
This popped up when building against a ROS2 source build. Not sure why it wasn't apparent before.


Levi-Armstrong (approved):

Levi-Armstrong:
Good catch. On earlier releases of console_bridge the export target was `console_bridge` then they changed it to `console_bridge::console_bridge`.

PR #728, opened 2022-03-25
Do not pull cache when building docker images

Levi-Armstrong:


Levi-Armstrong:
The cache was not the issue for some reason on the release it did not trigger the docker build for 0.8.6.

PR #729, opened 2022-03-25
Add ctest calls to windows_noetic_build.yml

johnwason:
It appears that tests were not being run correctly in windows_noetic_build.yml. I explicitly ran ctest on the packages and got a lot of unit test failures. This PR adds explicit ctest calls to the workflow script.


johnwason:
It doesn't look like the tests are being run correctly. I don't see any log output. I will look into it.

PR #732, opened 2022-03-28
Minor updates

Levi-Armstrong:


tesseract_environment/test/tesseract_environment_unit.cpp
```diff
@@ -51,7 +51,7 @@ std::string locateResource(const std::string& url)
 Eigen::Isometry3d tcpCallback(const tesseract_common::ManipulatorInfo& mi)
 {
   const std::string& tcp_offset_name = std::get<0>(mi.tcp_offset);
-  if (tcp_offset_name == "laser")
+  if (tcp_offset_name == "laser_callback")
```
mpowelson, line 54: Is this supposed to be changed?

tesseract_environment/src/environment.cpp
```diff
@@ -917,6 +942,14 @@ void Environment::environmentChanged()
     group_joint_names_cache_.clear();
   }
 
+  // Call the event callbacks
+  if (!event_cb_.empty())
+  {
+    CommandAppliedEvent event(commands_, revision_);
+    for (const auto& cb : event_cb_)
+      cb(event);
```
mpowelson, line 950: So all callbacks get called with this even regardless of event type? It seems you could also do this by storing them in a map by event type and only calling the ones for this event type.

mpowelson (approved):

tesseract_environment/src/environment.cpp
Levi-Armstrong: I was following a similar pattern to how [Qt event filters](https://doc.qt.io/qt-5/eventsandfilters.html#event-filters) work. They then just have a switch statement in there code to handle the events they care about.

Levi-Armstrong: I think this list should be small and have little impact but may need to revisit if that turns out to not be the case.

Issue #706, opened 2022-01-28
Add ability to search for all plugins of specific type


marip8:
Currently the each `tesseract` plugin implementation exports a unique symbol which generally corresponds to the name of the implementation class. A specific plugin is loaded by specifying its unique symbol name. This makes it difficult for an application that doesn't know these unique names (such as a GUI) to query for all available plugins (e.g. inverse kinematics, contact managers, visualization, planners, etc.) that match a particular interface. 

It looks like the `boost` solution for this is to use the `BOOST_DLL_ALIAS_SECTIONED` macro with a "section" name when exporting the plugin. A common section name serves as a namespace for multiple plugins exported with different symbol names.

One solution to this problem would be to update the `TESSERACT_ADD_PLUGIN` macro from 

https://github.com/tesseract-robotics/tesseract/blob/e9d6f064f14c84bb14c65c2ebfe529a325133a80/tesseract_common/include/tesseract_common/class_loader.h#L93-L95

to: 

```c++
#define TESSERACT_ADD_PLUGIN(DERIVED_CLASS, ALIAS, SECTION) \ 
  BOOST_DLL_ALIAS_SECTIONED(DERIVED_CLASS, ALIAS, SECTION)
```
where `SECTION` could be the name of the abstract base class (e.g. `ContactManager`, `InverseKinematics`, etc.) or something similar. A method could then be added to the plugin loader class to list all symbol names under a given section name per [this tutorial](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_dll/tutorial.html#boost_dll.tutorial.querying_libraries_for_symbols). 

An alternative would be to export all plugin implementations with the same alias (which would be something like the abstract base class name) and build them in separate libraries. `boost` provides an example of that [here](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_dll/tutorial.html#boost_dll.tutorial.symbol_shadowing_problem__linux_). However, this approach would require us to split up the compilation of several existing plugins that are currently in the same library (e.g. contact managers and possibly others) because a library presumably cannot have multiple definitions of the same symbol.

Levi-Armstrong:
I like the section approach.

marip8:
Should be a pretty minimally invasive change. I might be able to tackle this next week.

Levi-Armstrong:
This has been addressed by PR #741

PR #741, opened 2022-04-07
Updated plugin capability to support sections

Levi-Armstrong:
Address issue #706

### DIFF
--- a/tesseract_common/include/tesseract_common/class_loader.h
+++ b/tesseract_common/include/tesseract_common/class_loader.h
@@ -30,6 +30,7 @@
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <memory>
 #include <string>
+#include <vector>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
 namespace tesseract_common
@@ -40,7 +41,7 @@ namespace tesseract_common
  * suffix based on the OS.
  *
  * The plugin must be exported using the macro TESSERACT_ADD_PLUGIN.
- * In the example below, the first parameter is the derived object and the second is the assinged symbol name which is
+ * In the example below, the first parameter is the derived object and the second is the assigned symbol name which is
  * used for looding Example: TESSERACT_ADD_PLUGIN(my_namespace::MyPlugin, plugin)
  *
  *   auto p = ClassLoader::createSharedInstance<my_namespace::MyPluginBase>("my_plugin", "plugin");
@@ -74,6 +75,27 @@ struct ClassLoader
                                       const std::string& library_directory = "");
 
   /**
+   * @brief Get a list of available symbols under the provided section
+   * @param section The section to search for available symbols
+   * @param library_name The library name to load which does not include the prefix 'lib' or suffix '.so'
+   * @param library_directory The library directory, if empty it will enable search system directories
+   * @return A list of symbols if they exist.
+   */
+  static inline std::vector<std::string> getAvailableSymbols(const std::string& section,
+                                                             const std::string& library_name,
+                                                             const std::string& library_directory = "");
+
+  /**
+   * @brief Get a list of available sections
+   * @param library_name The library name to load which does not include the prefix 'lib' or suffix '.so'
+   * @param library_directory The library directory, if empty it will enable search system directories
+   * @return A list of sections if they exist.
+   */
+  static inline std::vector<std::string> getAvailableSections(const std::string& library_name,
+                                                              const std::string& library_directory = "",
+                                                              bool include_hidden = false);
+
+  /**
    * @brief Give library name without prefix and suffix it will return the library name with the prefix and suffix
    *
    * * For instance, for a library_name like "boost" it returns :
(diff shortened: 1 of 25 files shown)


tesseract_common/include/tesseract_common/class_loader.h
```diff
@@ -90,10 +112,31 @@ struct ClassLoader
...
 
-#define TESSERACT_ADD_PLUGIN(DERIVED_CLASS, ALIAS)                                                                     \
+// clang-format off
+#define TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, SECTION)                                                  \
   extern "C" BOOST_SYMBOL_EXPORT DERIVED_CLASS ALIAS;                                                                  \
+  BOOST_DLL_SECTION(SECTION, read) BOOST_DLL_SELECTANY                                                                 \
   DERIVED_CLASS ALIAS;
 
+#define TESSERACT_ADD_PLUGIN(DERIVED_CLASS, ALIAS)                                                                     \
+  TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, boostdll)
```
marip8, line 122: What do you anticipate using this for? It might be confusing to user creating a plugin if they find this macro before finding a macro for the specific plugin type they actually care about. If it doesn't have a direct use case now (other than testing) you might consider removing it

tesseract_common/include/tesseract_common/class_loader.h
```diff
@@ -90,10 +112,31 @@ struct ClassLoader
...
+  TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, InvKin)
+
+#define TESSERACT_ADD_DISCRETE_MANAGER_PLUGIN(DERIVED_CLASS, ALIAS)                                                    \
+  TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, DiscColl)
+
+#define TESSERACT_ADD_CONTINUOUS_MANAGER_PLUGIN(DERIVED_CLASS, ALIAS)                                                  \
+  TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, ContColl)
+
+#define TESSERACT_ADD_VISUALIZATION_PLUGIN(DERIVED_CLASS, ALIAS)                                                       \
+  TESSERACT_ADD_PLUGIN_SECTIONED(DERIVED_CLASS, ALIAS, Plotter)
```
marip8, line 137: It might make more sense to define these macros in the respective packages where their plugin interface is defined. If I were making a new plugin, my first thought would be to look in the package relating to that plugin type to get this macro.

tesseract_common/include/tesseract_common/plugin_loader.h
```diff
@@ -90,6 +91,19 @@ class PluginLoader
...
+   * @param section The section name to get all available plugins
+   * @return A list of available plugins under the provided section
+   */
+  inline std::vector<std::string> getAvailablePlugins(const std::string& section) const;
+
+  /**
+   * @brief Get the available sections within the provided search libraries
+   * @return A list of available sections
+   */
+  inline std::vector<std::string> getAvailableSections(bool include_hidden = false) const;
```
marip8, line 105: I think this method has its uses, but I think a more common use case is to search for plugins by interface type. It's much easier to programmatically search for plugins of a specific type than to sift through a bunch of nondescript strings in a vector (i.e. `ContColl`) to find the section name (which isn't really exposed anywhere) for the plugin type I care about. 

I think my preferred approach would be something like `plugin_loader.getAvailablePlugins<tesseract_kinematics::InverseKinematicsPlugin>()`.  The the plugin loader could implement this by accessing a static variable in the plugin interface class with the section name. Since the section names have to be less than 8 bytes and aren't really descriptive, I think it would make sense to make it a private static variable in the class; users don't really ever need to know or see that section name.

Levi-Armstrong: Currently that is accomplished by the `getAvailablePlugins(const std::string& section)`. Right now you can do the following.

``` c++
std::vector<std::string> plugins = plugin_loader.getAvailablePlugins("InvKin");
auto plugin = plugin_loader.instantiate<tesseract_kinematics::InverseKinematicsPlugin>(plugins[0]);
```

> The the plugin loader could implement this by accessing a static variable in the plugin interface class with the section name. Since the section names have to be less than 8 bytes and aren't really descriptive, I think it would make sense to make it a private static variable in the class; users don't really ever need to know or see that section name.

That would require that the plugin loader class is aware of each plugin type which I don't think we want to do. Am I missing something?

tesseract_common/include/tesseract_common/class_loader.h
Levi-Armstrong: Not sure just following the boost pattern of available macros.

tesseract_common/include/tesseract_common/plugin_loader.h
marip8: > `std::vector<std::string> plugins = plugin_loader.getAvailablePlugins("InvKin");
auto plugin = plugin_loader.instantiate<tesseract_kinematics::InverseKinematicsPlugin>(plugins[0]);`

This is more or less what I *don't* want to do because the section names are pretty nondescript and not exposed to the user (currently at least). 

To do what I'm describing I think you would need to have a simple base class for plugins that forces them to provide a section name.

```c++
class Plugin
{
protected:
  std::string getSectionName() = 0;
}

class InvKinFactory : public Plugin
{
...
protected:
  std::string getSectionName() { return "InvKin"; }
  friend class PluginLoader;
}

class PluginLoader
{
...
<template typename PluginT>
std::vector<std::string> getAvailablePlugins()
{
  return class_loader_.getAvailablePlugins(..., T::getSectionName());
}
...
}
```

An easier solution would also be to make the section name a public static variable in the plugin class. Then you could do something like:
```c++
std::vector<std::string> plugins = plugin_loader.getAvailablePlugins(InvKinFactory::SECTION_NAME);
```
That would work fine too, it just seems cleaner to hide the section names from the user completely since they're only used internally

tesseract_common/include/tesseract_common/class_loader.h
marip8: I don't think this is needed then. The idea is to be able to search for plugins of specific types; the value comes from the other macro definitions that are specific to `tesseract` classes not this generic one

tesseract_common/include/tesseract_common/class_loader.h
Levi-Armstrong: Done

tesseract_common/include/tesseract_common/plugin_loader.h
marip8: I haven't thought through it completely, but the base class may not actually be required. They may just need a static variable and friend class declaration

tesseract_common/include/tesseract_common/class_loader.h
Levi-Armstrong: I would like to leave this since I plan to break this out into its own package separated from Tesseract to allow use by other packages. I think this could be used by some of the other ros agnostic packages.

tesseract_common/include/tesseract_common/plugin_loader.h
Levi-Armstrong: @marip8 Take a look at the latest commit adding the protected SECTION_NAME and added the templated method to `PluginLoader` which calls the `getAvailablePlugins` with the SECTION_NAME under the hood.

Levi-Armstrong:
@marip8 If CI passes you good with it being merged?

marip8 (approved):
Looks good!

PR #746, opened 2022-04-15
Fix invalid iterator in bullet_cast_simple_manager

johnwason:
`cows_` in BulletCastSimpleManager sometimes uses `cast_cow` in one of the vectors instead of the uncast `cow`. This mismatch was resulting in a segfault on the lines changed. Search by name directly to avoid this problem.

### DIFF
--- a/tesseract_collision/bullet/src/bullet_cast_simple_manager.cpp
+++ b/tesseract_collision/bullet/src/bullet_cast_simple_manager.cpp
@@ -134,9 +134,7 @@ bool BulletCastSimpleManager::removeCollisionObject(const std::string& name)
   auto it = link2cow_.find(name);
   if (it != link2cow_.end())
   {
-    cows_.erase(std::find_if(cows_.begin(), cows_.end(), [&name](const auto& p) {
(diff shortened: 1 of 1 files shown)


tesseract_collision/bullet/src/bullet_cast_simple_manager.cpp
```diff
@@ -134,9 +134,7 @@ bool BulletCastSimpleManager::removeCollisionObject(const std::string& name)
   auto it = link2cow_.find(name);
   if (it != link2cow_.end())
   {
-    cows_.erase(std::find_if(cows_.begin(), cows_.end(), [&name](const auto& p) {
-      return p->getName() == name;
-    }));
```
Levi-Armstrong: It is late, I am not seeing difference between the two. Can you point out the issue?

johnwason: This commit is just for the clang-format. The full PR changes that from a pointer comparison to a string lambda comparison.

Levi-Armstrong (approved):

PR #751, opened 2022-04-21
Windows fixes with passing unit tests

johnwason:
This PR contains fixes and passing unit tests on Windows. Most of them are pretty minor changes to unit tests. The only major issues discovered is with SceneGraph. It was being copied in a few places, and this lead to invalid pointers stored in the link and joint maps. The problem was not showing on Linux because gcc was return-value optimizing away the copy operation, but MSVC was not. SceneGraph has the copy constructor deleted, but for some reason it was still allowed to be copied. I tried adding boost::noncopyable, but it was being copied in a few places. This issue should be fixed, but for now it doesn't appear to be causing any problems.


tesseract_state_solver/src/ofkt_state_solver.cpp
```diff
@@ -879,7 +879,7 @@ bool OFKTStateSolver::initHelper(const tesseract_scene_graph::SceneGraph& scene_
 void OFKTStateSolver::moveLinkHelper(std::vector<JointLimits::ConstPtr>& new_joint_limits, const Joint& joint)
 {
   auto* old_node = link_map_[joint.child_link_name];
-  const std::string& old_joint_name = old_node->getJointName();
+  const std::string old_joint_name = old_node->getJointName();
```
Levi-Armstrong, line 882: Was this causing an issue?

tesseract_scene_graph/src/graph.cpp
```diff
@@ -403,7 +403,10 @@ bool SceneGraph::removeJoint(const std::string& name, bool recursive)
   else
   {
     if (getInboundJoints(found->second.first->child_link_name).size() == 1)
-      removeLink(found->second.first->child_link_name, true);
+    {
+      std::string child_link_name = found->second.first->child_link_name;
+      removeLink(child_link_name, true);
```
Levi-Armstrong, line 408: What was the issue here?

tesseract_scene_graph/include/tesseract_scene_graph/graph.h
```diff
@@ -130,7 +130,9 @@ struct ShortestPath
 
 class SceneGraph
 #ifndef SWIG
-  : public Graph
+  : public Graph  //, public boost::noncopyable This data structure
+// should be declared noncopyable, but serialization currently
+// is creating copies
```
Levi-Armstrong, line 135: Should this be removed?

tesseract_state_solver/src/ofkt_state_solver.cpp
johnwason: `old_joint_name`  is being used after `old_node` is modified/deleted. I am not sure how it was working before.

tesseract_scene_graph/src/graph.cpp
johnwason: The reference is modified inside of `removeLink()`. Again I am not sure how it was working before.

tesseract_state_solver/src/ofkt_state_solver.cpp
Levi-Armstrong: Oh, I see the issue right after this the node is removed which invalidates the reference.

tesseract_scene_graph/include/tesseract_scene_graph/graph.h
johnwason: The structure should be declared `boost::noncopyable` at some point, but not in this PR. An issue should be created to fix this at some point.

Levi-Armstrong: Yea, it is odd that the copy destructor is deleted but on windows it is still coping? I see where you were switched to using out parameter but those should have been moving the object and not copying it correct?

Levi-Armstrong: Is its that we are using windows compiler 2019 which may not have a full implementation of c++17 standard so it is not correctly moving the object. Is it possible to upgrade to 2020? Also did you by chance try adding an explicit std::move before switching to using an out parameter?

johnwason: I tried adding `boost::noncopyable` to `SceneGraph` on Linux and am seeing the same compiler errors:

```
tesseract_scene_graph_serialization_unit.cpp:212:10: error: use of deleted function ‘tesseract_scene_graph::SceneGraph::SceneGraph(const tesseract_scene_graph::SceneGraph&)’
  212 |   return g;
```

I think you still need to have the object be copyable to use copy elision, since it may still copy in certain situations.

johnwason: I tried defining a dummy move constructor after adding `boost::noncopyable`, and it compiles. It looks like the default move constructor isn't being generated correctly.

johnwason: It compiles using GCC, I still have to test MSVC with an explicit move constructor.

johnwason: No, I am seeing the same errors with an explicit move constructor that I was seeing earlier. I think SceneGraph needs to be allocated on the heap, or used as a reference.

Levi-Armstrong: Then should we just add the boost::noncopyable or does that cause issues other places?

johnwason: I tried adding rebuilding `link_map_` and `joint_map_` to the move constructor, added `boost::noncopyable`, and it passed the scene graph unit tests. 

```
SceneGraph(SceneGraph&& other) : Graph(std::forward<Graph>(other)), link_map_(std::move(other.link_map_)),
    joint_map_(std::move(other.joint_map_)), acm_(std::move(other.acm_))
  {
    {
      link_map_.clear();
      joint_map_.clear();
      Graph::vertex_iterator i, iend;
      for (boost::tie(i, iend) = boost::vertices(*this); i != iend; ++i)
      {
        Link::Ptr link = boost::get(boost::vertex_link, *this)[*i];
        link_map_[link->getName()] = std::make_pair(link, *i);
      }
    }
    {
      Graph::edge_iterator i, iend;
      for (boost::tie(i, iend) = boost::edges(*this); i != iend; ++i)
      {
        Joint::Ptr joint = boost::get(boost::edge_joint, *this)[*i];
        joint_map_[joint->getName()] = std::make_pair(joint, *i);
      }
    }
  }
```

johnwason:
I fixed the copy constructor, and now the return value is working as expected on Windows without needing to modify anything else. I haven't tested on Linux so we have to wait for the CI to finish.

Levi-Armstrong:
@johnwason Thanks for the fixes!

PR #755, opened 2022-04-23
yaml_utils.h nullptr comparison fixup

johnwason:
`yaml_utils.h` uses comparisons to `nullptr` to check if a node is valid. This fails on newer versions of yamlcpp, and isn't necessary since the nodes support boolean comparisons already.


Levi-Armstrong:
Yea this was done to make clang-tidy happy. Since this breaks using newer version of yamlcpp just add `// NOLINT` like what is shown below and clang-tidy will pass.

``` c++
if (node["config"]) // NOLINT
```

PR #756, opened 2022-04-23
Fix JointTrajectory SWIG container

johnwason:
This PR fixes the SWIG wrapping of the `JointTrajectory` container.


Levi-Armstrong (approved):

Levi-Armstrong:
It looks like clang format needs to be ran.

PR #758, opened 2022-04-26
Add factory create callbacks to plugin factories

johnwason:
This PR adds static callback functions to `ContactManagersPluginFactory` and `KinematicsPluginFactory`. These functions are called before attempting plugin loading, and can be used to compile-time link plugin factory creation. This is necessary for Python Wheel distribution, as discussed in #754.


tesseract_collision/core/include/tesseract_collision/core/contact_managers_plugin_factory.h
```diff
@@ -256,12 +261,46 @@ class ContactManagersPluginFactory
...
+   * @brief Set a global discrete contact manager factory create callback
+   *
+   * Callback function is called before plugins to allow explicit
+   * allocation of contact manager factory
+   *
+   * @param fn Contact manager factory create callback function
+   */
+  static void
+  setGlobalCreateDiscreteContactManagerFactoryCallback(const CreateDiscreteContactManagerFactoryCallbackFn& fn);
+
```
Levi-Armstrong, line 274: Could we accomplish this by adding methods to the class like the one below? We have been going through and removing the use of function pointers to better support serialization.

``` c++ 
/**
   * @brief Add a discrete contact manager plugin
   * @param name The name
   * @param plugin The plugin
   */
  void addDiscreteContactManagerPlugin(const std::string& name, DiscreteContactManagerFactory::UPtr plugin);
```

johnwason: These are static functions so they shouldn't be serialized. These should be thought of as implementation details that are invisible to the user and only exist for specific use cases like wheels.

Levi-Armstrong: Would have the proposed method serve the same need so the python wheel  includes the plugin libraries, or does it need to be static?

Levi-Armstrong: My current concern is now there are two maps that could contain the same factories which one takes priority. 

Is it possible to create python specific function that takes a factory and adds the default factories or when the object is initialized in python add the defaults using the following?

``` c++

addDiscreteContactManagerPlugin("BulletDiscreteBVHManagerFactory", std::make_shared<tesseract_collision::tesseract_collision_bullet::BulletDiscreteBVHManagerFactory>());

addDiscreteContactManagerPlugin("BulletDiscreteSimpleManagerFactory", std::make_shared<tesseract_collision::tesseract_collision_bullet::BulletDiscreteSimpleManagerFactory>());

```

johnwason: Python doesn't know when these objects are created since they are created in various places in other libraries.

johnwason: We could add a single static callback to the plug-in loader and intercept requests there.

Levi-Armstrong: Sorry for all of the questions, I just want to make sure I understand things. Is this mainly so that the python wheel pulls in the plugins? If so, if the libraries were included would the current plugin loading approach find them and load them?

johnwason: I tried linking in the plugins and using the loader, but the auditwheel process seems to destroy that functionality. This is intentional so library collisions don't happen.

Levi-Armstrong: Was [this](https://stackoverflow.com/questions/19663366/dlopen-fails-in-boost-python-module) the potential issue?

Levi-Armstrong: Do you have the python wheel that you tried linking in the plugins?

johnwason: That may be part of it. In general you don't want to load symbols into Python because it can break things. Typically the solution was to use static libraries, but auditwheel uses some tricks to make it work by mangling shared libraries.

johnwason: Here is the pull request with the Python changes: https://github.com/tesseract-robotics/tesseract_python/pull/23

The callbacks are configured here:

https://github.com/johnwason/tesseract_python-1/blob/03c1e067925c8d9de7bbba01d2ca0c0682fe6c0e/tesseract_python/swig/tesseract_collision_python.i#L71

https://github.com/johnwason/tesseract_python-1/blob/03c1e067925c8d9de7bbba01d2ca0c0682fe6c0e/tesseract_python/swig/tesseract_kinematics_python.i#L90

johnwason: I am not sure. If the callbacks return nullptr the loader will fall back to the normal library loader. I don't know if the libraries will load since all the other libraries are mangled.

For users that need plugins another method like conda or apt makes more sense, but most people using python won't need more than the built in capabilities.

Levi-Armstrong:
@johnwason Lets add a specific compiler definition for python things like this so it is not part of the normal c++ api.

``` c++
#ifndef TESSERACT_PYTHON
ContinuousContactManager::UPtr
ContactManagersPluginFactory::createContinuousContactManager(const std::string& name,
                                                             const tesseract_common::PluginInfo& plugin_info) const
{
  if (continuous_manager_create_callback_)
  {
    auto cm = continuous_manager_create_callback_(plugin_info.class_name);
    if (cm)
    {
      return cm->create(name, plugin_info.config);
    }
  }

  try
  {
    auto it = continuous_factories_.find(plugin_info.class_name);
    if (it != continuous_factories_.end())
      return it->second->create(name, plugin_info.config);

    auto plugin = plugin_loader_.instantiate<ContinuousContactManagerFactory>(plugin_info.class_name);
    if (plugin == nullptr)
    {
      CONSOLE_BRIDGE_logWarn("Failed to load symbol '%s'", plugin_info.class_name.c_str());
      return nullptr;
    }
    continuous_factories_[plugin_info.class_name] = plugin;
    return plugin->create(name, plugin_info.config);
  }
  catch (const std::exception&)
  {
    CONSOLE_BRIDGE_logWarn("Failed to load symbol '%s'", plugin_info.class_name.c_str());
    return nullptr;
  }
}
#else
ContinuousContactManager::UPtr
ContactManagersPluginFactory::createContinuousContactManager(const std::string& name,
                                                             const tesseract_common::PluginInfo& plugin_info) const
{
 try
  {
    if (continuous_manager_create_callback_)
    {
      auto cm = continuous_manager_create_callback_(plugin_info.class_name);
      if (cm)
      {
        return cm->create(name, plugin_info.config);
      }
    }
 catch (const std::exception&)
  {
    CONSOLE_BRIDGE_logWarn("Failed to load symbol '%s'", plugin_info.class_name.c_str());
    return nullptr;
  }
}
#endif
```

johnwason:
Ok, I'll clean things up later this week.

johnwason:
I added the cmake option `TESSERACT_PLUGIN_FACTORY_CALLBACKS` and appropriate #ifdef blocks to the callback code to have it be disabled by default.

johnwason:
@Levi-Armstrong can we merge this?

Levi-Armstrong:
@johnwason Thanks for the reminder. I will review it this week.

tesseract_common/cmake/tesseract_macros.cmake
```diff
@@ -52,6 +52,10 @@ macro(tesseract_variables)
     set(TESSERACT_ENABLE_CODE_COVERAGE ON)
   endif()
 
+  if(NOT DEFINED TESSERACT_PLUGIN_FACTORY_CALLBACKS)
```
Levi-Armstrong, line 55: Can you documentation on when this should be enabled?

tesseract_kinematics/core/src/kinematics_plugin_factory.cpp
```diff
@@ -244,6 +244,17 @@ KinematicsPluginFactory::createFwdKin(const std::string& solver_name,
                                       const tesseract_scene_graph::SceneGraph& scene_graph,
                                       const tesseract_scene_graph::SceneState& scene_state) const
 {
+#ifdef TESSERACT_PLUGIN_FACTORY_CALLBACKS
```
Levi-Armstrong, line 247: I think it might be best to make this either or implementation.
``` c++
#ifndef TESSERACT_PLUGIN_FACTORY_CALLBACKS
DiscreteContactManager::UPtr
ContactManagersPluginFactory::createDiscreteContactManager(const std::string& name,
                                                           const tesseract_common::PluginInfo& plugin_info) const
{
  try
  {
    auto it = discrete_factories_.find(plugin_info.class_name);
    if (it != discrete_factories_.end())
      return it->second->create(name, plugin_info.config);

    auto plugin = plugin_loader_.instantiate<DiscreteContactManagerFactory>(plugin_info.class_name);
    if (plugin == nullptr)
    {
      CONSOLE_BRIDGE_logWarn("Failed to load symbol '%s'", plugin_info.class_name.c_str());
      return nullptr;
    }
    discrete_factories_[plugin_info.class_name] = plugin;
    return plugin->create(name, plugin_info.config);
  }
  catch (const std::exception&)
  {
    CONSOLE_BRIDGE_logWarn("Failed to load symbol '%s'", plugin_info.class_name.c_str());
    return nullptr;
  }
}
#else
DiscreteContactManager::UPtr
ContactManagersPluginFactory::createDiscreteContactManager(const std::string& name,
                                                           const tesseract_common::PluginInfo& plugin_info) const
{

  if (discrete_manager_create_callback_)
  {
    auto cm = discrete_manager_create_callback_(plugin_info.class_name);
    if (cm)
    {
      return cm->create(name, plugin_info.config);
    }
  }

    CONSOLE_BRIDGE_logWarn("Failed to create plugin '%s' using callback", plugin_info.class_name.c_str());
    return nullptr;
}
#endif
```

Levi-Armstrong (changes requested):

johnwason:
@Levi-Armstrong would it make more sense to move the callbacks to the plugin factory class? That would only modify one place in the code instead of for each type of plugin.

Levi-Armstrong:
> @Levi-Armstrong would it make more sense to move the callbacks to the plugin factory class? That would only modify one place in the code instead of for each type of plugin.

Works for me.

Levi-Armstrong:
> @Levi-Armstrong would it make more sense to move the callbacks to the plugin factory class? That would only modify one place in the code instead of for each type of plugin.

@johnwason I was looking into this but not sure where you were thinking it should go. I was looking to see if we can put it in the PluginLoader but not sure if it can go there. Where were you thinking it could go?

johnwason:
I haven't had a chance to work in this - I will look at it today.

johnwason:
This has been superseded by #782

PR #776, opened 2022-06-14
Add type erasure interface

Levi-Armstrong:


Levi-Armstrong:
@mpowelson, @marip8 and @marrts  This introduces a type erasure interface to handle a lot of the boiler plate code with type erasures and supports boost serialization.

PR #780, opened 2022-06-25
Update windows CI to leverage Colcon

Levi-Armstrong:


johnwason:
I am seeing failures in the tests. I think you need to add the environmental variables to disable vector instructions. You also need to check the return values for each step in a multi-line command.

Levi-Armstrong:
Yea your right. Not sure why colcon is not return an error code which it should with the provided option.  @johnwason why does it have issues with vector instructions?

johnwason:
@Levi-Armstrong There is a problem on windows with alignment and DLLs, so vector instructions don't seem to work properly. 

I use:

    set CXXFLAGS=%CXXFLAGS% -DEIGEN_DONT_ALIGN=1 -DEIGEN_DONT_VECTORIZE=1

to work around the problem.

You need to add 

    if %errorlevel% neq 0 exit /b %errorlevel%

after each line of a multi-line command in the test script. It runs as a `bat` file, and there is no `set -e` equivalent.

PR #782, opened 2022-06-27
Static plugin loading using symbol module resolution

johnwason:
This is another approach to #758. It works by adding "anchor" functions to the plugin libraries, which return an address that exists within the library. The function `boost::dll::symbol_location_ptr()` is then used to get the filename of the library. The library name is then added to the environmental variable for that plugin type.

This approach seems to work, and is a lot less invasive than other options. I will continue testing to see if it will work with Python.


johnwason:
@Levi-Armstrong is this approach ok with you?

Levi-Armstrong:
Why add it to the environment versus just adding it to the search libraries and search libraries path?

tesseract_common/include/tesseract_common/plugin_loader.hpp
```diff
@@ -110,6 +114,12 @@ std::shared_ptr<PluginBase> PluginLoader::instantiate(const std::string& plugin_
     }
   }
 
+  for (const auto& library : plugins_local)
```
Levi-Armstrong, line 117: I believe you found a bug, The above for loop should use `plugins_local` instead of search_libraries. I believe this can be removed if that change is made.

Levi-Armstrong: It looks like there is another for loop above that should also be using the same variable.

johnwason: Yeah this function has a couple strange issues that need to be cleaned up

tesseract_common/include/tesseract_common/plugin_loader.h
```diff
@@ -118,6 +118,19 @@ class PluginLoader
...
+   * 
+   * In some cases the name and location of a library is unknown at runtime, but a symbol can
+   * be linked at compile time. This is true for Python auditwheel distributions. This
+   * utility function will determine the location of the library, and add it to the library search
+   * environment variable so it can be found.
+   * 
+   * @param symbol_ptr Pointer to the symbol to find
+   * @param search_libraries_env The environmental variable to modify
+   */
+  static inline void addSymbolLibraryToSearchLibrariesEnv(const void* symbol_ptr, const std::string& search_libraries_env);
```
Levi-Armstrong, line 133: This should also update the environment variable for search paths.
``` c++
void addSymbolLibraryToSearchLibrariesEnv(const void* symbol_ptr, const std::string& search_libraries_env, const std::string& search_paths_env)
{
  std::string search_libraries_env_var_str;
  char* search_libraries_env_var = std::getenv(search_libraries_env.c_str());
  if (search_libraries_env_var != nullptr)
    search_libraries_env_var_str = search_libraries_env_var;

  std::string search_paths_env_var_str;
  char* search_paths_env_var = std::getenv(search_paths_env.c_str());
  if (search_paths_env_var != nullptr)
    search_paths_env_var_str = search_paths_env_var;

  boost::dll::fs::path lib_path = boost::dll::symbol_location_ptr(symbol_ptr);

  std::string separator {":"};
#ifdef _WIN32
  separator = ";";
#endif

  if (search_libraries_env_var_str.empty())
    search_libraries_env_var_str = lib_path.filename().string();
  else
    search_libraries_env_var_str = search_libraries_env_var_str + separator + lib_path.filename().string();

  if (search_paths_env_var_str.empty())
    search_paths_env_var_str = lib_path.parent_path().string();
  else
    search_paths_env_var_str = search_paths_env_var_str + separator + lib_path.parent_path().string();

#ifndef _WIN32
  setenv(search_libraries_env.c_str(),search_libraries_env_var_str.c_str(),1);
  setenv(search_paths_env.c_str(),search_paths_env_var_str.c_str(),1);
#else
  _putenv_s(search_libraries_env.c_str(),search_libraries_env_var_str.c_str());
  _putenv_s(search_paths_env.c_str(),search_paths_env_var_str.c_str());
#endif
}
```

Levi-Armstrong: The library environment variable should only contain the library name and not the full path.

Levi-Armstrong: For this to work I also had to update the decorate function to check if an extension already exists before adding a suffix.
``` c++
  if (actual_path.extension().empty())
    actual_path += boost::dll::shared_library::suffix();
```

johnwason: It is a lot safer to use absolute paths for the library loading with this method. Otherwise it may wind up loading an unexpected library it finds somewhere else in the search paths.

Levi-Armstrong: Paths provided by the environment variable are search first for this reason.

Levi-Armstrong: It may be good to have this method prepend  instead of append to the environment variable.

johnwason: I am not sure what the issue is with the library name or absolute path, it is working fine with the test I added.

johnwason: Could we add another environmental variable or a static field that has absolute paths?

Levi-Armstrong: I guess we could change it to loop over the libraries and check for full path. If full path these get searched first which makes sense to me.

johnwason: Yeah, that should work. I also added logic to check for `:` on Linux and `;` on Windows. This is an annoying difference between operating systems since `:` can be part of a path on Windows.

Levi-Armstrong: @marip8 What do you think here?

Levi-Armstrong: @johnwason I updated it to support libraries provided as full path. The changes can be seen if you follow the line to the PR below. We are waiting on this to be released on the build farm to switch this repository over to using.

johnwason: @Levi-Armstrong are you going to backport your updates or should I include them here?

Levi-Armstrong: I think it would be good to back port them here this way to get things merged in here for now. We are going to make some changes it boost_plugin_loader to better support this functionality but may be a week or so before it gets merged in.

Levi-Armstrong:
It looks like one  of the plugin loader tests is failing

johnwason:
I am still working on this PR

johnwason:
This PR is ready now

PR #787, opened 2022-06-28
Improve manipulability calculcation

rhaschke:
I just came across these minor improvements of the manipulability calculation when inspecting the code, because a student of mine complained that the eigenvalues are not sorted. Please have a look at the individual commit comments for more details.
For reference: [SelfAdjointEigenSolver](http://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html)

In principle, the calculation of the inverse expressions is redundant:
https://github.com/tesseract-robotics/tesseract/blob/afe92a7ab892dad856a010d3c7864489d7f3489a/tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h#L355-L360
because eigenvalues(A^-1) = eigenvalues^-1(A). Also, det(A^-1) = 1/det(A).
However, you apply some clipping to the eigenvalues as well...

### DIFF
--- a/tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
+++ b/tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
@@ -310,12 +310,16 @@ inline Manipulability calcManipulability(const Eigen::Ref<const Eigen::MatrixXd>
     Eigen::SelfAdjointEigenSolver<Eigen::MatrixXd> sm(m, false);
 
     data.eigen_values = sm.eigenvalues().real();
+    data.volume = 1;
 
-    // Set values near zero to zero
+    // Set eigenvalues near zero to zero. This also implies zero volume
     for (Eigen::Index i = 0; i < data.eigen_values.size(); ++i)
     {
       if (tesseract_common::almostEqualRelativeAndAbs(data.eigen_values[i], 0))
+      {
         data.eigen_values[i] = +0;
+        data.volume = +0;
+      }
     }
 
     // If the minimum eigen value is approximately zero set measure and condition to max double
@@ -330,18 +334,9 @@ inline Manipulability calcManipulability(const Eigen::Ref<const Eigen::MatrixXd>
       data.measure = std::sqrt(data.condition);
     }
 
-    data.volume = 1;
-    for (Eigen::Index i = 0; i < sm.eigenvalues().size(); ++i)
-    {
-      // If an eigen value is approximately zero set the volume to zero and return
-      if (tesseract_common::almostEqualRelativeAndAbs(data.eigen_values[i], 0))
-      {
-        data.volume = +0;
-        break;
-      }
-      data.volume = data.volume * data.eigen_values[i];
-    }
-    data.volume = std::sqrt(data.volume);
+    if (data.volume != 0)  // volume is sqrt of product of eigenvalues
+      data.volume = std::sqrt(data.eigen_values.prod());
+
     return data;
   };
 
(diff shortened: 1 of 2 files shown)


tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
```diff
@@ -330,18 +334,9 @@ inline Manipulability calcManipulability(const Eigen::Ref<const Eigen::MatrixXd>
...
-      // If an eigen value is approximately zero set the volume to zero and return
-      if (tesseract_common::almostEqualRelativeAndAbs(data.eigen_values[i], 0))
-      {
-        data.volume = +0;
-        break;
-      }
-      data.volume = data.volume * data.eigen_values[i];
-    }
-    data.volume = std::sqrt(data.volume);
+    if (data.volume != 0)  // volume is sqrt of product of eigenvalues
```
Levi-Armstrong, line 337: I see why this change was added, but comparing floats this way is unlikely to never be false.  What about switching it to use the following above `bool zero_volume {false}` then set it to true in the if statement and check it here?

Levi-Armstrong (changes requested):
Overall, looks good just had one comment.

tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
Levi-Armstrong: Another option would be to just calculate the volume because the product should always be greater than zero. It would also probably be good to add this just as a precaution `assert(data.eigen_values.prod() > 0)`. You don't have to add this as part of the PR, but will not mind if you do.

Levi-Armstrong: > To be honest, the assertion is not necessary either, as the eigenvalues of a matrix AA^t are the squares (and thus positive) of the (real) singular values of A.

Yea, though I have had issue in the past with this code where I was getting eigen values slightly negative which is why there is the check above. I believe the check should solve the issue but cannot be 100% sure.

Levi-Armstrong (approved):

Levi-Armstrong:
@rhaschke Does the switch in solver now sort the eigen values?

rhaschke:
> Does the switch in solver now sort the eigen values

Yes. Quoting the [Eigen doc](http://eigen.tuxfamily.org/dox/classEigen_1_1SelfAdjointEigenSolver.html#aaf4ed4172a517a4b9f0ab222f629e261): The eigenvalues are sorted in increasing order.

tesseract_kinematics/core/include/tesseract_kinematics/core/utils.h
Levi-Armstrong: That makes sense. If that is the case then the check may no longer be needed.

rhaschke:
A test is failing now. Will have a look. Strangely, I couldn't find any output of the failing test.
Is there a specific reason for not using `catkin_add_gtest()` but `add_gtest_discover_tests()`, which obviously requires a manual call to `catkin build` to run the tests:
https://github.com/tesseract-robotics/tesseract/blob/98964bca335bfd29fcb66738dfe7bc46c191649a/.github/workflows/bionic_build.yml#L97

cpetersmeier:
Since we switched to the `SelfAdjointEigenSolver`, which sorts the eigenvalues, the manipulability test has to be adapted as well.

I think I cannot push to your branch @rhaschke , but I introduced the necessary changes [here](https://github.com/cpetersmeier/tesseract/commit/c2f6ba830096709ac89f4bab9a1376a44c883de2).

Levi-Armstrong:
> A test is failing now. Will have a look. Strangely, I couldn't find any output of the failing test.
> Is there a specific reason for not using `catkin_add_gtest()` but `add_gtest_discover_tests()`, which obviously requires a manual call to `catkin build` to run the tests:

Yea, everything except tesseract_ros is agnostic to ros and only uses cmake feature. The use of colcon solves this issue but have not got around to updating this repository.

Levi-Armstrong (approved):

PR #789, opened 2022-06-30
Fix Windows CI by checking for error after colcon commands

Levi-Armstrong:


Levi-Armstrong:
@johnwason It looks like the tests that are failing are those that are writing files to the temp directory.

PR #792, opened 2022-07-02
CI: Drop AFTER_SCRIPT

rhaschke:
Explicit test running via `AFTER_SCRIPT` was required because industrial_test didn't use the correct make target for cmake-only packages. This is fixed with https://github.com/ros-industrial/industrial_ci/pull/785 and thus `AFTER_SCRIPT` is superfluous now.


rhaschke:
Unfortunately, the new scheme requires catkin tools 0.9 and thus is limited to the Focal build. For symmetry reasons, you might want to keep the old scheme everywhere. I just wanted to point out the new feature.

Levi-Armstrong:
> Unfortunately, the new scheme requires catkin tools 0.9 and thus is limited to the Focal build. For symmetry reasons, you might want to keep the old scheme everywhere. I just wanted to point out the new feature.

We

rhaschke:
Using colcon instead of catkin tools should be fine too. However, I don't see the switch yet ;-)

PR #801, opened 2022-07-14
Use absolute paths for bullet libraries

johnwason:
Currently the bullet libraries are being included using relative paths, and the linker is finding the wrong libraries in some unexpected paths. This PR uses `find_library()` on each library with the desired path, and includes the libraries as absolute paths.

### DIFF
--- a/tesseract_collision/bullet/CMakeLists.txt
+++ b/tesseract_collision/bullet/CMakeLists.txt
@@ -27,6 +27,14 @@ else()
   set(BULLET_LIBRARY_DIRS_ABS ${BULLET_LIBRARY_DIRS})
 endif()
 
+set(BULLET_LIBRARIES_ABS "")
+foreach(BULLET_LIB IN LISTS BULLET_LIBRARIES)
+  find_library(BULLET_LIB_ABS_${BULLET_LIB} ${BULLET_LIB} PATHS ${BULLET_LIBRARY_DIRS_ABS} NO_DEFAULT_PATH REQUIRED)
+  list(APPEND BULLET_LIBRARIES_ABS "${BULLET_LIB_ABS_${BULLET_LIB}}")
+  message(STATUS "BULLET_LIB=${BULLET_LIB} BULLET_LIB_ABS=${BULLET_LIB_ABS_${BULLET_LIB}}")
+endforeach()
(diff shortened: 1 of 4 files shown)


tesseract_collision/bullet/CMakeLists.txt
```diff
@@ -27,6 +27,14 @@ else()
   set(BULLET_LIBRARY_DIRS_ABS ${BULLET_LIBRARY_DIRS})
 endif()
 
+set(BULLET_LIBRARIES_ABS "")
```
Levi-Armstrong, line 30: Does the code above not produce absolute paths?

johnwason: No, that just changes the include directory. The libraries themselves are still without a path, and that was causing a lot of problems because it was finding them in the wrong place. Without the current changes, it wasn't even searching the directory specified by the bullet path because there was no `link_directories()` call.

johnwason: This is the content of `BulletConfig.cmake`

```cmake
set ( BULLET_FOUND 1 )
set ( BULLET_USE_FILE     "lib/cmake/bullet/UseBullet.cmake" )
set ( BULLET_DEFINITIONS  "-DBT_USE_DOUBLE_PRECISION" )
set ( BULLET_INCLUDE_DIR  "include/bullet" )
set ( BULLET_INCLUDE_DIRS "include/bullet" )
set ( BULLET_LIBRARIES    "LinearMath;Bullet3Common;BulletInverseDynamics;BulletCollision;BulletDynamics;BulletSoftBody" )
set ( BULLET_LIBRARY_DIRS "lib" )
set ( BULLET_ROOT_DIR     "C:/ws/tesseract_noetic_ci/tesseract_ws/install_isolated" )
set ( BULLET_VERSION_STRING "3.25" )
```

`BULLET_LIBRARIES` does not include a path, so it basically searches the filesystem.

Levi-Armstrong:
I think we should add a macro to tesseract_macros which locates bullet and creates a target which is used. I will see if I can move this into tesseract_macros.cmake

Levi-Armstrong:
Incorporated into PR #803

PR #803, opened 2022-07-17
Add find_bullet macro which creates a target to link against

Levi-Armstrong:


Levi-Armstrong:
This incorporates PR #801 changes into a macro which creates a target that is used within tesseract.

johnwason:
Do we need to export the Bullet target? The use of bullet should be self contained.

Levi-Armstrong:
> Do we need to export the Bullet target? The use of bullet should be self contained.

This macro is called in the tesseract_collision-config.cmake which handles creation of the target for any package linking against tesseract_collision

Levi-Armstrong:
@johnwason Everything is passing and should provide the same functionality. Let me know what you think.

johnwason:
I am running a test here: https://github.com/johnwason/tesseract_python-1/pull/2

johnwason:
Also see https://github.com/microsoft/vcpkg/pull/25784

I added the features double-precision and rtti to the vcpkg bullet package.

Use:

    bullet3[double-precision,rtti,multithreading]

with vcpkg to get the required features, once vcpkg completes the merge.

johnwason (approved):

PR #831, opened 2022-11-12
Add color, image and material classes

Levi-Armstrong:
These are taken from Ignition repository.

Remaining Tasks:

- [ ] Port unit tests
- [ ] Port PBR material
- [ ] Finish out material class


johnwason:
We already have a lot of these capabilities implemented. What do you think is missing?

PR #832, opened 2022-11-29
Add ToolPath and ToolPathSegment class

Levi-Armstrong:


marip8:
What are you planning to use these for?

Levi-Armstrong:
Primarily for tool path visualization in tesseract_qt and filtering operations in a larger task composer graph which removes unreachable waypoints, small segments and breaking up large segments at configuration flips.

marip8:
> Primarily for tool path visualization in tesseract_qt and filtering operations in a larger task composer graph which removes unreachable waypoints, small segments and breaking up large segments at configuration flips.

Sounds familiar...

Why not just use a composite instruction of Cartesian waypoints or move instructions instead? Seems like that probably has all the information necessary to do what you're looking for without having to add and maintain a new set of classes

Levi-Armstrong:
> Why not just use a composite instruction of Cartesian waypoints or move instructions instead? Seems like that probably has all the information necessary to do what you're looking for without having to add and maintain a new set of classes

Yea, I thought about that but but was not sure. It would be good to meet and discuss.

Levi-Armstrong:
I think the plan is to leverage the composite instruction class. Once we move tesseract_planning into the tesseract repo we can simplify/remove other container objects like joint_trajectory etc. We could also just move that package to tesseract repo now. @marip8 and @rjoomen what do you think about move command_language to the tesseract repo now?

PR #843, opened 2023-01-31
Add the ability to change KDL parameters from kinematics configuration

seanmc602:


tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
```diff
@@ -102,6 +102,22 @@ struct KDLChainData
...
+ * They are ultimately passed to the constuctor of the undelying ChainIkSolver.
+ *
+ * The defaults provided here are the same defaults imposed by the KDL library.
+ */
+struct KDLConfig
+{
+  Eigen::VectorXd weights = Eigen::VectorXd::Constant(6, 1.0);
+  double eps = 1E-5;
+  int max_iterations = 500;
+  double eps_joints = 1E-15;
```
Levi-Armstrong, line 118: nitpik, please use `{}`
``` c++
  Eigen::VectorXd weights {Eigen::VectorXd::Constant(6, 1.0)};
  double eps {1E-5};
  int max_iterations {500};
  double eps_joints {1E-15};
```

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
```diff
@@ -102,6 +102,22 @@ struct KDLChainData
...
+ * @brief The KDLConfig struct
+ *
+ * This contains parameters that can be used to customize the KDL solver for your application.
+ * They are ultimately passed to the constuctor of the undelying ChainIkSolver.
+ *
+ * The defaults provided here are the same defaults imposed by the KDL library.
+ */
+struct KDLConfig
+{
+  Eigen::VectorXd weights = Eigen::VectorXd::Constant(6, 1.0);
```
Levi-Armstrong, line 115: The default assumes 6 joint which is not good. It may be better to unit size as the default and add a check in the construction to handle this so it creates the correct size based on the number of actual joints.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
Levi-Armstrong: Also are these default the current ones used by KDL?

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
```diff
@@ -102,6 +102,22 @@ struct KDLChainData
...
 
+/**
+ * @brief The KDLConfig struct
+ *
+ * This contains parameters that can be used to customize the KDL solver for your application.
+ * They are ultimately passed to the constuctor of the undelying ChainIkSolver.
+ *
+ * The defaults provided here are the same defaults imposed by the KDL library.
+ */
+struct KDLConfig
```
Levi-Armstrong, line 113: Does this config only apply to KDL LMA?

Levi-Armstrong (changes requested):
Thank you for the PR.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
marip8: Yes, [per here](https://docs.ros.org/en/kinetic/api/orocos_kdl/html/classKDL_1_1ChainIkSolverPos__LMA.html#a5322326848d618e0ec2af8683344274e)

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
marip8: I think these may be task space coordinate weights (i.e., [x, y, z, rx, ry, rz]). It looks like [the default values are actually [1, 1, 1, 0.01, 0.01, 0.01]](https://docs.ros.org/en/kinetic/api/orocos_kdl/html/classKDL_1_1ChainIkSolverPos__LMA.html#a1e83b8d5bb692211d4216e491253afa4), which seems like position only IK. @seanmc602 is this correct?

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
marip8: Yes; the other solvers take similar, but not exactly the same, parameters

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
seanmc602: Yes, task space weights. I'll update the variable name to match the config name to be more clear.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
seanmc602: Michael is correct. I actually started down the path of using this struct for the other KDL factories but stopped when I noticed the different defaults for each.

How would you like me to proceed? Use the same struct but change the defaults to match the constructor for each version? (NR and Fwd)

Levi-Armstrong: I am fine with creating a different config for each which have different default and moving the struct into the header of the different ones.

Levi-Armstrong:
``` c++
struct KDLInvKinChainLMAConfig
{
}
```

Or you can define the struct in the the KDLInvKinChainLMA class where it would be `struct Config`.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
Levi-Armstrong, line 115: I would also use a fixed size like what was done in KDL. Also will need to add the Eigen align macro to this struct.
```  c++
Eigen::Matrix< double, 6, 1 >
```

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
Levi-Armstrong: @marip8 Do you have a preference?

marip8: I think having separate configuration structs would be good. I usually embed that struct inside the corresponding class, but either way is fine with me

Levi-Armstrong: Okay, @seanmc602 lets embed the config struct in the corresponding class.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_inv_kin_chain_nr.h
```diff
@@ -61,6 +61,26 @@ class KDLInvKinChainNR : public InverseKinematics
...
+   */
+  struct Config
+  {
+    double vel_eps{ 0.00001 };
+    int vel_iterations{ 150 };
+    double pos_eps{ 1e-6 };
+    int pos_iterations{ 100 };
+
+  public:
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
Levi-Armstrong, line 81: This can be removed since it does not contain Eigen fixed types.

tesseract_kinematics/kdl/src/kdl_inv_kin_chain_nr.cpp
```diff
@@ -71,9 +75,9 @@ KDLInvKinChainNR::KDLInvKinChainNR(const KDLInvKinChainNR& other) { *this = othe
 KDLInvKinChainNR& KDLInvKinChainNR::operator=(const KDLInvKinChainNR& other)
 {
   kdl_data_ = other.kdl_data_;
-  fk_solver_ = std::make_unique<KDL::ChainFkSolverPos_recursive>(kdl_data_.robot_chain);
-  ik_vel_solver_ = std::make_unique<KDL::ChainIkSolverVel_pinv>(kdl_data_.robot_chain);
-  ik_solver_ = std::make_unique<KDL::ChainIkSolverPos_NR>(kdl_data_.robot_chain, *fk_solver_, *ik_vel_solver_);
+  fk_solver_ = other.fk_solver_;
+  ik_vel_solver_ = other.ik_vel_solver_;
+  ik_solver_ = other.ik_solver_;
```
Levi-Armstrong, line 80: These changes need to be reverted because they cause issues with multi threading. The  solvers need to be unique to each object.

tesseract_kinematics/kdl/src/kdl_inv_kin_chain_lma.cpp
```diff
@@ -69,7 +72,7 @@ KDLInvKinChainLMA::KDLInvKinChainLMA(const KDLInvKinChainLMA& other) { *this = o
 KDLInvKinChainLMA& KDLInvKinChainLMA::operator=(const KDLInvKinChainLMA& other)
 {
   kdl_data_ = other.kdl_data_;
-  ik_solver_ = std::make_unique<KDL::ChainIkSolverPos_LMA>(kdl_data_.robot_chain);
+  ik_solver_ = other.ik_solver_;
```
Levi-Armstrong, line 75: This need to be reverted because it causes issues with mult threading. They need to be unique per object.

Levi-Armstrong (changes requested):

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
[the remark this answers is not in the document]
Levi-Armstrong: Oh, can this struct be removed now?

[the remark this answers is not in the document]
seanmc602: Yeah, it's leftover. My fault

Levi-Armstrong:
I think the only thing left is to updated unit tests. I think if you give the config a default in the header it will fix most of them. Also if there are unit tests for parsing the yaml file those will need to be updated to account for the new parameters. It may be easiest to add a parser code for each of the structs.

PR #847, opened 2023-02-21
Deploy documentation to separate repository

marip8:
Use the [option to deploy documentation to a separate repository](https://github.com/peaceiris/actions-gh-pages/tree/e5d60a65a4bd200dd1f942fa8e0f677ba7167370#%EF%B8%8F-deploy-to-external-repository-external_repository) in order to remove the large `gh-pages` branch from this repo to improve download time. Addresses #840 and #842

### DIFF
--- a/.github/workflows/api_docs.yml
+++ b/.github/workflows/api_docs.yml
@@ -25,3 +25,4 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs
           keep_files: true
+          external_repository: tesseract-robotics/tesseract_docs
--- a/.github/workflows/api_docs.yml
+++ b/.github/workflows/api_docs.yml
@@ -24,5 +24,6 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs
+          destination_dir: tesseract
           keep_files: true
           external_repository: tesseract-robotics/tesseract_docs
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -69,5 +69,5 @@ jobs:
 
       # PERSONAL_GITHUB_TOKEN needed here since we are pushing to a branch
       - name: Push benchmark result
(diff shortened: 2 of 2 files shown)


marip8:
@Levi-Armstrong is there a way to test this job before merging? I see it doesn't run on PRs

Levi-Armstrong:
How does this work? Does it push to that repository's gh_pages branch?

Levi-Armstrong:
Okay so it publishes to the gh_pages branch on the provided repository. Though this does not allow us to remove the gh_pages branch in this repo yet because we need to move benchmark to also publish to that repository.

marip8:
> Okay so it publishes to the gh_pages branch on the provided repository.

Yes, that's what I understood from the documentation. That's why I wanted to check before merging though.

> we need to move benchmark to also publish to that repository.

I just update the benchmarking job

marip8:
If this works, should we change the deploy branch from `gh-pages` to `tesseract`? Then we could use the same `tesseract_docs` repository to host documentation from this repo, `tesseract_planning`, `tesseract_ros`, `tesseract_ros2`, etc.

Levi-Armstrong:
Probably, we need to check on the GitHub website feature. I think you only get one per repo and not sure if you can point a repo to another repos branch for this feature.

marip8:
Okay; we can address that in future. Is there any way to test those workflows before merging, or do we just move forward and fix later?

gavanderhoorn:
Just to draw attention to it: it'd be great to remove the deployed documentation from this repository.

But to reduce the size of the one it's in now, it'd have to be pruned from history. That would require pruning the branch. I'm not sure whether that'd have any effect on `master` or not.

marip8:
> Can you have it upload it to the gh_pages branch but as a subdirectory called tesseract?

It appears that there is a [`destination_dir` variable](https://github.com/peaceiris/actions-gh-pages/tree/884a0224fd48faeb3bde89519e9d612d0585a679#%EF%B8%8F-deploy-to-subdirectory-destination_dir) that allows you to specify a subdirectory where the generated files should go. I think this should work

Levi-Armstrong (approved):

Levi-Armstrong:
I think after merging you should be able to trigger the CI for documentation manually to see if it works as expected.

PR #852, opened 2023-03-01
Performance improvements found using callgrind

Levi-Armstrong:


marip8 (approved):
Out of curiosity, do you have any before/after stats for the performance improvement?

Levi-Armstrong:
In the kdl state solver 8% of the time for calculating forward state was spent in this function. Now it does not show up.

PR #856, opened 2023-03-02
Benchmark CI Job Update

marip8:
Update per commit message. Addresses #854


marip8:
@Levi-Armstrong the required builds above won't run due to the path filters added in #851 (since I only modified the benchmark job). This should be good to merge after a review

Levi-Armstrong (approved):

Issue #834, opened 2022-12-16
ROS-Independent CI


marip8:
Currently the Linux CI jobs use a custom docker image with ROS1 components installed. This is a bit unnecessary and causes problems for downstream ROS-independent or ROS2 builds that use the docker image generated by `tesseract` CI (like [here](https://github.com/tesseract-robotics/tesseract_ros2/pull/35#issuecomment-1353811194)). We should update the CI to be independent of ROS as in https://github.com/tesseract-robotics/boost_plugin_loader/pull/12.

Levi-Armstrong:
What about using colcon but avoid using industrial_ci and ros dependendecies?

marip8:
Yes, that's what I'm suggesting. That's what we did in the plugin loader library

marip8:
I recently opened a PR to add this capability to `industrial_ci`: https://github.com/ros-industrial/industrial_ci/pull/818.

@Levi-Armstrong what do you think about transitioning the builds over to this branch of ICI? It would make the CI builds cleaner and make building downstream CI pipelines for `tesseract_ros/2` and other dependent projects easier to do.

Levi-Armstrong:
I don't have an issue with that. Have you received any feedback on your PR?

marip8:
Not yet and no comments on the corresponding issue I opened; I'm not sure who to ping for a review

marip8:
Closing; addressed in #861

Issue #828, opened 2022-10-19
Octomap segfault


marip8:
Here is a [PR](https://github.com/OctoMap/octomap/pull/288/files#diff-45151bec6492789e0b7d1128eb0a8488442d65ccc2756d631d68219c2a8cd180) which fixes the issue. I think we need to add the same logic to the pruneNode in tesseract_geometry

_Originally posted by @Levi-Armstrong in https://github.com/tesseract-robotics/tesseract_planning/issues/255#issuecomment-1278024415_

marip8:
Per @Levi-Armstrong 

> I tried to add the following in the prune function, but children is not accessible.
> 
> ```
>         if (node->children != NULL){
>           delete[] node->children;
>           node->children = NULL;
>         }
> ```

Levi-Armstrong:
I am starting to think there may be an issue with it finding different version of the library. The system installs version 1.9.3 and ros noetic has its own version of 1.9.8. If you run ldd on the example library and executable which is creating the error do you only see it linked against one version?

marip8:
> I am starting to think there may be an issue with it finding different version of the library.

This is for sure caused by using the system version of `liboctomap-dev` rather than the one provided by the ROS build farm. I suppose there is a patch somewhere between 1.9.3 and 1.9.8 that resolves this issue

marip8:
@Levi-Armstrong how do you feel about adding the correct version of `octomap` to the `dependencies.rosinstall`? It's not ideal, but I can't find any later versions distributed on PPAs, etc. It also looks like jammy only provides version 1.9.7, which may also not solve the issue

Levi-Armstrong:
I think add in it to the rosinstall is fine. Should we also add it to the CMake find_package for octomap also?

Levi-Armstrong:
If this occurs on you machine you might run ldd and see if it is linking against both libraries which can cause weird behaviors like this.

marip8:
It looks like it only links against the system version

PR #861, opened 2023-03-09
Use ROS-independent version of ICI

marip8:
Changes per commit message to address #834


marip8:
It looks like the Focal builds are failing on the octomap unit test due to the issue documented in #828. Also the Bionic build can't find `gtest` (maybe because it is no longer being provided by `ros_industrial_cmake_boilerplate`?).

Levi-Armstrong:
It does not look like the libraries are provided on 18.04. I am fine with removing this CI build because it will be EOL in April.

dependencies.repos
```diff
@@ -6,3 +6,11 @@
     local-name: opw_kinematics
     uri: https://github.com/Jmeyer1292/opw_kinematics.git
     version: 0.5.0
+- git:
+    local-name: fcl
+    uri: https://github.com/flexible-collision-library/fcl.git
+    version: 0.6.1
```
Levi-Armstrong, line 12: Why did you have to add fcl?

marip8: Version 0.6.1 is not distributed and `tesseract_collision` requires at least that version

Levi-Armstrong: Can we not install it from ros?

marip8: My hope is to make this a fully ROS-independent build to support either ROS1 or ROS2 downstream repos, so no...

Levi-Armstrong: Ah, makes sense.

marip8:
I merged a few of the jobs together to reduce code duplication. It's set up now to run all Ubuntu and Windows jobs in a matrix to support the same build steps for multiple distributions. I also merged the two code quality jobs (code coverage and clang tidy) in a matrix. It was getting difficult to make sure all the variables were set correctly, so this should help make introducing jobs for new distros easier

Levi-Armstrong:
Looks good!

marip8:
Looks like something is wrong with the Windows build since it's failed the last couple of merged PRs, but this PR should be ready for review

Levi-Armstrong:
Windows is failing because I updated the ros_industrial_cmake_bolilerplate version and it no longer can find gtest.

marip8:
> I would just add gtest to the vcpkg list

That apparently doesn't work, unless it's under a different name than `ros-industrial-cmake-boilerplate`

Levi-Armstrong:
> > I would just add gtest to the vcpkg list
> 
> That apparently doesn't work, unless it's under a different name than `ros-industrial-cmake-boilerplate`. I also can't find it on the [`vcpkg` website](https://vcpkg.io/en/packages.html)

Need to add `gtest` to the list.

Levi-Armstrong:
https://github.com/microsoft/vcpkg/tree/master/ports/gtest

marip8:
Ah my mistake; was stuck on RICB for some reason

Levi-Armstrong:
The issue now is that octomap repository is to being build first. I not sure why but it could be because it has a top level CMakeLists.txt file and colcon is not searching deeper to the sub directories which have a package.xml.

Levi-Armstrong:
I would just add a separate rosinstall file for windows and remove fcl and octomap. Then add fcl and octomap back to vcpkg list.

Levi-Armstrong:
LGTM, I recommend squash merging this when you are ready

PR #871, opened 2023-03-21
Changed fcl to a system dependency

rjoomen:
`fcl` is available as a system dependency using `rosdep` in most distros.


rjoomen:
Sorry, `fcl` version on Ubuntu Focal is too old. BTW, as far as I tried `fcl` 0.7.0 also works, so that could be updated in the dependencies. I didn't run the tests though.

PR #873, opened 2023-03-30
Add trajectory link command

Levi-Armstrong:


tesseract_environment/src/environment.cpp
```diff
@@ -1355,6 +1352,12 @@ bool Environment::applyCommandsHelper(const Commands& commands)
         success &= applySetActiveDiscreteContactManagerCommand(cmd);
         break;
       }
+      case tesseract_environment::CommandType::ADD_TRAJECTORY_LINK:
+      {
+        auto cmd = std::static_pointer_cast<const AddTrajectoryLinkCommand>(command);
+        success &= applyAddTrajectoryLinkCommand(cmd);
+        break;
+      }
```
mpowelson, line 1360: Just looking for similar places it looks like this needs to be added in tesseract_environment_unit.cpp as well as some places in tesseract_qt and tesseract_ros

tesseract_environment/src/environment.cpp
```diff
@@ -1423,23 +1545,23 @@ bool Environment::applyAddCommand(AddLinkCommand::ConstPtr cmd)
 
   if (!link_exists && joint_exists)
   {
-    CONSOLE_BRIDGE_logWarn("Tried to add link (%s) which already exists with a joint provided which does not exist. "
+    CONSOLE_BRIDGE_logWarn("Tried to add link (%s) which does not exists with a joint provided which does not exist. "
```
mpowelson, line 1548: I think this is still wrong.  "Tried to add link (%s) which does not exist with a joint provided which **already** exists."

tesseract_environment/src/environment.cpp
```diff
@@ -1382,37 +1385,156 @@ bool Environment::applyCommandsHelper(const Commands& commands)
...
+      return false;
+    }
+
+    if (state.position.rows() == 0)
+    {
+      CONSOLE_BRIDGE_logWarn("Tried to add trajectory link (%s) with empty position.", cmd->getLinkName().c_str());
+      return false;
+    }
+
+    if (state.joint_names.size() != state.position.rows())
```
mpowelson, line 1450: Should .rows be .size()? Can't you compile eigen with different settings to change the shape of vectors? That would probablyhorribly break tesseract anyway

tesseract_environment/src/environment.cpp
```diff
@@ -1382,37 +1385,156 @@ bool Environment::applyCommandsHelper(const Commands& commands)
...
+    tesseract_scene_graph::SceneState scene_state = state_solver->getState(state.joint_names, state.position);
+    if (joint_names.empty() || !tesseract_common::isIdentical(state.joint_names, joint_names, false))
+    {
+      joint_names = state.joint_names;
+      active_link_names = scene_graph_->getJointChildrenNames(joint_names);
+
+      if (std::find(active_link_names.begin(), active_link_names.end(), cmd->getParentLinkName()) !=
+          active_link_names.end())
+      {
+        CONSOLE_BRIDGE_logWarn("Tried to add trajectory link (%s) where parent link is an active link.",
```
mpowelson, line 1466: Why is that not allowed?

tesseract_environment/src/environment.cpp
Levi-Armstrong: Fixed

tesseract_environment/src/environment.cpp
Levi-Armstrong: Yea, I have branches on all of the other repos and some may already be PR's. After merged I will cleanup and make sure CI passes on those repos.

tesseract_environment/src/environment.cpp
Levi-Armstrong: Yep, fixed.

tesseract_environment/src/environment.cpp
Levi-Armstrong: Currently when created the joint which attaches the new link it assumes that the parent_link is fixed for all states in the trajectory. Not sure how to make it work if the parent link is allowed to change state to state because it is active.

mpowelson: Wait. I just looked at this again. This is only checking if it is an active link of the current trajectory, right? Not global active links of the environment? If so, then that makes sense. I was just thinking if you had 2 arms on a base link that could move  - e.g. a gantry or something

PR #765, opened 2022-05-25
Trajectory Collision Logging

marrts:
The goal of this is to improve the terminal output when performing a collision check on a trajectory. In doing this I also created objects for holding collision information from a trajectory that may be useful elsewhere.

Two examples of partial trajectory outputs from running the unit tests:
short_traj_output
long_traj_output

As compared to the previous output that was much harder to read:
image

The trajectory collision object provides a function that outputs a std::stringstream in the form of the table seen above. All the columns are adjusted to always be a constant width depending on the longest string in that column, so the output should always look clean, similar to what is done with the trajopt iterations.

PR #887, opened 2023-05-11
implemented benchmarking for checking trajectories

marrts:
Finally getting around to benchmarking the collision checking for #765


marrts:
I just copied all the trajectories that were made and used in the main environment unit test, but I didn't bother to do the first only/last only type tests, only the full trajectory.

Levi-Armstrong (approved):

PR #896, opened 2023-05-30
Update VHACD to latest (v4.1, tag 454913f)

rjoomen:
This updates the outdated [VHACD library](https://github.com/kmammou/v-hacd) to the latest version. It also removes OMP and OpenCL dependencies, as VHACD now works without these.


rjoomen:
@Levi-Armstrong I'm not sure how to satisfy clang-tidy. Using NOLINTBEGIN/NOLINTEND is only supported from clang-format 14 onwards, and TESSERACT_COMMON_IGNORE_WARNINGS_PUSH/TESSERACT_COMMON_IGNORE_WARNINGS_POP doesn't work for analyzer checks. Any ideas?

Levi-Armstrong:
Clang tidy only provides a way to NOLINT individual lines, so there is three options. I recommend option 2, but if it is taking to long then I would be okay with option 3 and create an issue to fix it later since this is a utility.

1. The first is add the `// NOLINT` to each issue. 
2. The second is fix the issue, which is usually the same amount of work as adding the `// NOLINT` to each line.
3. Comment out the `target_clang_tidy` in the CMakeLists.txt for this target.

rjoomen:
Thanks for the suggestions, I tried option:

4. A line filter to exclude VHACD.h

But unfortunately, it doesn't seem to work, even though I found [an issue](microsoft/clang-tools-extra#22), mentioning exactly this. Do you know why this is?

Otherwise, I'll go for option 3, as with 654 warnings options 1&2 are a lot of work.

rjoomen:
@Levi-Armstrong friendly ping.

Levi-Armstrong:
I am not sure why it is not working, but I would just comment out the target_clang_tidy line in the CMakeLists.txt file.

Levi-Armstrong:
Thanks for making the update.

PR #903, opened 2023-06-14
Removed gcc-specific options from clang config

rjoomen:


Levi-Armstrong:
It is not clear to me why these should or need to be remove?

rjoomen:
These two removed warning options are for gcc, and do not exist for clang, so they actually produced extra warnings in clangd for me.

Levi-Armstrong:
Ah, looking at the diff I did not see there were different options for clang and gcc.

Levi-Armstrong (approved):

PR #905, opened 2023-06-15
WIP: Add a 22.04 (jammy) workflow

marrts:
Recently had a [failed pipeline](https://github.com/tesseract-robotics/tesseract_ros2/actions/runs/5280629534/jobs/9553023345?pr=47) on tesseract_ros2 that appears to be specifically related to a change that occurred from 20.04 to 22.04 causing an issue in tesseract_planning. tesseract and tesseract_planning should both be building on 22.04 without issues and this will ensure that.

### DIFF
--- a/.github/workflows/ubuntu_focal.yml
+++ b/.github/workflows/ubuntu_focal.yml
@@ -1,0 +1,109 @@
+name: Ubuntu-Focal
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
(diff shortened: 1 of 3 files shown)


marrts:
I did not realize this is still using some ROS stuff under the hood for CI which doesn't exist in 22.04 because there is no ROS, only ROS 2 in 22.04.

marip8:
Closing in favor of #912

PR #911, opened 2023-07-07
Reduce CI-built docker image size

marip8:
The `tesseract` docker images that are produced in CI are fairly large (4-7 Gb), mainly due to the artifacts of the build that are not deleted. This PR adds a final step to the CI to remove the `build` directories of the target and upstream workspaces to reduce the size of the saved docker image


marip8:
I think we could probably also uninstall all `-dev` libraries as well before the test stage to further reduce. Playing around with the docker image manually showed that reduced the size of the `/usr` directory from ~3.3Gb to ~2.2Gb

Levi-Armstrong:
Can't you also remove the source?

marip8:
We could remove the source too, but it's only ~180Mb (I think due to a clone depth of 1), so I didn't think it was worth it. I can add it if you want though

Levi-Armstrong:
> We could remove the source too, but it's only ~180Mb (I think due to a clone depth of 1), so I didn't think it was worth it. I can add it if you want though

Nah, just thinking of other things that may be able to reduce it.

What about the following commands? These are commonly used when generating dockers, but never looked into how much space they save.
```
apt-get clean && rm -rf /var/lib/apt/lists/*
```

Levi-Armstrong (approved):

PR #915, opened 2023-07-20
Draft: Add basic planar joint support in the urdf parser

mpowelson:
Levi and I discussed how to implement this. The better solution would probably be to implement planar joints in the state solver. However, that would touch a lot of things. Instead we decided to get an MVP working by inserting 3 joints for each planar joint in the URDF. They are named `<urdf_name>_<x, y, or yaw>_planar`. Then the ROS environment monitor will listen to TF and update these TFs as appropriate


tesseract_urdf/src/urdf_parser.cpp
```diff
@@ -282,4 +275,107 @@ void writeURDFFile(const tesseract_scene_graph::SceneGraph::ConstPtr& sg,
...
+    }
+    case JointType::FLOATING: {
+      std::throw_with_nested(std::runtime_error("FLOATING joints are not yet supported"));
+    }
+    default:
+      return { {}, { joint } };
+  }
+}
+
+void addLink(tesseract_scene_graph::SceneGraph::UPtr& sg,
```
Levi-Armstrong, line 351: I would probably have this pass the scenegraph by reference.

Levi-Armstrong:
I been looking at how MoveIt accomplishes this and think it aligns with what we would like to do. In this case add the concepts of variables where a joint can have multiple variables. A good point of reference is the joint model [here](https://github.com/ros-planning/moveit/blob/1b351dddd45c1524621702cee63f1ecb94fac9e0/moveit_core/robot_model/include/moveit/robot_model/joint_model.h).

PR #916, opened 2023-07-21
Added IKFast factory boilerplate

rjoomen:
This adds IKFast boilerplate code [as suggested by](https://github.com/tesseract-robotics/tesseract/discussions/777#discussioncomment-2956246) @marip8.


Levi-Armstrong:
Just so this code gets tested could we updated the ikfast test plugins to utilize this?

rjoomen:
Yes, that's a good idea. And the docs could then be simplified a lot (also, there's a number of errors/typos in the [IKFast solver/plugin section](https://tesseract-docs.readthedocs.io/en/latest/_source/core/packages/tesseract_kinematics_doc.html?highlight=ikfast#create-tesseract-ikfast-solver), but those would be gone then).

rjoomen:
BTW, I also had the idea to enhance the IKFast plugin boilerplate to use a discretization 'resolution' parameter for each free joint (making use of the joint limits from the URDF) instead of the list of exact joint values that is required currently. Because if you have for example a free joint with a 210 degree range and want to sample it in one degree steps, it requires a long and impractical list of values (I tested this).

Levi-Armstrong:
Sounds good to me.

rjoomen:
Would you mind to already merge, and I'll add the free joint enhancements later?

Levi-Armstrong:
Waiting on the joint enhancement is good, but I would like to update the unit tests to leverage this so this code is unit tested before merging.

rjoomen:
I updated the unit tests.

tesseract_kinematics/core/include/tesseract_kinematics/core/kinematic_group.h
```diff
@@ -50,8 +50,8 @@ struct KinGroupIKInput
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
   // LCOV_EXCL_STOP
 
-  KinGroupIKInput(const Eigen::Isometry3d& p, std::string wf, std::string tl)
-    : pose(p), working_frame(std::move(wf)), tip_link_name(std::move(tl))
+  KinGroupIKInput(Eigen::Isometry3d p, std::string wf, std::string tl)
```
Levi-Armstrong, line 53: Eigen types have to be passed by reference. If clang-tidy is complaining then add a nolint for the error it is giveing.

Levi-Armstrong: If they every updated to 3.4 this will not be the case but currently is a limitation of  versions < 3.4.

rjoomen: Reverted the change. I should've left that file alone 🙄

Levi-Armstrong:
It looks like clang-tidy is not happy

tesseract_kinematics/ikfast/include/tesseract_kinematics/ikfast/ikfast_factory_boilerplate.h
```diff
@@ -0,0 +1,118 @@
+/**
+ * @file ikfast_factory.h
```
marrts, line 2: The filename in the header comment is wrong
```suggestion
 * @file ikfast_factory_boilerplate.h
```

Levi-Armstrong:
> @rjoomen I think this [code](https://github.com/tesseract-robotics/tesseract/blob/663df9fa8aea99de124f115b96b68c8037e33e8e/tesseract_kinematics/ikfast/include/tesseract_kinematics/ikfast/external/ikfast.h#L290C1-L359) can be removed and should solve the clang-tidy issue.

Actually you may be able to just remove the -DIKFAST_HAS_LIBRARY from [here](https://github.com/tesseract-robotics/tesseract/blob/663df9fa8aea99de124f115b96b68c8037e33e8e/tesseract_kinematics/ikfast/CMakeLists.txt#L4).

rjoomen:
Now all that's left are clang-tidy issues in the solvers and external/ikfast.h. Shouldn't these have been suppressed by TESSERACT_COMMON_IGNORE_WARNINGS_PUSH?

Levi-Armstrong:
I just pushed a commit which should fix the clang-tidy issue which was tested locally.

rjoomen:
Great, thanks!

Levi-Armstrong:
@rjoomen Hey I am working on improving unit test coverage and not sure if we currently have a way to test ikfast free joints? #1140

rjoomen:
I'm sorry, I don't know. And I'm not using IKfast anymore, as I have a 7-DOF robot and am using Trac-IK for that.

PR #928, opened 2023-08-21
Add support for KDL::ChainIkSolverPos_NR_JL, which takes joint limits…

rjoomen:
… into account


tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_inv_kin_chain_nr_jl.h
```diff
@@ -0,0 +1,120 @@
...
+  std::vector<std::string> getJointNames() const override final;
+  Eigen::Index numJoints() const override final;
+  std::string getBaseLinkName() const override final;
+  std::string getWorkingFrame() const override final;
+  std::vector<std::string> getTipLinkNames() const override final;
+  std::string getSolverName() const override final;
+  InverseKinematics::UPtr clone() const override final;
+
+private:
+  KDL::JntArray q_min_;                                            /**< @brief Minimum joint positions */
```
Levi-Armstrong, line 104: Is there a reason not to add `q_min` and `q_max` to the KDLChainData structure?

Levi-Armstrong (changes requested):
After going through this and seeing how the q_min and q_max are used, I would recommend updating KDLChainData to automatically parse this data. Then add a option to current `KDLInvKinChainNR` to enable limits to avoid a new class being introduce. Let me know if you see any issues with this request.

Levi-Armstrong:
> Reason for this being a new class is that ChainIkSolverPos_NR_JL is a different solver, just like ChainIkSolverPos_LMA. On the other hand it is just a variation upon _NR, so adding use_limits as a parameter to differentiate between ChainIkSolverPos_NR and ChainIkSolverPos_NR_JL would also make sense. It's also easy to implement, so let me know if this is indeed how you'd like it and I'll modify the PR.

Yea, lets do that. It will limit the additional unit tests needed.

rjoomen:
I added the joint limits to the KDLChainData.

Upon reconsideration I'm not sure about removing ChainIkSolverPos_NR_JL and adding it as an option to ChainIkSolverPos_NR. In KDL and hence MoveIt these are separate plugins, with ChainIkSolverPos_NR_JL being the one that is used much more often. In my testing, ChainIkSolverPos_NR_JL also performed a lot better. To prevent confusion for people familiar with KDL or MoveIt I'd suggest to keep the current naming (NR and NR_JL).

If you still are in favor of adding JL as an option to ChainIkSolverPos_NR, I think it would be good to make using joint limits the default, or making joint limits an option without a default, forcing users to make a deliberate choice. The first case might silently break/improve existing setups, the second would throw errors in existing setups, forcing users to add the option. Leaving the default at no joint limits would set the inferior solver as default, though, which I'd consider undesirable.

tesseract_kinematics/kdl/include/tesseract_kinematics/kdl/kdl_utils.h
```diff
@@ -100,6 +100,8 @@ struct KDLChainData
   std::string tip_link_name;                /**< @brief Link name of last kink in the kinematic object */
   std::map<std::string, int> segment_index; /**< @brief A map from chain link name to kdl chain segment number */
   std::vector<std::pair<std::string, std::string>> chains; /**< The chains used to create the object */
+  KDL::JntArray q_min;                                     /**< @brief Lower joint limits */
+  KDL::JntArray q_max;                                     /**< @brief Upper joint limits */
```
Levi-Armstrong, line 104: Can we move this into the kdl data structure and parse it during that step?

Levi-Armstrong:
As long as it is unit tested, I don't have a preference.

Levi-Armstrong:
Thank you, the unit test coverage is the same as the others.

PR #931, opened 2023-08-31
CI build on branches starting with 'dev**'

johnwason:
This PR adds triggers so that branches starting with "dev/" will be built on GitHub actions. This is a convenience when developing on a branch before a pull request is started.

### DIFF
--- a/.github/workflows/clang_format.yml
+++ b/.github/workflows/clang_format.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - 'dev**'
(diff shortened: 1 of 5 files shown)


Levi-Armstrong:
I think we should just remove the branches because this rep only has master.

johnwason:
If you build for all branches a pull request might build twice when you push to a branch that has an open pull request, once for the push and once for the pull request. The solution here is you push to dev/ initially and then to a branch without the prefix for the pull request.

PR #935, opened 2023-09-06
Fix Ubuntu Jammy release build

Levi-Armstrong:


.github/workflows/ubuntu.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           then
             echo "BUILD_TYPE=Release" >> $GITHUB_ENV
           else
-            echo "BUILD_TYPE=Debug" >> $GITHUB_ENV
+            echo "BUILD_TYPE=Release" >> $GITHUB_ENV
```
rjoomen, line 87: I assume this is temporary? Release vs. Release?

Levi-Armstrong: Yes

PR #941, opened 2023-09-22
Find collision components optionally for environment test suite

marip8:
Changes per commit message


Levi-Armstrong:
I really do not like this for the unit tests, because it is to easy for this to get broken you will never know that the tests are not being ran as part of the pipeline. Why is this needed?

marip8:
> Why is this needed?

I almost never build FCL locally, so when I also build unit tests locally this causes my whole build to fail

Levi-Armstrong:
I understand but I would rather this not be default behavior for unit tests.

PR #943, opened 2023-10-04
Make `free_joint_state` parsing more generalized

marrts:
Fixed an issue where the IKFast plugin would fail to parse your `free_joint_state` list if it looked like this:
```
free_joint_states:
   - [0.0]
   - [-0.1]
   - [0.1]
   - [-0.2]
   - [0.2]
```

### DIFF
--- a/tesseract_kinematics/ikfast/include/tesseract_kinematics/ikfast/ikfast_factory_boilerplate.h
+++ b/tesseract_kinematics/ikfast/include/tesseract_kinematics/ikfast/ikfast_factory_boilerplate.h
@@ -70,10 +70,10 @@ public:
       // Get the free joint states
       if (YAML::Node free_joint_states_node = config["free_joint_states"])
       {
-        for (auto it = free_joint_states_node.begin(); it != free_joint_states_node.end(); ++it)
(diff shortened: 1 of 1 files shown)


rjoomen (approved):

Levi-Armstrong:
Does this make sense, because this is allowing something that should a single value to be entered as a vector?

rjoomen:
It is my understanding it works like this:
One free joint:
```
free_joint_states:
   - [0.0]
   - [-0.1]
   - [0.1]
   - [-0.2]
   - [0.2]
```
Two free joints:
```
free_joint_states:
   - [0.0, 0.0]
   - [-0.1, 0.0]
   - [0.1, 0.1]
   - [-0.2, 0.1]
   - [0.2, 0.2]
```

But before this PR one had to add indices:
```
free_joint_states:
   1: [0.0]
   2: [-0.1]
   3: [0.1]
   4: [-0.2]
   5: [0.2]
```
Etc.

Am I right, @marrts?

marrts:
> Am I right, @marrts?

Yes, that is exactly right. Now it works with either numbers or dashes.

Levi-Armstrong:
Thanks for the clarification. LGTM

PR #946, opened 2023-10-11
Unused includes cleanup

rjoomen:
Used clangd to check all files for unused includes

### DIFF
--- a/tesseract_collision/bullet/src/create_convex_hull.cpp
+++ b/tesseract_collision/bullet/src/create_convex_hull.cpp
@@ -27,6 +27,7 @@
 TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <console_bridge/console.h>
 #include <boost/program_options.hpp>
+#include <iostream>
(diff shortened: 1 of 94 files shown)


Levi-Armstrong:
Could we add a CI run which does this check?

rjoomen:
clang-tidy has [misc/include-cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html), but this is [very recent](https://reviews.llvm.org/rGc28506ba4b69). Our CI has all misc checks [already switched on](https://github.com/tesseract-robotics/tesseract/blob/master/.clang-tidy#L21), but the clang-tidy used is too old.

RICB has IWYU [functionality](https://github.com/ros-industrial/ros_industrial_cmake_boilerplate#include-what-you-use-iwyu), that could be an option. I've tried that once locally, I believe, but could not get it working. Maybe you're able add it to our CI?

Levi-Armstrong (approved):
LGTM, good to merge when CI passes.

PR #954, opened 2023-11-05
Fix tesseract collision when calculate distance is disabled. 

Levi-Armstrong:


Levi-Armstrong:
@johnwason The latest release of yaml-cpp is broken, how do I update the windows CI to have it pull a specific version for this package?

johnwason:
@Levi-Armstrong here is the fix to the issue https://github.com/jbeder/yaml-cpp/pull/1242

I think the solution is to check to see if yaml-cpp target already exists. This is probably going to be a problem in other places as well since the error made it into a release version. Here is an example for jsoncpp with the same issue (that still hasn't been fixed).

https://github.com/tesseract-robotics/trajopt/blob/2d965b931015fba9a83ca697e114e3c07ef7c25b/trajopt/CMakeLists.txt#L20

johnwason:
Oh I didn't notice that you created that pull request...

Levi-Armstrong:
@johnwason for now how can I set the version for yaml-cpp in your vcpkg action for this package?

johnwason:
You can use the `revision` option for vcpkg-action to select a git version for the vcpkg repository. I don't see an easy way to set the package version using the current vcpkg command. I can create a manifest in the future so we can select specific version of packages.

https://github.com/johnwason/vcpkg-action

PR #961, opened 2023-11-10
Remove use of Industrial CI

Levi-Armstrong:


.github/workflows/ubuntu_test.yml
```diff
@@ -0,0 +1,91 @@
...
+        shell: bash
+        run: |
+          apt update
+          apt upgrade -y
+          apt install -y cmake curl lsb-release ca-certificates gnupg python3 python3-distutils python3-pip ccache git
+          python3 -m pip install vcstool -q
+          python3 -m pip install colcon-common-extensions -q
+          cd $GITHUB_WORKSPACE/target_ws/src
+          ./.github/workflows/add_ros_apt_sources.sh
+          apt install -y liboctomap-dev libfcl-dev libconsole-bridge-dev libbullet-dev libbullet-extras-dev libyaml-cpp-dev libtinyxml2-dev libassimp-dev liborocos-kdl-dev libpcl-dev liblapack-dev libboost-dev libboost-filesystem-dev libboost-serialization-dev libboost-program-options-dev libboost-graph-dev libbenchmark-dev libccd-dev libgtest-dev
```
marip8, line 49: Any reason to not use `rosdep` here? Seems like it would be a lot more convenient and less error-prone

marip8:
Have you thought about turning this into a simple Github action instead? Might be easier to apply to other tesseract repos in future as well

.github/workflows/ubuntu_test.yml
Levi-Armstrong: It looks like rodep is a python tool so that should be fine.

Levi-Armstrong:
> Have you thought about turning this into a custom Github action instead? Might be easier to apply to other tesseract repos in future as well

Yea, that is not a bad idea. Have you ever created one?

marip8:
No I haven't, but I don't think it's too difficult. I think it can be as simple as a YAML file defining what steps should run as a part of the custom action and what the input variables can be. You can also have it call custom scripts in the action repo, etc.

marip8:
It might make sense to do make a similar custom Github action for building Docker images as well, since the Docker files for `tesseract` through `tesseract_ros2` are not all that dissimilar from one another

PR #963, opened 2023-11-13
Use colcon-action

Levi-Armstrong:


marip8:
We might consider moving the scripts `add_ros_apt_sources.sh` and `add_sources.sh` (from `tesseract_planning`) to this new `colcon-action` repo and adding optional steps in the build chain

Levi-Armstrong:
> We might consider moving the scripts `add_ros_apt_sources.sh` and `add_sources.sh` (from `tesseract_planning`) to this new `colcon-action` repo and adding optional steps in the build chain. We'll need to run those scripts in order to install ROS tooling libraries

Add a new option `ros-enabled` to the action.

PR #968, opened 2023-11-24
Added application for performing convex decomposition

marip8:
Addition per commit message


rjoomen:
Didn't test it, but LGTM. Essentially this is a variant of [TestVHACD](https://github.com/kmammou/v-hacd/blob/master/app/TestVHACD.cpp), you could consider to use the same parameter range checks (and possibly the same parameter format?).

Levi-Armstrong:
@marip8 Just need to fix clang-tidy issues and this can be merged.

marip8:
> @marip8 Just need to fix clang-tidy issues and this can be merged.

Looks like all the `clang-tidy` issues come from `VHACD.h`. How do we ignore errors from this file?

rjoomen:
> Looks like all the `clang-tidy` issues come from `VHACD.h`. How do we ignore errors from this file?

Seems `TESSERACT_COMMON_IGNORE_WARNINGS_PUSH` does not work for newer clang-tidy. You could add [`// NOLINTBEGIN`](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics) just after `TESSERACT_COMMON_IGNORE_WARNINGS_PUSH` and and `// NOLINTEND` just before `TESSERACT_COMMON_IGNORE_WARNINGS_POP`. That solves it for me.

rjoomen:
It seems adding NOLINTBEGIN and END to the macros does not work. I suggested adding them to VHACD.h, does that work?

PR #971, opened 2023-12-08
Add jacobian transform error calculation function

Levi-Armstrong:


marrts:
This works for me, should probably duplicate or make a slightly modified version of the unit tests for `calcTransformError` in `tesseract_common_unit.cpp` https://github.com/tesseract-robotics/tesseract/blob/985e4c0b99782de5be29481a262debe8ec1cf3e1/tesseract_common/test/tesseract_common_unit.cpp#L2114

PR #981, opened 2024-01-16
Fix conda tests

johnwason:
Fixes the conda ci tests.


johnwason:
Some tests were filtered but the conda tests passed.

Levi-Armstrong (approved):

PR #986, opened 2024-02-06
Add hpp-fcl as a collision checker

rjoomen:
This adds [HPP-FCL](https://github.com/humanoid-path-planner/hpp-fcl), the improved fork of [FCL](https://github.com/flexible-collision-library/fcl), as a collision checker.

Notable features:
- (Claimed) Improved performance
- [State of the art GJK collision checking](https://arxiv.org/abs/2205.09663)
- Safety margins, lower bound distance calculation, GJK early stopping (can we use any of these?)
- Actively developed

Most of this work was simply renaming, though a few types and methods differed slightly. By directly comparing the `fcl_\*.\*` and `hpp_fcl_\*.\*` source files one can easily check the required modifications.

Preliminary benchmarking looks promising.

Status: currently these unit tests fail:

Tests are run using the `master` and `devel` branches of hpp-fcl.

In the table below, _master 1_ uses the [current workaround](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/fcl/src/fcl_utils.cpp#L318) for distance contact normals, and _master 2_ uses `fcl_result.normal`. For _devel_ there is no difference in test success between the two approaches for normal calculation. This suggests that [this closed issue](https://github.com/humanoid-path-planner/hpp-fcl/issues/307), regarding normals and nearest points solves the workaround as currently used in the FCL plugin.

| Test | master 1 | master 2 | devel |
| :--- | --- | --- | --- |
| HPP_FCLDiscreteBVHCollisionBoxSphereUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionBoxSphereConvexHullUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionBoxCylinderUnit | | x | |
| HPP_FCLDiscreteBVHCollisionBoxConeUnit | | x | |
| HPP_FCLDiscreteBVHCollisionBoxBoxUnit | x | x | x |
| HPP_FCLDiscreteBVHCollisionBoxBoxConvexHullUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionBoxCapsuleUnit | x | x | x |
| HPP_FCLDiscreteBVHCollisionLargeDataSetConvexHullUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionSphereSphereUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionSphereSphereConvexHullUnit | x | | |
| HPP_FCLDiscreteBVHCollisionMeshMeshUnit | | x | x |
| HPP_FCLDiscreteBVHCollisionMultiThreadedConvexHullUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionOctomapSphereUnit | | | x |
| HPP_FCLDiscreteBVHCollisionOctomapSphereConvexHullUnit | x | x | x |
| HPP_FCLDiscreteBVHCollisionOctomapSphereMeshUnit* | x | x | x |
| HPP_FCLDiscreteBVHCollisionCloneUnit | | x | |
| HPP_FCLDiscreteBVHContactManagerConfigUnit | x | x | |
| HPP_FCLDiscreteBVHCollisionCompoundCompoundUnit | | | x |
| Sum | 12 | 16 | 7 |

\* This test was disabled for FCL already

Interesting to note is that multiple runs of the tests results in a varying number of failed tests (the table lists the maximum number found). Perhaps some of the tests are at the edge of collision and the numerical optimization sometimes finds a collision and sometimes not.


rjoomen:
There seems to be a CI problem with `rosdep`:
```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies (ROS distro is not set. Make sure `ROS_DISTRO` environment variable is set, or use `--rosdistro` option to specify the distro, e.g. `--rosdistro indigo`):
tesseract_collision: Cannot locate rosdep definition for [hpp-fcl]
```
If I test this locally `hpp-fcl` is indeed not found if I unset the `ROS_DISTRO` environment variable.

dependencies.repos
```diff
@@ -10,6 +10,10 @@
     local-name: fcl
     uri: https://github.com/flexible-collision-library/fcl.git
     version: 0.6.1
+- git:
+    local-name: hpp-fcl
+    uri: https://github.com/humanoid-path-planner/hpp-fcl.git
+    version: v2.4.1
```
Levi-Armstrong, line 16: Can we update this to point to the latest hash to see if the unit tests will pass?

Levi-Armstrong:
> Safety margins, lower bound distance calculation, GJK early stopping (can we use any of these?)

Depending on how it is implemented. In had to hack the safety margin piece in by creating the `FCLCollisionObjectWrapper` which inherits from `fcl::CollisionObject<double>` to expand the AABB for BVH.


What does this mean ` GJK early stopping (can we use any of these?)`?

rjoomen:
> What does this mean ` GJK early stopping`?

See [here](https://github.com/humanoid-path-planner/hpp-fcl/blob/master/include/hpp/fcl/narrowphase/narrowphase.h#L520), [here](https://github.com/humanoid-path-planner/hpp-fcl/blob/master/include/hpp/fcl/collision_data.h#L263), and [here](https://github.com/humanoid-path-planner/hpp-fcl/pull/371).

(And `(can we use any of these?)` referred to whether any of the listed functions (safety margins, lower bound, early stopping) might be of use for Tesseract.)

dependencies.repos
rjoomen: Yes, but I wanted to test this locally first.

Levi-Armstrong:
Yes, we would set this to the contact threshold normal collision checking and in trajopts case it would be set to safety_margin + safety_margin_buffer.

rjoomen:
Tested with master: same result (as expected, v2.4.1 is very recent).

dependencies.repos
rjoomen: I've done more extensive testing, see the table in the first post. I've also updated the PR to use the `devel` branch.

rjoomen:
> Not sure what to do here with so many tests failing. I know between FCL and Bullet there were slight numerical difference; have you looked at the tests to see if that is the case here?

I've not looked in detail yet. But the hpp-fcl devel branch is under heavy development (some 100 commits since I started this PR), so I'd like to wait for v3 to be released before continuing this PR. I'm also mostly using cast collision checking, so for me it's not urgent anyway.

Levi-Armstrong:
Do you want me to create a separate repository for this similar to the newly created tesseract_collision_physx?

rjoomen:
> Do you want me to create a separate repository for this similar to the newly created tesseract_collision_physx?

That's a good idea, please do.

Levi-Armstrong:
@rjoomen What name do you prefer for the repository: `tesseract_collision_hpp_fcl`, `tesseract_collision_hppfcl` or `...` ?

rjoomen:
I don't really mind, but as the subproject is called `hpp_fcl`, maybe `tesseract_collision_hpp_fcl` is the most logical.

Levi-Armstrong:
@rjoomen The repository has been created. Let me know if you run into any access issues. 

https://github.com/tesseract-robotics/tesseract_collision_hpp_fcl

PR #989, opened 2024-02-20
Trajectory player fixes

rjoomen:
- Fix trajectory player for trajectories that do not start from time=0, e.g. when in a trajectory set
- Fix trajectory interpolator for size=1 trajectories

This is related to https://github.com/tesseract-robotics/tesseract_qt/pull/99.

### DIFF
--- a/tesseract_visualization/test/trajectory_player_unit.cpp
+++ b/tesseract_visualization/test/trajectory_player_unit.cpp
@@ -40,7 +40,7 @@ void CheckTrajectory(const tesseract_common::JointTrajectory& trajectory, int fi
   TrajectoryPlayer player;
   player.setTrajectory(trajectory);
 
-  EXPECT_NEAR(player.trajectoryDurationStart(), first, 1e-5);
+  EXPECT_NEAR(player.trajectoryDurationBegin(), first, 1e-5);
   EXPECT_NEAR(player.trajectoryDurationEnd(), last, 1e-5);
   EXPECT_NEAR(player.currentDuration(), first, 1e-5);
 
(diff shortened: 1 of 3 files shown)


marrts:
This is working for me and is definitely a nice addition, just have a couple of notes:
1. I'm confused as to what you are accomplishing with the "Fix trajectory interpolator for size=1 trajectories" I see you added a check for size being 1, but I don't know what the behavior change is.
2. It looks like this will also require a PR on tesseract_planning as [one of the examples](https://github.com/tesseract-robotics/tesseract_planning/blob/c791e7488de19c4f3a5dfc2eccf2624e415aa427/tesseract_examples/src/online_planning_example.cpp#L318) points to `trajectoryDuration()` method

rjoomen:
1) When a trajectory has only one waypoint, the TrajectoryInterpolator would incorrectly set the initial state and total time to 0, leading to the trajectory slider in the workbench showing only zeros. With this fix, if for example a trajectory set has one trajectory with 10 points with times [0...10] and a second set with only one point with time=11 the slider will correctly show 11 for the second set (and 0-10 for the first). Obviously there is nothing to slide, but at least the numbers shown are correct. Trajectories of size one are not very useful, of course, but sometimes my planner receives this as input, as only the start is reachable (IK), but not the goal.
2) See https://github.com/tesseract-robotics/tesseract_planning/pull/441

tesseract_visualization/test/trajectory_player_unit.cpp
```diff
@@ -52,7 +52,7 @@ TEST(TesseracTrajectoryPlayerUnit, TrajectoryTest)  // NOLINT
   TrajectoryPlayer player;
   player.setTrajectory(trajectory);
 
-  EXPECT_NEAR(player.trajectoryDuration(), 9, 1e-5);
```
Levi-Armstrong, line 55: Probably be good to a test for the trajectoryDurationStart() method.

Levi-Armstrong (changes requested):

tesseract_visualization/test/trajectory_player_unit.cpp
rjoomen: That's a good one. Will do tomorrow

Levi-Armstrong (approved):

PR #1001, opened 2024-05-28
Add support for jerk limits

Levi-Armstrong:


Levi-Armstrong:
@johnwason Do you know what the issue is with the conda builds?

johnwason:
I don't see the obvious reason for the failure. I will have to look into this.

PR #1002, opened 2024-05-29
Add binary data serialization support functions

Levi-Armstrong:


Levi-Armstrong:
@marip8, @marrts and @rjoomen These new utility function will allow us to update the ros messages to send binary data instead of xml string which is significantly smaller payload.

marrts:
I assume that also mean save to files would be a lot smaller? That would be nice.

Levi-Armstrong:
> I assume that also mean save to files would be a lot smaller? That would be nice.

Yea, those utilities already exist and are significantly smaller when saving to file.

PR #1013, opened 2024-06-08
Fix command serialization

Levi-Armstrong:
There are 656 lines of serialization code within tesseract and these two line had the incorrect syntax.

### DIFF
--- a/tesseract_environment/src/command.cpp
+++ b/tesseract_environment/src/command.cpp
@@ -40,14 +40,14 @@ template <class Archive>
 void save(Archive& ar, const CommandType& g, const unsigned int /*version*/)
 {
   int value = static_cast<int>(g);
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
@johnwason Does this fix the issue you were seeing?

johnwason:
Still seeing the crashes and test failures

Levi-Armstrong:
> Still seeing the crashes and test failures
Does the example you posted on boost also still fail?

johnwason:
I am looking at this old test case: https://github.com/johnwason/tesseract/actions/runs/8242279639

I need to find the commit locally. I will change the `&=` to `&` in this known test case and see if it makes any difference. There are too many changes in master to isolate the problem.

johnwason:
I didn't realize that some of the `&=` were in the equals functions... I changed those back and my test case is now failing.

PR #1015, opened 2024-06-19
Updated Dockerfile to support 24.04

marip8:
Adds changes similar to #1006 to support Docker image build on 24.04

PR #1006, opened 2024-05-30
Add CI for Ubuntu Noble

marip8:

PR #1016, opened 2024-06-23
Fixes for building on Ubuntu Noble

rjoomen:
Pending a rebase after #1006 and #1015 have been merged.


Levi-Armstrong:
@rjoomen Ping me when this ready. I assume you are working on changes to get the CI to pass?

rjoomen:
Yes, there's a test failing and I didn't have time to look into it. Will have time tomorrow, I hope.

cboostjvisser:
I have been investigating the failing tests. I have not been able to pinpoint the exact problem (yet). 
But here are my findings:

### The failing tests are: 

- https://github.com/tesseract-robotics/tesseract/blob/2a92fa8db2953591976fc91062aa3a6e31847838/tesseract_common/test/plugin_loader_unit.cpp#L85
- https://github.com/tesseract-robotics/tesseract/blob/2a92fa8db2953591976fc91062aa3a6e31847838/tesseract_common/test/plugin_loader_unit.cpp#L155

There are already comments about failing on Ubuntu 18.04 and a `#if BOOST_VERSION > 106800 && !__APPLE__` surrounding the tests.

### Boost investigation:

- Ubuntu 22.04(Jammy) uses Boost 1.74
- Ubuntu 24.04(Noble) uses Boost 1.83

- Downgrading just [Boost.DLL](https://github.com/boostorg/dll)(library module) from 1.83 to 1.74 does not help. 
- Downgrading whole the Ubuntu Boost installation from 1.83 to 1.74 does fix the tests. 

`dlopen()` seems to be called with exact the same arguments. But does not return handle in 1.83. See here:
https://github.com/boostorg/dll/blob/6c60dde50bf67138c90cc84938111866813feaff/include/boost/dll/detail/posix/shared_library_impl.hpp#L118

### Tests might not be correct?
I'm also questioning if the tests are correct. 
If I read this post: https://stackoverflow.com/a/17716576
It seems like we need to load a local library with "./" leading the libname, or else `DT_RPATH or DT_RUNPATH` needs to be set. 

The current `tesseract_common/class_loader.hpp` always sets `boost::dll::load_mode::search_system_folders` if there is no library_path specified. 
For example:
https://github.com/tesseract-robotics/tesseract/blob/2a92fa8db2953591976fc91062aa3a6e31847838/tesseract_common/include/tesseract_common/class_loader.hpp#L92
That will trigger the "./" preprending to be disabled in the Boost.DLL library: https://github.com/boostorg/dll/blob/6c60dde50bf67138c90cc84938111866813feaff/include/boost/dll/detail/posix/shared_library_impl.hpp#L97

## Conclusion
Removing all the `boost::dll::load_mode::search_system_folders` in  `tesseract_common/class_loader.hpp` seem to fix the tests. But not sure what other problems that might cause. 
I was not able to find what causes the tests to work with Boost 1.74 but not with Boost 1.83.

rjoomen:
Additional information:

Changing lines 85 and 86 to:
```
EXPECT_TRUE(ClassLoader::isClassAvailable(symbol_name, lib_name, ".")); 
auto plugin = ClassLoader::createSharedInstance<TestPluginBase>(symbol_name, lib_name, ".");
```
And adding this before line 155:
```
plugin_loader.search_paths.insert(".");
```

Also fixes the tests.

rjoomen:
> The current `tesseract_common/class_loader.hpp` always sets `boost::dll::load_mode::search_system_folders` if there is no library_path specified. For example:
> 
> https://github.com/tesseract-robotics/tesseract/blob/2a92fa8db2953591976fc91062aa3a6e31847838/tesseract_common/include/tesseract_common/class_loader.hpp#L92
> 
> That will trigger the "./" preprending to be disabled in the Boost.DLL library: https://github.com/boostorg/dll/blob/6c60dde50bf67138c90cc84938111866813feaff/include/boost/dll/detail/posix/shared_library_impl.hpp#L97

### Conclusion

Because `class_loader` sets mode `search_system_folders`, the Tesseract class loaders by design do not use the local directory if no search path is given.

@Levi-Armstrong, should we modify the class loader or the tests to deal with this?

Levi-Armstrong:
I would update the unit test for now.

Levi-Armstrong (approved):

PR #1024, opened 2024-07-15
Update meta.yaml

Levi-Armstrong:


Levi-Armstrong:
@johnwason Would you have time to take a look and see what the issue is?

johnwason:
It looks like the backtrace package produces several different CMake boost components based on the available capabilities of the operating system. For instance, it will generate the `stacktrace_backtrace`, `stacktrace_noop`, and/or `stacktrace_basic` on different platforms. Conda Windows has `backtrace_noop`, while Conda Linux has `backtrace_basic`. None have `backtrace_stacktrace`. My recommendation is to search for `backtrace_noop` as a COMPONENT, and then add `backtrace_stacktrace` and `backtrace_basic` as `OPTIONAL_COMPONENTS` in `find_boost()`. Then select which one was found.

PR #1029, opened 2024-07-19
Add utility to check unknown keys in yaml node

Levi-Armstrong:


Levi-Armstrong:
@rjoomen I had to revert your change to the clang-format config because it was not compatible on 20.04.

PR #1034, opened 2024-07-25
Fix use of boost stacktrace

Levi-Armstrong:


Levi-Armstrong:
@johnwason Do you mind reviewing? This seems to have fixed the conda linux build but windows and macosx seem to have other issues.

PR #1039, opened 2024-07-30
Use --build-only for conda build in conda action to speed up test

johnwason:


johnwason:
> Does this not build the tests?

Conda has its own "tests" that are separate from ctest. The ctest is still run at the end of the build phase.

johnwason:
This is the script that is being run to build the package using conda:

https://github.com/tesseract-robotics/tesseract/blob/1e25fba2ebf27393809836510fe7562aca88a2ae/.github/workflows/conda/recipe/build.sh

Normally ctest would not be in this script. We are only using conda-build because that is the only way to get the correct environment configuration. We don't care about the packaging phase for this test. The packaging was extremely slow, taking around 15 minutes on Windows.

Levi-Armstrong (approved):

PR #1043, opened 2024-09-06
Make checkKin optional in debug

marrts:
If a manipulator group has a lot of joints the kinematic check can take a very long time. This optionally removes that check which was previously always run in debug mode.

### DIFF
--- a/tesseract_environment/include/tesseract_environment/environment.h
+++ b/tesseract_environment/include/tesseract_environment/environment.h
@@ -204,7 +204,8 @@ public:
    * @return A kinematics group
    */
   std::unique_ptr<tesseract_kinematics::KinematicGroup> getKinematicGroup(const std::string& group_name,
-                                                                          const std::string& ik_solver_name = "") const;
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
I do not think this is necessary because it should only run if built in debug mode. This should run if you build in release.

Levi-Armstrong:
Is this check running when built in release?

marrts:
> Is this check running when built in release?

No, but I was developing in debug with a high DOF system and it was taking 30+ seconds to run the check which was annoying. At first I thought something broke. 

I figured this way it wouldn't introduce any breaking changes but would just allow users to get around it if they need to test something else

Levi-Armstrong:
I would prefer not to make this type of change, but instead update the if statement to also check if TESSERACT_ENABLE_TESTING is enabled which should solve the issue you are having. These types of checks are intended to be ran during unit tests.

Levi-Armstrong:
Replaced by PR #1048

PR #1046, opened 2024-09-16
Add geometry type CompoundMesh

Levi-Armstrong:
It is possible that a single resource can contain multiple meshes and currently these are added as separate geometries to a link and the association is lost for the most part. It is possible to loop over all geometries for a given link and check the if the resource are the same but this is not ideal. This simplifies things so all meshes for a single resource will be stored in a CompoundMesh if more than one mesh exists.

This was issue because in rviz and tesseract_qt_studio if a visual mesh was a dae file with multiple meshes it would load the full mesh file for every submesh stored resulting in multiple copies of the same mesh. This was causing performance issues during rendering. Having the compound mesh solves issue because all meshes for a given resource are stored together.


Levi-Armstrong:
@johnwason Do you have any ideas why the conda builds cannot locate resources in tesseract_support?

johnwason:
Sorry I missed this question. Did you get it working?

PR #1079, opened 2024-12-10
Get a brief string summary of a contact map

marrts:
This can be useful for motion planning debugging.

### DIFF
--- a/tesseract_collision/core/include/tesseract_collision/core/types.h
+++ b/tesseract_collision/core/include/tesseract_collision/core/types.h
@@ -290,6 +290,47 @@ public:
   bool operator==(const ContactResultMap& rhs) const;
   bool operator!=(const ContactResultMap& rhs) const;
 
+  /** @brief Get a brief summary of the most frequently colliding link pair
+   * @return A string stream containing the collision summary
+   */
+  std::stringstream getCollisionSummary() const
+  {
+    std::stringstream ss;
+    std::map<KeyType, int> collision_counts;
+    std::map<KeyType, double> closest_distances;
+
+    // Initialize distances map with max values
+    for (const auto& pair : data_)
+    {
+      if (!pair.second.empty())
+      {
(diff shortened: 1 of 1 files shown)


tesseract_collision/core/include/tesseract_collision/core/types.h
```diff
@@ -290,6 +290,47 @@ class ContactResultMap
   bool operator==(const ContactResultMap& rhs) const;
   bool operator!=(const ContactResultMap& rhs) const;
 
+  /** @brief Get a brief summary of the most frequently colliding link pair
+   * @return A string stream containing the collision summary
+   */
+  std::stringstream getCollisionSummary() const
```
Levi-Armstrong, line 296: Why not just return a string?

Levi-Armstrong (changes requested):

tesseract_collision/core/include/tesseract_collision/core/types.h
Levi-Armstrong: @marrts I think we can just return a string here.

Levi-Armstrong (approved):

tesseract_collision/core/include/tesseract_collision/core/types.h
```diff
@@ -290,6 +290,46 @@ class ContactResultMap
   bool operator==(const ContactResultMap& rhs) const;
   bool operator!=(const ContactResultMap& rhs) const;
 
+  /** @brief Get a brief summary of the most frequently colliding link pair
+   * @return A string containing the collision summary
+   */
+  std::string getCollisionSummary() const
```
Levi-Armstrong, line 296: I think we also just name this `getSummary()`.

marrts: Good point, I don't know what other summary you'd expect from a `ContactResultMap`

Levi-Armstrong (approved):

PR #1080, opened 2024-12-11
Add pose vector for collision objects, as CompoundMesh shapes can have multiple collision objects.

### LINKED ISSUE #1053: CompoundMesh does not work with FCL
[CompoundMesh ](https://github.com/tesseract-robotics/tesseract/pull/1046) does not work with the FCL collision checker. I suppose it is because for Bullet, one CompoundMesh is added as one compound object, whereas for FCL, one CompoundMesh is added a multiple objects.

I've taken a look, it seems FCL does not support compound objects, apart from [BHVModel](https://github.com/flexible-collision-library/fcl/blob/master/include/fcl/geometry/bvh/BVH_model.h), which only supports meshes or pointclouds as submodels.

And for reference, neither does HPP-FCL: [#340](https://github.com/humanoid-path-planner/hpp-fcl/issues/340), [#448](https://github.com/humanoid-path-planner/hpp-fcl/issues/448).

rjoomen:
Fixes #1053

### DIFF
--- a/tesseract_collision/fcl/include/tesseract_collision/fcl/fcl_utils.h
+++ b/tesseract_collision/fcl/include/tesseract_collision/fcl/fcl_utils.h
@@ -116,7 +116,7 @@ public:
     for (unsigned i = 0; i < collision_objects_.size(); ++i)
     {
       CollisionObjectPtr& co = collision_objects_[i];
-      co->setTransform(pose * shape_poses_[i]);
+      co->setTransform(pose * collision_poses_[i]);
       co->updateAABB();  // This a tesseract function that updates abb to take into account contact distance
     }
   }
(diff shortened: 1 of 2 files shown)


Levi-Armstrong:
Ah, I now understand the issue. Do you have a test case that you have been using to add to the unit test frame work? I will take a look at this PR later this week. Thank you for tracking this down.

rjoomen:
> Ah, I now understand the issue. Do you have a test case that you have been using to add to the unit test frame work? I will take a look at this PR later this week. Thank you for tracking this down.

I don't have a test case, as I used my own production code. But I think adding a [setCollisionObjectsTransform](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/fcl/include/tesseract_collision/fcl/fcl_utils.h#L119) call for the CompoundMesh to the existing test case should to the trick.

rjoomen:
Replaced by #1083

Issue #1059, opened 2024-11-06
Consider replacing Tesseract URDF version property with XML namespace


marip8:
We should consider using a URDF namespace rather than a version tag for `tesseract`-specific URDF elements. URDF already has a version tag, so it's a bit confusing (and feels unnecessary) to specify an additional `tesseract_version` property.

Here's an example of what this might look like:

```xml
<robot name="my_robot" xmlns:tesseract="http://github.com/tesseract-robotics">
  ...
  <collision>
    <geometry>
      <tesseract:convex_mesh file="" convert=""/>
    </geometry>
  </collision>
  ...
</robot>
```

Levi-Armstrong:
@marip8 Do you have time to work on this? I think this might be good to get into the next release.

Levi-Armstrong:
@marip8 Can this be closed?

marip8:
Closing; addressed by #1081

PR #1081, opened 2024-12-13
Update URDF parser to use Tesseract XML namespace

marip8:
This PR addresses #1059 by removing the URDF tag `tesseract_version` and moving the custom geometry implementations to a `tesseract` XML namespace. This should allow us to make further Tesseract-specific URDF additions in a clearer manner. Users will need to specify this XML namespace as follows:

```diff
- <robot name="my_robot" xmlns:xacro="http://ros.org/wiki/xacro">
+ <robot name="my_robot" xmlns:xacro="http://ros.org/wiki/xacro" xmlns:tesseract="http://github.com/tesseract-robotics">
  ...
</robot>
```

I think the URL is somewhat arbitrary, but should generally point to a location that gives the user more information about what elements are available in this namespace.

This PR also removes the behavior in `tesseract_version` < 2 of automatically converting collision mesh files into into convex hulls. Now, if users want `tesseract` to automatically convert mesh files to convex hulls, they will need to add the attribute `tesseract:make_convex="true|false"` either at the global level (as an attribute in the `robot` element) or at the individual mesh level (i.e., as an attribute in the `mesh` element)

```diff
<collision>
  <geometry>
-    <mesh filename="" />  <!-- Previously, this mesh was automatically converted to a convex hull under the hood when `tesseract_version` was < 2 -->
+    <mesh filename="" tesseract:make_convex="true"/>  <!-- Now use this element to get the same behavior -->
  </geometry>
</collision>
```


Levi-Armstrong:
>This PR also removes the behavior in tesseract_version < 2 of automatically converting collision mesh files into into convex hulls. Now, if users want tesseract to automatically convert mesh files to convex hulls, they will need to use the <tesseract:convex_mesh file="" convert="true"/> element instead of just the mesh element.

I think more discussion is needed because this will have significant impact throughout Tesseract because most leverage xacro's outside of tesseract with no easy way to enable this convex hull functionality instead of duplicating and updating with the tesseract  namespace to get same behavior. I think there still needs to be a global way to enable this.

Levi-Armstrong:
It looks like you can namespace attributes so we could do the following to avoid conflict with the urdf version? Also could just remove the `convex_mesh` element and just leverage a namespace attribute in the mesh element `tesseact:convex_hull="false"` and get rid of the version number.

```
<robot name="my_robot" xmlns:xacro="http://ros.org/wiki/xacro" xmlns:tesseract="http://github.com/tesseract-robotics" tesseract:version="2">
```

marip8:
I like the idea of using namespace attributes where possible better than adding new elements. That way existing tools can still function (mostly) correctly without needing to understand Tesseract-specific attributes (e.g., Rviz would still render `<mesh/>` elements whereas it wouldn't know how to render `<tesseract:convex_mesh/>` elements).

Is the global attribute really that useful? I wonder if it would be more confusing rather than helpful, especially when you start thinking about composing xacros where one xacro might want it to be on and another off, but at the highest level there can only be one value of that attribute that gets applied to all underlying xacros. I would rather just specify it at the individual mesh level to be very explicit for clarity and consistency

Levi-Armstrong:
> Is the global attribute really that useful? I wonder if it would be more confusing rather than helpful, especially when you start thinking about composing xacros where one xacro might want it to be on and another off, but at the highest level there can only be one value of that attribute that gets applied to all underlying xacros. I would rather just specify it at the individual mesh level to be very explicit for clarity and consistency

While agree that is it better to be explicit, though I do not think there is an easy way around having a global attribute which does not add significant burden to the developer. This is not easy to accomplish if you using third party package like fanuc, motoman, etc., because without a global attribute you would have to copy all third party xacros and modify them when you want to use convex hulls. I think we can provide the global option but change the default to `false` so now the default behavior is to leverage detailed meshes and they must explicit set this to true to  get the current behavior.

marip8:
Yeah, I've run into a similar problem; I ended up running xacro to create a URDF of the fully defined system, then manually added all of the custom namespace elements/attributes that I needed, which is somewhat tedious. I'll add the global attribute and set the default to *not* convert mesh to convex hulls, and I'll add the mesh element specific override attributes.

tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
```diff
@@ -23,12 +23,19 @@ static std::string getTempPkgPath()
 TEST(TesseractURDFUnit, parse_mesh)  // NOLINT
 {
   tesseract_common::GeneralResourceLocator resource_locator;
+  const bool global_make_convex = false;
```
marip8, line 26: @Levi-Armstrong how do you want to handle unit tests for the `global_make_convex = true` case? I don't really want to copy-paste all the unit tests, but I don't know how many unit tests are worth making with the inverse case

tesseract_urdf/include/tesseract_urdf/box.h
```diff
@@ -41,12 +41,14 @@ class XMLDocument;
 
 namespace tesseract_urdf
 {
+extern const char* BOX_ELEMENT_NAME;
```
marip8, line 44: @Levi-Armstrong this seems to be a problem for windows. Do you know how to handle this for windows? If I don't make it `extern`, then I get the compiler `unused-variable` error

Levi-Armstrong: Why are these needed?

Levi-Armstrong: I would recommend create a single header like constants.h and leveraging `constexpr std::string_view BOX_ELEMENT_NAME = "box";`

tesseract_urdf/test/tesseract_urdf_collision_unit.cpp
```diff
@@ -13,6 +13,9 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 TEST(TesseractURDFUnit, parse_collision)  // NOLINT
 {
   tesseract_common::GeneralResourceLocator resource_locator;
+  const bool global_make_convex = false;
+  const auto parse_collision_fn =
+      std::bind(&tesseract_urdf::parseCollision, std::placeholders::_1, std::placeholders::_2, global_make_convex);
```
Levi-Armstrong, line 18: I believe clang-tidy is going to want you to replace this with a lambda.

Levi-Armstrong:
Also thinking we should just make the default true, because I think it is fine that we are different as long as there is ability to change the behavior. This would also require less changes possible throughout the repos.

tesseract_urdf/include/tesseract_urdf/box.h
marip8: > Why are these needed?

I thought it would be better to save the URDF element names into variables rather than typing the strings multiple times where they get used

marip8:
> Also thinking we should just make the default true, because I think it is fine that we are different as long as there is ability to change the behavior. This would also require less changes possible throughout the repos.

I would prefer to keep the default to be false (i.e., the global option is false when the attribute is not present) and make users specify that attribute explicitly. Automatically converting all mesh collision geometry to convex hulls without the users explicit instruction can result in collision checking errors that are very confusing and hard to explain for someone who doesn't know that this conversion is going on under the hood.

I can add this attribute to the rest of the URDFs in `tesseract_support` to make sure the downstream demos and tests still run the same way. I'm not sure this is really necessary though because I think all of the geometries are already convex hulls

Levi-Armstrong:
My concern is that tesseract has had this default behavior since the beginning and this change is likely to break a lot existing implementation. If we go with false then I would recommend that it is required that the urdf contain the global attribute and have no default.

marip8:
Okay, let's make `tesseract:make_convex` required then, and in the case that it doesn't exist, I'll add an error message saying that it is missing and that the previous default value was true.

Levi-Armstrong:
@marip8 Do think that we should remove convex_mesh?

marip8:
> @marip8 Do think that we should remove convex_mesh support in the urdf and just use mesh with the attribute?

Yes, I completely removed the `convex_mesh` element and added the `tesseract:make_convex` attribute both globally and in the `mesh` element

marip8:
After adding b681a5c to fix the Windows build, I see the following error in clang-tidy and MacOS:

```
error: lambda capture 'global_make_convex' is not required to be captured for this use [clang-diagnostic-unused-lambda-capture]
```

@Levi-Armstrong any idea how to proceed?

Levi-Armstrong:
> After adding [b681a5c](https://github.com/tesseract-robotics/tesseract/commit/b681a5ca3d72562dce2da7b6141cdf7ca69ee263) to fix the Windows build, I see the following error in clang-tidy and MacOS:
> 
> ```
> error: lambda capture 'global_make_convex' is not required to be captured for this use [clang-diagnostic-unused-lambda-capture]
> ```
> 
> @Levi-Armstrong any idea how to proceed?

Does the following work?

``` cpp
const auto parse_collision_fn = [global_make_convex = false](const tinyxml2::XMLElement* xml_element,
                                                        const tesseract_common::ResourceLocator& locator) {
    return tesseract_urdf::parseCollision(xml_element, locator, global_make_convex);
  };
```

marip8:
It appears that I fixed the issue by only providing `&` in the capture instead of `&global_make_convex`. Not sure why including everything by reference is okay but capturing a single variable by reference is not...

tesseract_urdf/include/tesseract_urdf/octree.h
```diff
@@ -42,6 +43,8 @@ class XMLDocument;
 
 namespace tesseract_urdf
 {
+static constexpr std::string_view OCTREE_ELEMENT_NAME = "tesseract:octree";
```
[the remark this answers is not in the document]
marip8: No; everything prefixed with `tesseract:` are elements that are not officially supported by the URDF specification. The only geometries in the URDF specification are `box`, `cylinder`, `sphere`, and `mesh` per [here](https://wiki.ros.org/urdf/XML/link#Elements)

Levi-Armstrong:
I think the issue is that it is const.

Levi-Armstrong:
Lets hold off merging this until I finish fixing 0.28 release.

tesseract_urdf/src/mesh.cpp
```diff
@@ -121,3 +157,5 @@ tinyxml2::XMLElement* tesseract_urdf::writeMesh(const std::shared_ptr<const tess
 
   return xml_element;
```
marip8, line 158: If we save a `ConvexMesh` into the URDF as a `<mesh>` element and then load that mesh from the generated URDF, it will become a `Mesh` even though it is technically convex. Will that cause problems downstream in Tesseract/Trajopt? Does Tesseract/Trajopt only think meshes are convex if they are explicitly `ConvexMesh` types?

If so, I think we will need to add the `tesseract:make_convex` attribute whenever a `ConvexMesh` is encountered, such that it gets converted back into a `ConvexMesh` during the parse of the URDF.

An alternative would be to create a `isConvex` utility that is used in the parse to determine if a mesh is convex. Then we could set its type appropriately. Or better yet, we could do away with the `ConvexMesh` type altogether and add the `isConvex` utility directly to the `Mesh` base class

```suggestion
  // If the mesh is a convex mesh, set the `tesseract:make_convex` attribute true
  if (std::dynamic_pointer_cast<const tesseract_geometry::ConvexMesh>(mesh))
  {
    xml_element->SetAttribute("tesseract:make_convex", true);
  }

  return xml_element;
```

marip8: Per our discussion today and on a little further investigation, it appears that this problem currently exists in `tesseract` now. For `tesseract` URDF version < 2, all collision meshes automatically get converted to convex meshes. So, if you were to write a URDF from a `tesseract` environment, you would be saving new collision meshes that are convex hulls of the shapes that were used to generate the environment (although they would be saved with the notation that they are convex meshes).

https://github.com/tesseract-robotics/tesseract/blob/9658a37c3231bafc845c1cd280174128e88b0992/tesseract_urdf/src/geometry.cpp#L171-L185

> If so, I think we will need to add the tesseract:make_convex attribute whenever a ConvexMesh is encountered, such that it gets converted back into a ConvexMesh during the parse of the URDF.

Adding the block of code in the previous comment should produce the same behavior that `tesseract` has now

tesseract_urdf/test/tesseract_urdf_convex_mesh_unit.cpp
```diff
@@ -25,10 +25,14 @@ TEST(TesseractURDFUnit, parse_convex_mesh)  // NOLINT
   tesseract_common::GeneralResourceLocator resource_locator;
   {
     std::string str =
-        R"(<convex_mesh filename="package://tesseract_support/meshes/box_2m.ply" scale="1 2 1" extra="0 0 0"/>)";
+        R"(<tesseract:convex_mesh filename="package://tesseract_support/meshes/box_2m.ply" scale="1 2 1" extra="0 0 0"/>)";
```
Levi-Armstrong, line 28: I thought convex_mesh was removed as part of this PR or is that not the case?

marip8: Looks like I had removed this unit test from the CMakeLists but forgotten to delete the file. It's deleted now, and there should not be any more references to `<tesseract:convex_mesh>` elements

tesseract_urdf/test/tesseract_urdf_mesh_unit.cpp
Levi-Armstrong: @marip8 Is this still relevant?

marip8: The previous unit tests check the parsing of meshes both with and without the conversion to convex hulls. This unit test just checks the workflow for parsing a mesh without converting to convex hull.

I think I could re-introduce the tests from the convex mesh unit test (that I deleted) and turn those into the tests that check that `global_make_convex = true` case

Levi-Armstrong: Yea, it is probably good idea to keep and update the tests you deleted.

rjoomen:
Good initiative. Will the new style Tesseract URDF satisfy the official urdfdom XSD? See https://github.com/ros/urdfdom/pull/200 for the latest version (WIP).

marip8:
> Good initiative. Will the new style Tesseract URDF satisfy the official urdfdom XSD? See [ros/urdfdom#200](https://github.com/ros/urdfdom/pull/200) for the latest version (WIP).

Yes, I believe so. These changes do not overwrite any elements or attributes that are already specified in the urdfdom XSD. It only adds a few new namespaced geometry elements (e.g., `tesseract:capsule`, `tesseract:octree`, etc.) and a new namespaced attributed (`tesseract:make_convex`) for the existing `robot` and `mesh` elements

Levi-Armstrong:
@marip8 You good if I squash merge this?

rjoomen:
> > Good initiative. Will the new style Tesseract URDF satisfy the official urdfdom XSD? See [ros/urdfdom#200](https://github.com/ros/urdfdom/pull/200) for the latest version (WIP).
> 
> Yes, I believe so. These changes do not overwrite any elements or attributes that are already specified in the urdfdom XSD. It only adds a few new namespaced geometry elements (e.g., `tesseract:capsule`, `tesseract:octree`, etc.) and a new namespaced attributed (`tesseract:make_convex`) for the existing `robot` and `mesh` elements

Great. Maybe add this to the

marip8:
> @marip8 You good if I squash merge this?

Yes it should be ready for merge now and squashing the history is fine with me

rjoomen:
> > Would it be possible to add this to the unit tests?
> 
> What do you mean exactly? There are unit tests that parse all of these elements with their new names and test the `tesseract:make_convex` attribute. Do you mean testing compatibility with urdfdom XSD?

Yes

PR #1083, opened 2024-12-15
Fix use of compound mesh in FCL

### LINKED ISSUE #1053: CompoundMesh does not work with FCL

Levi-Armstrong:
Fix #1053 

@rjoomen After looking through the code I went a slight different route to solve another issue with get the correct shape index. Do you mind testing this?

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -35,7 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - runner: macos-12
+          - runner: macos-13
             vcpkg_triplet: x64-osx-dynamic-release
             arch: x64
             homebrew_root: /usr/local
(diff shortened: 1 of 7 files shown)


Levi-Armstrong:
@rjoomen Hey, I did not realize you had pushed changes and I force pushed after I noticed my local clang tidy build failed. What changes did you make?

rjoomen:
No changes, I just rebased on master using Github.

rjoomen:
I'll look into this tomorrow.

rjoomen:
Seems like a better solution, and it works, thanks. I added a small simplification.

PR #1099, opened 2025-01-07
Add path API functions to GeneralResourceLocator

johnwason:
This PR adds API functions to GeneralResourceLocator to add paths to the locator using a constructor overload or the new `addPath()` function.

### DIFF
--- a/tesseract_common/include/tesseract_common/resource_locator.h
+++ b/tesseract_common/include/tesseract_common/resource_locator.h
@@ -36,6 +36,8 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <boost/serialization/export.hpp>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
+#include <tesseract_common/filesystem.h>
+
 namespace boost::serialization
 {
 class access;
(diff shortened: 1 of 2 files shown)


tesseract_common/include/tesseract_common/resource_locator.h
```diff
@@ -82,7 +84,18 @@ class GeneralResourceLocator : public ResourceLocator
...
+   */
   GeneralResourceLocator();
+  /**
+   * @brief Construct a new General Resource Locator object using the provided paths and/or the TESSERACT_RESOURCE_PATH
+   * environment variable
+   *
+   * @param paths A vector of paths to search for resources
+   * @param use_environment If true, the TESSERACT_RESOURCE_PATH environment variable will also be used
+   */
+  GeneralResourceLocator(const std::vector<tesseract_common::fs::path>& paths, bool use_environment = true);
```
Levi-Armstrong, line 98: Is the use_environment needed? If they do not what to use the environment variable shouldn't they not defined the environment variable?

johnwason: The system or ROS may set these variables automatically. It is better to have the option to opt in or out of the env variable.

johnwason: I have made the suggested changes

Levi-Armstrong: I guess, why we are at should we just expose a loadEnvironmentVariable function?

johnwason: I renamed and made the function public

Levi-Armstrong (approved):

Levi-Armstrong:
It looks like you might need to squash everything and rebase before I can merge.

Levi-Armstrong:
@johnwason It looks like the mac CI is failing.

PR #1120, opened 2025-02-01
Fix collision shape caching

Levi-Armstrong:


rjoomen:
This is not working yet, right? See [here](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/bullet/src/bullet_utils.cpp#L438) for Bullet, and [here](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/fcl/src/fcl_utils.cpp#L215) for FCL.

Shouldn't [this](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/bullet/src/bullet_utils.cpp#L368-L371) then also be commented out?

rjoomen:
> This is not working yet, right? See [here](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/bullet/src/bullet_utils.cpp#L438) for Bullet, and [here](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/fcl/src/fcl_utils.cpp#L215) for FCL.
> 
> Shouldn't [this](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/bullet/src/bullet_utils.cpp#L368-L371) then also be commented out?

@Levi-Armstrong

Levi-Armstrong:
@rjoomen You are correct those shouldn't be commented out.

PR #1130, opened 2025-03-14
Documentation Update

marip8:
This PR starts the process of porting the documentation for this repository from the contents of the `tesseract_docs` repository. First it adds a Github action that runs Doxygen and uploads the resulting artifacts to Github Pages where the documentation website can be hosted.

It also ports the documentation pages from Sphinx (ReST files) to Doxygen (`.dox` text files). The integration of Sphinx with Doxygen seems a bit complicated for what its worth. By using only Doxygen directly, I was able to:
- Generate full API documentation, with included API search engine (not currently included in existing Sphinx documentation AFAIK)
- Easily port several description pages (getting started, SRDF, URDF) while retaining the same features as the Sphinx documentation (e.g., linking, code blocks, etc.)
- Introduce new features (supported natively by Doxygen) like integrating code snippets directly from files in this repository into documentation pages
- Generate a reasonable-looking website with minor customizations to a CSS theme

There is a lot more work to be done, both on porting the rest of the existing documentation and fine-tuning the look of the website, but here are a few images of the generated pages:

image

image


Levi-Armstrong:
@marip8 You okay with this being merged?

marip8:
> @marip8 You okay with this being merged?

Yes, this should be good to go now, and we can add to it going forward

Levi-Armstrong:
@marip8 The only thing holding this up is the CI is failing.

Levi-Armstrong:
I don't think it likes the changes to the yaml files.

marip8:
Right, I forgot about that. It looks like doxygen will still recognize the snippet behind a YAML comment symbol (the documentation still builds correctly on my machine), so it should be good to go now

Issue #1057, opened 2024-11-04
Incomplete tip link list from `KinematicGroup::getAllPossibleTipLinkFrames`


marip8:
The available tip links for the `KinematicGroup` currently only include the tip links of the IK solver and direct children of those tip links:

https://github.com/tesseract-robotics/tesseract/blob/67a83c76d12a9bde43c9d96e4e747b558155bbc5/tesseract_kinematics/core/src/kinematic_group.cpp#L90-L95

It seems like this list should also include any link that is attached via a fixed joint **tree/chain** to the tip link(s) defined by the IK solver. It seems like this ought to be relatively easy to implement using the graph search capability of the scene graph.

I've run into this issue several times before where the TCP frame you want to use for planning was added several fixed joints beyond the IK solver tip frame to correspond more directly to geometry in the robot. When a frame like this is used for planning, the error is a very unhelpful generic `map::at` exception. It would also be nice to maybe wrap and rethrow that exception with more detailed information

Levi-Armstrong:
> It seems like this list should also include any link that is attached via a fixed joint tree/chain to the tip link(s) defined by the IK solver. It seems like this ought to be relatively easy to implement using the graph search capability of the scene graph.

The call to scene_graph.getLinkChildrenNames should get all children not just the immediate child. Do you have an example where this does not work as expect?

marip8:
I think this does in fact function as expected. In the use-case where I thought I encountered this issue, I wasn't overly familar with the URDF; it turns out we had specified a TCP link that was a "cousin" of the tip link (it could be reached -- through fixed joints --  by going up one link in the hierarchy and then down to the specified TCP link). Is it worth trying to support a use-case like this, maybe in the case that the tip link can't be found in the existing map? It seems like it could be done using `SceneGraph::getShortestPath` and checking that all joints are fixed

marip8:
Per our discussion yesterday, maybe a better approach would be to walk backwards from the specified tip link until we hit an actuated joint, and run `getLinkChildrenNames` from the child link of that actuated joint instead. This would capture all "cousin" frames that could potentially be specified as TCP frames

Levi-Armstrong:
@marip8 Can this be closed?

PR #1136, opened 2025-04-01
`KinematicsGroup`: return complete list of possible tip links and working frames

marip8:
Addresses #1057


tesseract_kinematics/core/src/kinematic_group.cpp
```diff
@@ -34,8 +34,67 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
...
+#include <tesseract_scene_graph/joint.h>
 #include <tesseract_state_solver/state_solver.h>
 
+/**
+ * @brief Returns the names of the links that comprise a fixed-joint kinematic tree with the input link
+ * @details This method first traverses from the input link "up" the scene graph hierarchy, following the link's
+ * in-bound joints, to find the input link's highest level parents that are connected to it by fixed joints. Next, it
+ * identifies all the fixed-joint children of these highest level parent links. Together, the highest level parent links
+ * and their fixed-joint children comprise a set of links (including the input link), between all of which exists a tree
+ * of fixed joints. If the link does not have any fixed-joint parents, the name of the link itself will be returned.
```
Levi-Armstrong, line 46: Why not just return an empty vector?

Levi-Armstrong (commented):
I just had one question. This is a great improvement!

tesseract_kinematics/core/src/kinematic_group.cpp
marip8: Mostly for convenience. In the class constructor I push the contents of the returned vector into `working_frames_` and `inv_kin_tip_links_map_`, so it's more convenient to also have the nominal working frame/tip link be a part of that vector than having to add it to those class members separately. I tried to rename this method and update its documentation to indicate that the return value is the set of links involved in a fixed joint kinematic tree that includes the nominal link itself; I can add more detail there if you want

PR #1137, opened 2025-04-02
Add validate method to ContactManagerConfig

Levi-Armstrong:
This adds a validate method to the ContactManagerConfig which is used in the contact manager's apply config method. This check for common issues where a users sets data but does not set the override type for either margin data or the allowed collision matrix. If a discrepancy is found it throws an exception.

Note: The default acm_override_type has been changed to NONE to allow validations requiring the user to set the override type when providing acm data.

### DIFF
--- a/tesseract_collision/core/include/tesseract_collision/core/types.h
+++ b/tesseract_collision/core/include/tesseract_collision/core/types.h
@@ -428,11 +428,18 @@ struct ContactManagerConfig
   /** @brief Additional AllowedCollisionMatrix to consider for this collision check.  */
   tesseract_common::AllowedCollisionMatrix acm;
   /** @brief Specifies how to combine the ContactAllowedValidator from acm with the one preset in the contact manager */
-  ACMOverrideType acm_override_type{ ACMOverrideType::OR };
+  ACMOverrideType acm_override_type{ ACMOverrideType::NONE };
 
   /** @brief Each key is an object name. Objects will be enabled/disabled based on the value. Objects that aren't in the
    * map are unmodified from the defaults*/
   std::unordered_map<std::string, bool> modify_object_enabled;
+
+  /**
+   * @brief Check for errors and throw exception if they exist
+   * If margin_data_override_type is set to NONE but margin pair data exist or non zero default margin
+   * If acm_override_type is set to NONE but allowed collision entries exist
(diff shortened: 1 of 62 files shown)


tesseract_collision/core/include/tesseract_collision/core/types.h
```diff
@@ -428,11 +428,18 @@ struct ContactManagerConfig
...
+  ACMOverrideType acm_override_type{ ACMOverrideType::NONE };
 
   /** @brief Each key is an object name. Objects will be enabled/disabled based on the value. Objects that aren't in the
    * map are unmodified from the defaults*/
   std::unordered_map<std::string, bool> modify_object_enabled;
+
+  /**
+   * @brief Check for errors and throw exception if they exist
+   * If margin_data_override_type is set to NONE but margin pair data exist or non zero default margin
+   * If acm_override_type is set to NONE but allowed collision entries exist
```
marip8, line 440:
```suggestion
   * @brief Check for errors and throw exception if they exist
   * @throws an exception if `margin_data_override_type` is set to `NONE`, but margin pair data exist or a non-zero default margin is specified; or if `acm_override_type` is set to `NONE` but allowed collision entries exist.
```

Fix for better rendering with Doxygen

tesseract_collision/core/src/types.cpp
```diff
@@ -365,6 +365,23 @@ ContactManagerConfig::ContactManagerConfig(double default_margin)
...
+{
+  if (margin_data_override_type == CollisionMarginOverrideType::NONE)
+  {
+    if (!margin_data.getPairCollisionMargins().empty())
+      throw std::runtime_error("ContactManagerConfig, margin data overide type is NONE but pair collision margins "
+                               "exist!");
+
+    if (!tesseract_common::almostEqualRelativeAndAbs(margin_data.getDefaultCollisionMargin(), 0.0))
+      throw std::runtime_error("ContactManagerConfig, margin data overide type is NONE but has non zero default "
+                               "margin!");
```
marip8, line 378: Why is this a problem? The default collision margin should be used regardless of whether you specify any types of overrides, right?

Levi-Armstrong: No longer valid

PR #1166, opened 2025-06-25
Update yaml_utils.cpp

Levi-Armstrong:
Fix processYamlIncludeDirective to make a deep copy of nodes


Levi-Armstrong:
replaced by PR #1172

Issue #1189, opened 2025-08-04
Failure to build tesseract_collision


wglockner:
### Version

`0.31.0`

### OS Version

Ubuntu 22.04

### Describe the bug

Tesseract fails to build at the tesseract_collision package during a clean build in new workspace.

### To Reproduce

```bash
# Build the extended workspace as per the instructions to speed up a clean build.
mkdir ~/tesseract_ext_ws/src -p && cd ~/tesseract_ext_ws/src
git clone git@github.com:tesseract-robotics/tesseract.git
cd ~/tesseract_ext_ws/src/tesseract/
git checkout 0.31.0
cd ~/tesseract_ext_ws/
vcs import --shallow --debug src < src/tesseract/dependencies_with_ext.repos
source /opt/ros/humble/setup.bash
rosdep install --from-paths src --ignore-src -r -y
# Remove tesseract from the extended workspace.
rm -rf ~/tesseract_ext_ws/src/tesseract
colcon build --symlink-install --cmake-args -DTESSERACT_BUILD_FCL=OFF -DBUILD_RENDERING=OFF -DBUILD_STUDIO=OFF
source install/setup.bash
# Build the tesseract workspace.
mkdir ~/tesseract_ws/src -p && cd ~/tesseract_ws/src
git clone git@github.com:tesseract-robotics/tesseract.git
cd ~/tesseract_ws/src/tesseract/
git checkout 0.31.0
cd ~/tesseract_ws/
vcs import --shallow --debug src < src/tesseract/dependencies.repos
source /opt/ros/humble/setup.bash
rosdep install --from-paths src --ignore-src -r -y
# Below is the failing step.
colcon build --symlink-install --cmake-args -DTESSERACT_BUILD_FCL=OFF -DBUILD_RENDERING=OFF -DBUILD_STUDIO=OFF
```

### Expected behavior

Summary: [number of packages] packages finished [build time]

### Relevant log output

```shell
--- stderr: tesseract_collision                                 
Building Test Suite components
Building VHACD components
CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_RENDERING
    BUILD_STUDIO


In file included from /home/[user]/tesseract_ext_ws/src/tesseract/tesseract_collision/bullet/src/bullet_factories.cpp:32:
[...]

PR #1190, opened 2025-08-04
Update boost_plugin_loader dependency to main to fix build issue.

wglockner:
Fix to resolve build failure referenced in issue #1189.

### DIFF
--- a/dependencies.repos
+++ b/dependencies.repos
@@ -5,7 +5,7 @@
 - git:
     local-name: boost_plugin_loader
     uri: https://github.com/tesseract-robotics/boost_plugin_loader.git
-    version: 0.3.0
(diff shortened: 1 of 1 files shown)


dependencies.repos
```diff
@@ -5,7 +5,7 @@
 - git:
     local-name: boost_plugin_loader
     uri: https://github.com/tesseract-robotics/boost_plugin_loader.git
-    version: 0.3.0
+    version: main
```
marip8, line 8: Per request in https://github.com/tesseract-robotics/tesseract/pull/1191#issuecomment-3164548586, I will say we always prefer to track named versions (tags or commit hashes) over named branches that can become moving targets over time

PR #1191, opened 2025-08-06
Update boost_plugin_loader to version 0.3.2

Levi-Armstrong:


jdham-iastate:
Hi! Just for my own clarity so I can help my team better in the future, why did you choose the changes in this merge rather than the one in #1190 ? wglockner and I want to be more useful at contributing in the future.

marip8:
I think this PR doesn't directly address #1189, so it wasn't really intended to compete with #1190 per se. We added a bug fix in the plugin loader version 0.3.2 and wanted to include it in Tesseract, so the update was made. I'm still investigating why the issue reported in #1189 happens; all the CI checks passed for 0.30.x and 0.31.x with the older version of the plugin loader, so it's still unclear why this issue shows up.

PR #1193, opened 2025-08-12
Add yaml encode/decode for commonly used Eigen::Matrix<double, 6, 1>

marrts:
This Eigen matrix is commonly used for coefficients or tolerances in Cartesian space (x,y,z,r,p,y)


Levi-Armstrong:
> > Looks good; @marrts could add a test for this?
> 
> Done, I couldn't find any tests for any of the eigen yaml encode/decode methods, so I added tests for all of them

Awesome and thank you!

PR #1198, opened 2025-08-13
Use per object max margin for the broadphase and the actual link pair margin for the narrow phase, instead of the overall max margin everywhere

rjoomen:
I believe the margin used for AABB expansion in Bullet should be divided by 2, as otherwise the margin is effectively duplicated when two expanded AABBs are checked against each other in the broadphase. The margin is set to `getMaxCollisionMargin()` [here](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/bullet/src/bullet_cast_bvh_manager.cpp#L530).
For reference, for FCL [this is done as well](https://github.com/tesseract-robotics/tesseract/blob/master/tesseract_collision/fcl/src/fcl_discrete_managers.cpp#L398).

Edit: I have significantly expanded this PR.
I added functionality to CollisionMarginPairData and CollisionMarginData to provide a maximum margin per object, instead of only a max margin for all objects together. The collision managers are updated to use the per object max margin, or the actual pair margin. This significantly speeds up collision checking when pair margins vary widely, as previously the worst case margin was always used, leading to needless narrowphase collision checks.

### DIFF
--- a/tesseract_common/include/tesseract_common/collision_margin_data.h
+++ b/tesseract_common/include/tesseract_common/collision_margin_data.h
@@ -95,9 +95,9 @@ public:
 
   /**
    * @brief Get the largest pair collision margin
-   * @return Max pair contact distance threshold
+   * @return Max pair contact distance threshold if objects exist
    */
-  double getMaxCollisionMargin() const;
+  std::optional<double> getMaxCollisionMargin() const;
 
   /**
    * @brief Get the largest collision margin for the given object
@@ -149,7 +149,7 @@ private:
   PairsCollisionMarginData lookup_table_;
 
   /** @brief Stores the largest collision margin */
-  double max_collision_margin_{ std::numeric_limits<double>::lowest() };
+  std::optional<double> max_collision_margin_;
 
   /** @brief Stores the maximum collision margin for each object */
   std::unordered_map<std::string, double> object_max_margins_;
--- a/tesseract_common/src/collision_margin_data.cpp
+++ b/tesseract_common/src/collision_margin_data.cpp
@@ -71,7 +71,7 @@ std::optional<double> CollisionMarginPairData::getCollisionMargin(const std::str
   return {};
 }
 
-double CollisionMarginPairData::getMaxCollisionMargin() const { return max_collision_margin_; }
+std::optional<double> CollisionMarginPairData::getMaxCollisionMargin() const { return max_collision_margin_; }
 
 std::optional<double> CollisionMarginPairData::getMaxCollisionMargin(const std::string& obj) const
 {
@@ -89,7 +89,8 @@ void CollisionMarginPairData::incrementMargins(double increment)
   if (lookup_table_.empty())
     return;
 
-  max_collision_margin_ += increment;
+  assert(max_collision_margin_.has_value());
+  max_collision_margin_.value() += increment;  // NOLINT
   for (auto& pair : lookup_table_)
     pair.second += increment;
 
@@ -103,7 +104,8 @@ void CollisionMarginPairData::scaleMargins(double scale)
   if (lookup_table_.empty())
     return;
 
-  max_collision_margin_ *= scale;
+  assert(max_collision_margin_.has_value());
+  max_collision_margin_.value() *= scale;  // NOLINT
   for (auto& pair : lookup_table_)
     pair.second *= scale;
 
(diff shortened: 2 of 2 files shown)


tesseract_collision/bullet/src/bullet_utils.cpp
```diff
@@ -530,7 +530,7 @@ const tesseract_common::VectorIsometry3d& CollisionObjectWrapper::getCollisionGe
 void CollisionObjectWrapper::getAABB(btVector3& aabb_min, btVector3& aabb_max) const
 {
   getCollisionShape()->getAabb(getWorldTransform(), aabb_min, aabb_max);
-  const btScalar& d = getContactProcessingThreshold();
+  const btScalar& d = getContactProcessingThreshold() / 2.0;
```
Levi-Armstrong, line 533: I think it might be better to make this calculation here, this way we do not need to worry about if bullet uses shape processing threshold any other place.

``` c++
void BulletCastBVHManager::onCollisionMarginDataChanged()
{
  auto margin = static_cast<btScalar>(contact_test_data_.collision_margin_data.getMaxCollisionMargin() / 2.0);
```

rjoomen: Eh, I commented with some explanation, but that apparently got lost. Will retry tomorrow.

Levi-Armstrong: @rjoomen Friendly ping

rjoomen: The reason to not do it in onCollisionMarginDataChanged() is that the ContactProcessingThreshold of the object is then set to half the margin. This threshold is also used [here](https://github.com/bulletphysics/bullet3/blob/d1a4256b3a019117f2bb6cb8c63d6367aaf512e2/src/BulletCollision/CollisionDispatch/btCollisionDispatcher.cpp#L77) or [here](https://github.com/tesseract-robotics/tesseract/blob/de5f4d8e4cd3a5fa9737fc5571727f02b8b69016/tesseract_collision/bullet/src/bullet_cast_simple_manager.cpp#L410), where I do not believe a halved margin would be expected. By only halving the margin in getAABB() we make sure it's only used in the broadphase.

rjoomen: @Levi-Armstrong I've expanded this PR, see the modified description.

Levi-Armstrong: > The reason to not do it in onCollisionMarginDataChanged() is that the ContactProcessingThreshold of the object is then set to half the margin. This threshold is also used [here](https://github.com/bulletphysics/bullet3/blob/d1a4256b3a019117f2bb6cb8c63d6367aaf512e2/src/BulletCollision/CollisionDispatch/btCollisionDispatcher.cpp#L77) or [here](https://github.com/tesseract-robotics/tesseract/blob/de5f4d8e4cd3a5fa9737fc5571727f02b8b69016/tesseract_collision/bullet/src/bullet_cast_simple_manager.cpp#L410), where I do not believe a halved margin would be expected. By only halving the margin in getAABB() we make sure it's only used in the broadphase.

This makes sense, it would be good to document what you just explained where you make the change.

Levi-Armstrong:
@rjoomen This on my todo list for this week.

tesseract_collision/fcl/src/fcl_discrete_managers.cpp
```diff
@@ -395,7 +395,7 @@ void FCLDiscreteBVHManager::onCollisionMarginDataChanged()
 
   for (auto& cow : link2cow_)
   {
-    cow.second->setContactDistanceThreshold(collision_margin_data_.getMaxCollisionMargin() / 2.0);
+    cow.second->setContactDistanceThreshold(collision_margin_data_.getMaxCollisionMargin(cow.second->getName()) / 2.0);
```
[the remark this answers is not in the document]
rjoomen: For consistency we could of course modify this to only divide the `contact_distance_` by 2.0 when returning the AABB in `FCLCollisionObjectWrapper::updateAABB()`. Then the contact distance property of the wrappers would be more meaningful, and consistent with Bullet.

rjoomen:
I'll take a good look at this on Monday.

PR #1199, opened 2025-08-14
Improved logging

johnwason:
This PR contains some preliminary work on an improved logging system inspired by Robot Raconteur. It is designed to carry more information about the log event for displaying to the user in an "event viewer" format, or for clients to automatically interpret the log information to make decisions.

The `TesseractLogRecord` has this initial format:

```cpp
struct TesseractLogRecord
    {
        /** The log record error level */
        LogLevel level{LogLevel::Warning};
        /** An identifier for an error. Either a gcc style "diagnostic ID" or a UUID in hex string  */
        std::string error_ident;
        /** The component that generated the message (typically package name or namespace) */
        std::string component;
        /** The subcomponent that generate the message (typically class name) */
        std::string subcomponent;
        /** The instance ID of the object that generated the message (pointer or UUID)*/
        std::string object_instance;
        /** The path of the object. Use string UUIDs of taskflow nodes */
        std::vector<std::string> object_path;
        /** Human readable log message */
        std::string message;
        /** Time of logging event */
        std::chrono::system_clock::time_point time;
        /** The sourcecode filename */
        std::string source_file;
        /** The line within the sourcecode file */
        uint32_t source_line{0};
        /** The source thread */
        std::string thread_id;
        /** Machine readable parameters attached to event */
        std::unordered_map<std::string, std::any> parameters;
    };
```
This improvement also introduces the concept of `LogManager` and `LogRecordHandler`. Ideally all classes in tesseract will carry an instance variable to `std::weak_ptr<LogManager>`, or possibly a uint32 ID to the log manager currently being used. For the Robot Raconteur service there may be more than one active client so it is necessary to send the log to the correct manager so it is received by the relevant client.

See also https://github.com/robotraconteur/robotraconteur/blob/master/RobotRaconteurCore/include/RobotRaconteur/Logging.h , https://github.com/robotraconteur/robotraconteur/blob/master/RobotRaconteurCore/src/Logging.cpp , and https://github.com/robotraconteur/robotraconteur_standard_robdef/blob/master/group1/com.robotraconteur.eventlog.robdef

This work is still incomplete and is being submitted for discussion.

### DIFF
--- a/tesseract_common/CMakeLists.txt
+++ b/tesseract_common/CMakeLists.txt
@@ -101,7 +101,8 @@ add_library(
   src/types.cpp
   src/stopwatch.cpp
   src/timer.cpp
-  src/yaml_utils.cpp)
+  src/yaml_utils.cpp
+  src/logging.cpp)
 target_link_libraries(
   ${PROJECT_NAME}
   PUBLIC Eigen3::Eigen
--- a/tesseract_common/include/tesseract_common/logging.h
+++ b/tesseract_common/include/tesseract_common/logging.h
@@ -1,0 +1,207 @@
+/**
+ * @file logging.h
+ * @brief Tesseract Logging
+ *
+ * Improved Tesseract logging inspired by Robot Raconteur log system
+ * 
+ * @author John Wason
+ * @date August 14, 2025
+ * @version TODO
+ * @bug No known bugs
+ *
+ * @copyright Copyright (c) 2025, Wason Technology, LLC
+ *
+ * @par License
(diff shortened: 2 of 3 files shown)


johnwason:
It may also make sense to make this a separate library that can be used by other related packages.

johnwason:
@Levi-Armstrong do you want to pursue the improved logging?

Levi-Armstrong:
Improving logging is still on the 1.0 release list but I have not had time to look at this yet. How would something like this integrate with spdlog if we wanted to use it?

PR #1200, opened 2025-08-14
Check trajectory return contact location

marrts:
The goal here is to return where in the trajectory a contact occurred. Now you can get detailed information about between which steps and at which substep a contact occurred, including the actual joint states at those values.

In creating this I also discovered that the `ContactTestType type` in `ContactRequest` was being used both for the individual contact checks AND dictated the trajectory collision checking behavior. This meant that if you used `ContactTestType::ALL` it would always check every state with the `ALL` setting, but it also would continue to contact check other states in the trajectory after a contact had been found. On the other hand, if you used `ContactTestType::FIRST` it would stop checking states once a contact was found. 

With this new change I'm introducing it may be useful to find all states that are in collision, but you may want to do that quickly using the `FIRST` methodology. There was no way to do that, so I added a field to `CollisionCheckConfig` called `exit_on_first_contact` which defaults to `true`.

I benchmarked this new one against the previous version of the code and go the following graph, virtually no chnage in performance (it appears to actually be slightly more efficient in DEBUG mode than it was before because of how it creates objects now):
CheckTrajBenchmark

### DIFF
--- a/tesseract_collision/core/include/tesseract_collision/core/types.h
+++ b/tesseract_collision/core/include/tesseract_collision/core/types.h
@@ -493,166 +493,6 @@ struct CollisionCheckConfig
 };
 
 /**
- * @brief Represents a contact occurrence inside a trajectory step.
- *
- * A substep value of -1 indicates a contact at the discrete step (no substep).
- */
-struct ContactLocation
-{
-  int step{ -1 };    /**< The trajectory step index where contact occurred */
-  int substep{ -1 }; /**< The substep index (-1 if not applicable) */
-
-  ContactLocation() = default;
-  ContactLocation(int s) : step(s) {}
-  ContactLocation(int s, int sub_s) : step(s), substep(sub_s) {}
-
-  bool operator==(const ContactLocation& rhs) const { return step == rhs.step && substep == rhs.substep; }
-  bool operator!=(const ContactLocation& rhs) const { return !operator==(rhs); }
-};
-
-/**
- * @brief Per-step contact information.
- *
- * Stores a compact list of substep indices where contacts were detected for a given step.
- * A stored substep value of -1 represents a contact at the step itself (no substep).
- */
-struct StepContact
-{
-  int step = -1;
-  std::vector<int> substeps;  // may contain -1 to indicate a step-only contact
-
-  StepContact() = default;
-  explicit StepContact(int s) : step(s) {}
-};
-
-/**
- * @brief Efficient return type for checkTrajectory functions that replaces boolean return
- * while tracking contact locations with minimal overhead.
- *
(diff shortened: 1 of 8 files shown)


tesseract_environment/include/tesseract_environment/utils.h
```diff
@@ -156,13 +159,14 @@ bool checkTrajectory(std::vector<tesseract_collision::ContactResultMap>& contact
  * @param manip The kinematic joint group
  * @param traj The joint values at each time step
  * @param config CollisionCheckConfig used to specify collision check settings
- * @return True if collision was found, otherwise false.
+ * @return TrajectoryContactResult containing contact locations found.
```
marrts, line 162: This is incorrect, got updated

marrts:
It's also worth noting I made some changes to how things like end states are treating during trajectory contact checking.

Previously, for LVS checks, the last state would get coupled with the previous state, but at time t=1, indicating it was actually at the next state. In trying to add tests for all my new code, I found this confusing and hard to work with. I restructured things so that end states would actually be read as a new separate state at time t=0. I feel this is more intuitive when looking at the results data.

tesseract_collision/core/include/tesseract_collision/core/types.h
```diff
@@ -485,6 +497,9 @@ struct CollisionCheckConfig
   /** @brief Secifies the mode used when collision checking program/trajectory. Default: ALL */
   CollisionCheckProgramType check_program_mode{ CollisionCheckProgramType::ALL };
 
+  /** @brief If true, the collision check will exit on first state contact. Default: true */
```
Levi-Armstrong, line 500: It looks like the description needs to be updated

Levi-Armstrong (changes requested):

marrts:
> @marrts, Thank you for running the benchmarks. It might be good to add to the repo to aid when making changes in the future; what do you think?

Yeah, I can do that. I have some local changes that have the version of checkTrajectory that I am overwriting to generate this head to head comparison, but I feel that would be silly to keep around. I can just add the new benchmarks if that makes sense to you?

tesseract_collision/core/include/tesseract_collision/core/types.h
marrts: Good catch, reworded to:
```C++
/** @brief The condition in which the trajectory collision check will exit. Default: FIRST */
```

Levi-Armstrong (approved):

Levi-Armstrong:
@marrts It looks like there is a few clang-tidy errors. Once these are fixed I can get this merged.

Levi-Armstrong:
I fixed the clang-tidy errors and once this passes I will merge.

PR #1204, opened 2025-09-01
Raise TESSERACT_CXX_VERSION from 17 to 20

rjoomen:
This allows us to make use of improved C++ language features.

### DIFF
--- a/tesseract_collision/core/include/tesseract_collision/core/types.h
+++ b/tesseract_collision/core/include/tesseract_collision/core/types.h
@@ -394,6 +394,17 @@ enum class CollisionCheckProgramType : std::uint8_t
   INTERMEDIATE_ONLY
 };
 
+/** @brief Controls when collision checking exits or continues collecting results */
+enum class CollisionCheckExitType : std::uint8_t
+{
+  /** @brief Exit as soon as any collision is detected */
+  FIRST,
+  /** @brief For LVS, stop at first collision per step; continue to next step */
+  ONE_PER_STEP,
+  /** @brief Collect all collisions */
+  ALL
+};
+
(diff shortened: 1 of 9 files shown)


rjoomen:
This works for all Tesseract projects except tesseract_qt (and perhaps tesseract_ros2), because of [an issue](https://bugzilla.redhat.com/show_bug.cgi?id=2280366) [(more info](https://stackoverflow.com/questions/63513984/can-class-template-constructors-have-a-redundant-template-parameter-list-in-c2) in (EOL) Qt5. It has been fixed apparently, but I believe only for commercial releases in extended support.

BTW: [This](https://github.com/ros2/rviz/pull/1187) seems like an interesting way to implement combined Qt5/Qt6 support.

Levi-Armstrong:
I have looked into this and my only concern with this is it could lead to incompatibility when trying to release tesseract_ros2 on jammy. @rjoomen, @marip8 and @marrts What are your thoughts on this?

rjoomen:
> I have looked into this and my only concern with this is it could lead to incompatibility issues when trying to release tesseract_ros2 on jammy. @rjoomen, @marip8 and @marrts What are your thoughts on this?

The ros2/rviz issue I linked to seems to indicate that backporting the Qt6 support to Humble (on Jammy) and maybe even to Jazzy (on Noble) might be difficult. So that would mean plugins, ie. the Tesseract Workbench, should stay on Qt5 as well. And that would in turn force both tesseract_ros2 and tesseract_qt to stay on C++17, right?

Levi-Armstrong:
The concern I had was more related to supported c++ 20 features between 22.04 and 24.04. For example 22.04 standard gcc version does not support `#include <format>` where 24.04 would so we would need to be careful to avoid unsupported features for 22.04. Locally we can upgrade the gcc but I do not believe we can do that when performing a ros release on 22.04.

PR #1205, opened 2025-09-05
Implement a condensed summary for ContactTrajectoryResults

marrts:
Print a condensed summary of contacts over a trajectory. Report the first contact observed at each step, regardless of if multiple contacts and substeps with contacts are present. Useful for quick, but informative, debugging.

Of the form

```
Step 14.4: [link_a, link_b]->0.001
```

This means link_a and link_b colliding 40% of the way from step 14 to step 15 and the contact distance observed was 0.001


marrts:
> @marrts It looks like there are clang-tidy errors.

I really should be better about checking that first, working on it now

Levi-Armstrong (approved):

PR #1208, opened 2025-09-11
Check pair margin AABB before narrow phase check

Levi-Armstrong:
@marrts


marrts:
I know this isn't ready to merge yet, but I ran some tests on my application that go through a bunch of various plans with several unique collision pairs, although none exceed the default collision margin, just reduce it.

On commit 53ad1b4c345fcb7ca3ad30359393825673e5fb0b (before @rjoomen's recent changes): averaged 6 min 43 sec
On current master: averaged 6 min 15 sec
This PR rebased on master: averaged 6 min 23 sec

This was only 2 trials each, but this PR didn't show improvement for me, basically the same as the current master, which did for sure show improvement from the previous version

PR #1215, opened 2025-11-18
Update documenation to include Why Tesseract and Tesseract vs MoveIt

Levi-Armstrong:


robinsonmm:
Ok. I think there is a lot of good stuff here... 

I think we would just start with Tesseract: Next generation motion planning framework for industrial applications...

Then you can offer up the, or get into... "already using MovIt?" 
Tesseract is designed as your next-generation motion planning framework — reusing your ROS
> infrastructure while upgrading your planning, optimization, and environment model.

Line 11 - again maybe instead of "alternative" go with next generation industrial planning framework...

If you are starting a new project today and care about process quality, cycle time,
and long-term maintainability, Tesseract is ready to step in and delivery on those requirements today.

Lines 20-24 - I think we need to be more explicit... MoveIt2 is the general purpose - research community supported, Tesseract is your fully open source industrial ready soltuion... I think we can stay silent on MoveIt Pro as it is no longer open source....

Maybe we can keep the table at line 33... but either make MoveIt, MoveIt2 because i do not think we can for sure say these things about MoveItPro... you can just DROP the MoveIt side of the table as you make the case with Tesseract 

I think we just want to be careful making broad claims about MoveIt not doin certain things, because one there is a difference between the flavors of MoveIt (2 and Pro) and ppl can do the work once they fork it to make it how they like, much like we see folks do with Tesseract... all your other points about Tesseract are definitely plusses and just make Tesseract awesome w/out needing to say anything negative about MoveIt... 

Maybe just migrate to "Tesseract may be used alongside or as an alternative to MoveIt" and leave it at that... 

Just my two cents as we try to be diplomatic... thanks! Overall, love it, really helps layout the "why and what is Tesseract"

jf---:
OMPL is called in tesseract; what is a little hard to grasp is how tesseract & OMPL (i guess i think of OMPL as nearly synonymous with `moveit`) interact.
I suppose first a plausible path is searched for (ompl) and optimised to yield quality of motion / comply with set constraints (trajopt).

Levi-Armstrong:
> OMPL is called in tesseract; what is a little hard to grasp is how tesseract & OMPL (i guess i think of OMPL as nearly synonymous with `moveit`) interact. I suppose first a plausible path is searched for (ompl) and optimised to yield quality of motion / comply with set constraints (trajopt).

Tesseract has moved everything over to leverage tesseract_task_composer which is a computation graph framework with the default implementation using taskflow but one could easily write a tbb plugin. This frameworks has been leveraged for complex image/point_cloud processing, motion planning, etc. 

Several motion planning pipelines/graphs have been written to solve a variety of processes like painting, grinding, pick-n-place, etc where OMPL is one of those task in a much larger pipeline. I have personally used OMPL in variety of ways; one of those being standalone for freepspace task with constraints leveraging the constraint manifold features of OMPL along with different trajectory level optimization function. I have also used it to provide a coarse seed to TrajOpt to further optimize. In the end it came down to what is the tool or combination of tools to solve the motion planning problem.

PR #1240, opened 2026-01-02
Fix environment memory leak

spinningthelock:
Fixes memory leak in Tesseract Environment. SharedPtr will not slice  when downcasted. SharedPtr of command will hold the all the information of the derived command not only the base Command class as expected. This results in continuous large memory usage as command buffer is never cleared. Proposed fix insures only the lightweight Command class is retained in the command buffer.

### DIFF
--- a/tesseract_environment/src/environment.cpp
+++ b/tesseract_environment/src/environment.cpp
@@ -1290,7 +1290,7 @@ bool Environment::Implementation::applyAddCommand(const AddLinkCommand::ConstPtr
     return false;
 
   ++revision;
-  commands.push_back(cmd);
+  commands.push_back(std::make_shared<const Command>(*cmd));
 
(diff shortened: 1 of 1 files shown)


arjuna-fourgrowers:
I dint know that tesseract-ros toMsg will up-cast the history. I will close this PR.

rjoomen:
The cloning and applying an updated collision geometry is what we do as well. With a complicated environment this cloning is not cheap though, so having a way of squashing the command history would be a good addition to Tesseract.
Also, imagine a very long running setup, where this semi-fixed environment that is cloned for the actual planning work still regularly receives some modifications, e.g. because some setpoints were changed. This would still lead to a large, not very useful command history with a lot similar commands and no way of clearing/squashing it.

PR #1241, opened 2026-01-08
Add uuid to link class

rjoomen:
Last Tesseract dev meeting we talked about using uuids to reference links in all of Tesseract, instead of strings. Reason is that links are referenced a lot by name (a string) for collision checking etc. and string operations are quite expensive (consider all the OrderedLinkPair creating and comparing). Replacing this by a uuid (16 bytes, numerical) should give a measurable performance boost.

This PR just adds a uuid to the Link class as a first start.


rjoomen:
Per the discussion in the Tesseract meeting I'll change this to use the hash of the name instead of a uuid.

PR #1242, opened 2026-01-08
Add hash to link class

rjoomen:
Last Tesseract dev meeting we talked about using hashes to reference links in all of Tesseract, instead of strings. Reason is that links are referenced a lot by name (a string) for collision checking etc. and string operations are quite expensive (consider all the OrderedLinkPair creating and comparing). Replacing this by a hash (size_t) should give a measurable performance boost.

This PR just adds a hash to the Link class as a first start.

(Replaces #1241)

### DIFF
--- a/tesseract_scene_graph/include/tesseract_scene_graph/link.h
+++ b/tesseract_scene_graph/include/tesseract_scene_graph/link.h
@@ -195,7 +195,7 @@ public:
 
   const std::string& getName() const;
 
-  const std::size_t& getHash() const;
+  std::size_t getHash() const;
 
   /// inertial element
   Inertial::Ptr inertial;
--- a/tesseract_scene_graph/include/tesseract_scene_graph/link.h
+++ b/tesseract_scene_graph/include/tesseract_scene_graph/link.h
@@ -183,7 +183,7 @@ public:
   using Ptr = std::shared_ptr<Link>;
   using ConstPtr = std::shared_ptr<const Link>;
 
-  Link(std::string name);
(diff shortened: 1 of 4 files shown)


tesseract_scene_graph/include/tesseract_scene_graph/link.h
```diff
@@ -195,6 +195,8 @@ class Link
 
   const std::string& getName() const;
 
+  const size_t& getHash() const;
```
Levi-Armstrong: I would use `std::size_t`.

Levi-Armstrong (changes requested):
Should we also add a hash to the Joint class?

tesseract_scene_graph/include/tesseract_scene_graph/link.h
```diff
@@ -195,6 +195,8 @@ class Link
 
   const std::string& getName() const;
 
+  const std::size_t& getHash() const;
```
marip8:
```suggestion
  std::size_t getHash() const;
```

I'm not sure there's really any reason to return a reference here instead of a copy of the hash value. I guess you save a few bytes, but it seems better to not return a reference

rjoomen: Good one. Done.

marip8 (approved):
Looks good to me

Levi-Armstrong (approved):

rjoomen:
Continued in https://github.com/tesseract-robotics/tesseract/issues/1275

PR #1245, opened 2026-01-29
Add macro to allow disabling use of thread_local

Levi-Armstrong:
@johnwason, @rjoomen and @marip8  I added the ability to disable use of thread_local through tesseract and trajopt. It seems to have a significant impact on the number of allocation and reduce the peak slightly. Most of the Peak is a result of the thread local in the collision evaluators in trajopt_ifopt. Also I did not seem much difference in performance with or without this enable. Do you have any thoughts on the subject?

**Without use of thread_local:**
image

**With use of thread_local:**
image


johnwason:
Also I use `boost::string_ref` or my own custom string classes in a lot of places to avoid allocating strings in the API for Robot Raconteur.

PR #1250, opened 2026-02-17
Remove unused field active from ContactTestData

rjoomen:
Used nowhere.


tesseract_collision/bullet/src/bullet_cast_bvh_manager.cpp
```diff
@@ -369,7 +369,6 @@ const std::vector<std::string>& BulletCastBVHManager::getCollisionObjects() cons
 void BulletCastBVHManager::setActiveCollisionObjects(const std::vector<std::string>& names)
 {
   active_ = names;
-  contact_test_data_.active = &active_;
```
Levi-Armstrong: If its not used should active be removed from contact_test_data object

rjoomen: Sorry, I don't understand your question.

rjoomen: It's used nowhere, except in some assertions. Everything is removed is this PR, and it builds correctly.

PR #1272, opened 2026-04-07
Leverage schema for plugins throughout

Levi-Armstrong:
## Summary

This PR makes `PropertyTree` schemas the authoritative validation and metadata layer for collision and kinematics plugin configuration.

It adds schema composition and plugin-specific schema helpers, registers schemas for built-in plugins, introduces safe plugin-library lifetime management, and changes plugin factories to validate configurations transactionally before committing them.

The existing YAML structure is preserved, but invalid configurations now fail earlier and are validated more strictly.

## Motivation

Plugin configuration previously had several limitations:

- Plugin-specific YAML could be decoded before the selected plugin schema was available.
- Plugin schemas registered through static initialization required the plugin library to be loaded first, creating a discovery/validation bootstrap problem.
- Registered validators could reference code from a plugin library that had already been unloaded.
- Invalid configurations often reached `create()` and returned `nullptr` instead of failing when the configuration was loaded.
- Plugin discovery metadata was not represented generically enough for schema-driven tools and editors.

### PropertyTree schema support

- Added reusable schema composition with `PropertyTreeBuilder::compose()`.
- Added inline mutually exclusive branches with `beginOneOf()` and `endOneOf()`.
- Added `minimumLength()` and `maximumLength()` string constraints.
- Added standard plugin-container schema helpers.
- Added generic plugin metadata for:
  - Outer configuration keys
  - Search paths and search libraries
  - Plugin-loader sections
  - Plugin factory base types
- Improved optional-container handling so required descendants are only validated when their parent container is present.
- Strengthened derived-type validation so a selected concrete type must have both:
  - A registered relationship to the expected base type
  - A registered concrete schema

### Safe plugin-library lifetime

Added `SchemaRegistry::loadAndRetainPluginLibraries()`.

The registry now acquires and retains shared library lifetime tokens from `boost_plugin_loader`. This ensures that validators and schema callbacks registered by a plugin cannot outlive the library containing their implementation.

Libraries are loaded before acquiring the registry mutex so static schema registration can safely call back into the registry. Registered schemas are destroyed before retained library handles during registry teardown.

### Two-stage plugin validation

Collision and kinematics factories now load configuration in two stages:

1. Permissively validate and decode only plugin discovery information.
2. Load and retain the requested plugin libraries.
3. Allow static plugin schema registration to complete.
4. Rebuild the complete schema and strictly validate the full configuration.
5. Decode and commit the candidate loader and plugin maps only after validation succeeds.

Factory state is therefore transactional: failed validation does not partially modify configured search paths, libraries, loaders, or plugin maps.

Schema registrations and retained library handles are process-wide and intentionally remain available after a later validation failure.

### Collision plugin schemas

- Added schemas for Bullet and FCL collision manager plugins.
- Added factory-family validation for discrete and continuous plugins.
- Added validation for Bullet allocator-size options.
- Added decoding for:
  - `max_custom_collision_algorithm_element_size`
  - `use_epa_penetration_algorithm`
- Invalid classes, wrong factory families, missing class names, malformed values, and unknown fields are rejected during configuration loading.

### Kinematics plugin schemas

Added schemas for:

- KDL
- OPW
- UR
- REP
- ROP
- Test IKFast plugins

The schemas validate required links, solver-specific options, array lengths, explicit robot parameters, positioner sampling configuration, and nested forward/inverse kinematics plugins.

`PositionerSampleResolution` now centralizes typed decoding of its name, value, and optional bounds.

### Plugin discovery information

Introduced `PluginDiscoveryInfo` as the common source of `search_paths` and `search_libraries` for:

- `ProfilesPluginInfo`
- `KinematicsPluginInfo`
- `ContactManagersPluginInfo`
- `TaskComposerPluginInfo`

YAML and cereal support were added for the new type.

### Documentation, examples, and dependencies

- Expanded PropertyTree documentation for composition, one-of groups, plugin helpers, strict validation, two-stage loading, and plugin-library lifetime.
- Added examples for schema composition and inline one-of validation.
- Updated `boost_plugin_loader` to `0.4.5`, including the Windows and Jammy dependency manifests.

### YAML compatibility

The existing YAML topology and key names are preserved:

- `search_paths` and `search_libraries` remain flat fields.
- Existing outer keys such as `kinematic_plugins` and `contact_manager_plugins` remain unchanged.
- Plugin entries continue to use `class` and `config`.
- Existing cereal field names are preserved.

Validation is intentionally stricter:

- Unknown fields are rejected.
- Discovery fields must be lists of strings.
- Plugin classes must belong to the expected factory family.
- Every selected concrete plugin must have a registered schema.
- Required and typed plugin configuration is validated during factory loading.
- A present plugin container must contain a `plugins` map; an empty container should use `plugins: {}`.
- Omitted optional plugin sections remain valid.

### Custom plugin migration

Custom collision and kinematics plugins must:

1. Provide a schema for the concrete plugin configuration.
2. Register that schema under the exact class alias used in YAML.
3. Register the concrete type as derived from the appropriate factory base.
4. Ensure the plugin library is discoverable before strict configuration validation.

Search paths that were previously added only after constructing a configured factory must now be supplied through YAML, environment/package defaults, or before loading the configuration.

### ABI compatibility

This PR should be treated as an **ABI-breaking change**.

Notable ABI-impacting changes include:

- New virtual `schema()` methods on collision and kinematics factory interfaces
- Changes to public PropertyTree and SchemaRegistry implementations
- Plugin information structures now using `PluginDiscoveryInfo` inheritance
- Changes to several public concrete factory interfaces

Tesseract libraries, downstream applications, and collision/kinematics plugin libraries should be rebuilt together. Existing plugin binaries should not be loaded into a host built from this change.

No existing factory `create()` signatures are removed, and the new `schema()` methods are non-pure to reduce source-level migration requirements.

## Test coverage

Tests were added or expanded for:

- Schema composition and inline one-of behavior
- String-length constraints
- Optional versus explicitly present containers
- Strict derived-type registration
- Plugin discovery metadata
- Flat YAML and cereal compatibility
- Transactional factory loading
- Malformed discovery information
- Existing collision YAML compatibility
- Bullet and FCL plugin validation
- KDL, OPW, UR, REP, ROP, and IKFast configuration validation
- Wrong factory families, unknown classes, missing classes, invalid values, and unknown fields

### DIFF
--- a/collision/bullet/include/tesseract/collision/bullet/bullet_factories.h
+++ b/collision/bullet/include/tesseract/collision/bullet/bullet_factories.h
@@ -53,6 +53,7 @@ class BulletDiscreteBVHManagerFactory : public DiscreteContactManagerFactory
 public:
   std::unique_ptr<DiscreteContactManager> create(const std::string& name,
                                                  const YAML::Node& config) const override final;
+  tesseract::common::PropertyTree schema() const override;
 };
 
 class BulletDiscreteSimpleManagerFactory : public DiscreteContactManagerFactory
@@ -60,6 +61,7 @@ class BulletDiscreteSimpleManagerFactory : public DiscreteContactManagerFactory
 public:
   std::unique_ptr<DiscreteContactManager> create(const std::string& name,
                                                  const YAML::Node& config) const override final;
+  tesseract::common::PropertyTree schema() const override;
 };
 
 class BulletCastBVHManagerFactory : public ContinuousContactManagerFactory
@@ -67,6 +69,7 @@ class BulletCastBVHManagerFactory : public ContinuousContactManagerFactory
 public:
   std::unique_ptr<ContinuousContactManager> create(const std::string& name,
                                                    const YAML::Node& config) const override final;
+  tesseract::common::PropertyTree schema() const override;
 };
 
 class BulletCastSimpleManagerFactory : public ContinuousContactManagerFactory
@@ -74,6 +77,7 @@ class BulletCastSimpleManagerFactory : public ContinuousContactManagerFactory
 public:
   std::unique_ptr<ContinuousContactManager> create(const std::string& name,
                                                    const YAML::Node& config) const override final;
+  tesseract::common::PropertyTree schema() const override;
 };
 
 PLUGIN_ANCHOR_DECL(BulletFactoriesAnchor)
--- a/collision/bullet/src/bullet_factories.cpp
+++ b/collision/bullet/src/bullet_factories.cpp
@@ -37,52 +37,104 @@ TESSERACT_COMMON_IGNORE_WARNINGS_POP
 #include <tesseract/collision/discrete_contact_manager.h>
 #include <tesseract/collision/continuous_contact_manager.h>
 
+#include <tesseract/common/schema_registration.h>
+#include <tesseract/common/property_tree.h>
+
+namespace YAML
+{
+template <>
+struct convert<tesseract::collision::TesseractCollisionConfigurationInfo>
+{
+  static bool decode(const Node& node, tesseract::collision::TesseractCollisionConfigurationInfo& rhs)
+  {
+    if (node.IsNull())
+    {
+      rhs = tesseract::collision::TesseractCollisionConfigurationInfo();
+      return true;
+    }
+
+    bool share_pool_allocators{ false };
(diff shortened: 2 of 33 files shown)


jf---:
wow heavy refactoring @Levi-Armstrong 
can you provide a little context for us reading along perhaps?
could be something of a breaking change?

PR #1274, opened 2026-04-13
Move from string-based links and joints to id-based logic (using hashes)

rjoomen:
For reference: this was squash merged, the full 259 commit history is backed up here: https://github.com/rjoomen/tesseract/commits/archive/pr-1274-pre-squash


johnwason:
@rjoomen how are you generating IDs for the links? You mentioned hash values but they may not be unique for strings.

Levi-Armstrong:
@rjoomen Is this ready to review?

rjoomen:
> @rjoomen Is this ready to review?

Yes! I wanted to add some benchmark results, and Claude is looking into increasing test coverage, but basically it's ready for review. I'm curious to hear your thoughts. If you'd like we could also discuss in a call after you've taken a good look.

rjoomen:
See benchmark results here: https://github.com/tesseract-robotics/tesseract/issues/1275

rjoomen:
Code coverage is up nicely. I'll keep fixing clang-tidy warnings, but in the meantime I'm looking forward to your reaction.

johnwason:
@rjoomen I have some ideas to deal with hash collisions inspired by open addressing hash tables. Let me provide an example before merging.

rjoomen:
@Levi-Armstrong Friendly ping. I know it's a bit a of a huge PR, but if you look at [types.h](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/common/include/tesseract/common/types.h), [types.cpp](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/common/src/types.cpp) and the [design document](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/IDENTITY_DESIGN.md) you'll see it's actually fairly straightforward.

BTW, this PR also slips in the following fixes for bugs I encountered:

| Commit | Category | Description |
|---|---|---|
| `ed4075b70` | Bug fix | Three Octree bugs: (1) custom `pruneNode` left the children array allocated, tripping `assert(children == NULL)` on octomap ≥ 1.9; (2) `parseOctree` didn't forward `prune="true"` to the `Octree` constructor, so serialization round-trips were wrong; (3) `clone()` dropped `pruned_` and `binary_octree_` fields. |
| `989617727` | Bug fix | `kdl_sub_tree_builder` used `floating_joint_values.at(id)` on a freshly-built (empty) `KDLTreeData`, throwing `std::out_of_range` for any scene with a floating joint. Switched to `operator[]` to match the sibling full-tree builder. |
| `9b211e8b3` | Bug fix (test) | Data race in `collision_multi_threaded_unit.hpp`: threads called `location[link_id]` (non-const, can insert) inside the OpenMP parallel region. Bound a `const auto&` alias and switched to `at()` for safe concurrent reads. |
| `ade2a0c73` | Bug fix (serialization) | `Inertial` cereal serialize wrote `izz` under the NVP name `"iyz"` (copy-paste duplicate of the previous line), so on load `izz` was overwritten by `iyz` and the actual `izz` value was discarded. Corrected to `cereal::make_nvp("izz", obj.izz)`. Same commit also default-initializes `Joint::axis` to `Eigen::Vector3d::UnitX()` (was uninitialized). |

Levi-Armstrong:
@rjoomen Thank you for all of the work. I think it might be good to have a meeting to discuss, but I think the Tesseract Monthly is too far out. I will set something up on the calendar to see when everyone is available. I also think @johnwason wanted to connect with you on the hashing component before this gets merged which we could discuss then also.

johnwason:
@rjoomen I would highly recommend using a typedef for the hash instead of uint64_t for future flexibility. I have a suspicion we may need to increase the width in the future.

rjoomen:
> @rjoomen I would highly recommend using a typedef for the hash instead of uint64_t for future flexibility. I have a suspicion we may need to increase the width in the future.

Done

Levi-Armstrong:
> You mentioned hash values but they may not be unique for strings.

@johnwason Is this still a concern?

Levi-Armstrong:
Should it be named in NameIdType instead of NameIdValue? I am not suggesting just yet to make the change, but I want to get feedback.

rjoomen:
> Should it be named in NameIdType instead of NameIdValue? I am not suggesting just yet to make the change, but I want to get feedback.

Makes sense, ending with Type seems to be the convention in Tesseract. But then perhaps NameIdValueType is better, since it is not the type of NameId, but of its member value_.

rjoomen:
> @rjoomen Thank you for all of the work. I think it might be good to have a meeting to discuss, but I think the Tesseract Monthly is too far out. I will set something up on the calendar to see when everyone is available. I also think @johnwason wanted to connect with you on the hashing component before this gets merged which we could discuss then also.

I'm available most of the time this week.

rjoomen:
> @Levi-Armstrong Friendly ping. I know it's a bit a of a huge PR, but if you look at [types.h](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/common/include/tesseract/common/types.h), [types.cpp](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/common/src/types.cpp) and the [design document](https://github.com/rjoomen/tesseract/blob/feature/integer-link-ids/IDENTITY_DESIGN.md) you'll see it's actually fairly straightforward.
> 
> BTW, this PR also slips in the following fixes for bugs I encountered:
> Commit 	Category 	Description
> `ed4075b70` 	Bug fix 	Three Octree bugs: (1) custom `pruneNode` left the children array allocated, tripping `assert(children == NULL)` on octomap ≥ 1.9; (2) `parseOctree` didn't forward `prune="true"` to the `Octree` constructor, so serialization round-trips were wrong; (3) `clone()` dropped `pruned_` and `binary_octree_` fields.
> `989617727` 	Bug fix 	`kdl_sub_tree_builder` used `floating_joint_values.at(id)` on a freshly-built (empty) `KDLTreeData`, throwing `std::out_of_range` for any scene with a floating joint. Switched to `operator[]` to match the sibling full-tree builder.
> `9b211e8b3` 	Bug fix (test) 	Data race in `collision_multi_threaded_unit.hpp`: threads called `location[link_id]` (non-const, can insert) inside the OpenMP parallel region. Bound a `const auto&` alias and switched to `at()` for safe concurrent reads.
> `ade2a0c73` 	Bug fix (serialization) 	`Inertial` cereal serialize wrote `izz` under the NVP name `"iyz"` (copy-paste duplicate of the previous line), so on load `izz` was overwritten by `iyz` and the actual `izz` value was discarded. Corrected to `cereal::make_nvp("izz", obj.izz)`. Same commit also default-initializes `Joint::axis` to `Eigen::Vector3d::UnitX()` (was uninitialized).

All PR'ed individually to master: #1291, #1293, #1294, #1292

collision/bullet/include/tesseract/collision/bullet/bullet_cast_simple_manager.h
```diff
@@ -143,9 +154,9 @@ class BulletCastSimpleManager : public ContinuousContactManager
 private:
   std::string name_;
   /** @brief A list of the active collision objects */
-  std::vector<std::string> active_;
+  std::unordered_set<tesseract::common::LinkId> active_;
   /** @brief A list of the collision objects */
-  std::vector<std::string> collision_objects_;
+  std::vector<tesseract::common::LinkId> collision_objects_;
```
Levi-Armstrong, line 160: Why is this not switched to using an unordered_set?

collision/bullet/include/tesseract/collision/bullet/bullet_discrete_bvh_manager.h
```diff
@@ -132,9 +147,9 @@ class BulletDiscreteBVHManager : public DiscreteContactManager
 private:
   std::string name_;
   /** @brief A list of the active collision objects */
-  std::vector<std::string> active_;
+  std::unordered_set<tesseract::common::LinkId> active_;
   /** @brief A list of the collision objects */
-  std::vector<std::string> collision_objects_;
+  std::vector<tesseract::common::LinkId> collision_objects_;
```
Levi-Armstrong, line 153: same here

collision/bullet/include/tesseract/collision/bullet/bullet_utils.h
```diff
@@ -148,8 +148,8 @@ class CollisionObjectWrapper : public btCollisionObject
 };
 
 using COW = CollisionObjectWrapper;
-using Link2Cow = std::map<std::string, COW::Ptr>;
-using Link2ConstCow = std::map<std::string, COW::ConstPtr>;
+using Link2Cow = std::unordered_map<tesseract::common::LinkId, COW::Ptr>;
```
Levi-Armstrong, line 151: Should we use std::map here given the type change?

collision/bullet/include/tesseract/collision/bullet/bullet_cast_simple_manager.h
rjoomen: active_ is switched to an unordered_set as looking up the active state of objects is done frequently, and find is O(1) on average for this container. Finding in collision_objects_ is done much less frequently, and therefore changing to unordered_set was not necessary, and vector is often more convenient.

collision/bullet/include/tesseract/collision/bullet/bullet_discrete_bvh_manager.h
rjoomen: See above

collision/bullet/include/tesseract/collision/bullet/bullet_utils.h
rjoomen: unordered_map benefits here from the fast hash-based lookup using LinkId's stored hash.

Levi-Armstrong:
@rjoomen I sent you an email, though I wanted to know what days and times next week work for you to have a meeting to review.

Levi-Armstrong:
@rjoomen I just sent out the invite for Monday.

Levi-Armstrong:
@rjoomen Based on the last developer meeting I believe we want to get this into the next release 0.36.0 but we wanted to get the following items updated. @johnwason let me know if I missed anything.

- [ ] Update so has is 128bit to reduce the chance of collisions, but we would only use 64bit for the key.
- [ ] Validate that the hash function is the right one for our use case.
- [ ] I vaguely remember discussing adding or moving where we check for collisions to catch issues.

rjoomen:
> @rjoomen Based on the last developer meeting I believe we want to get this into the next release 0.36.0 but we wanted to get the following items updated. @johnwason let me know if I missed anything.
> 
>     * [ ]  Update so has is 128bit to reduce the chance of collisions, but we would only use 64bit for the key.
> 
>     * [ ]  Validate that the hash function is the right one for our use case.
> 
>     * [ ]  I vaguely remember discussing adding or moving where we check for collisions to catch issues.

@Levi-Armstrong I'm sorry for the delay, I was on holidays. I'm going to work on this this week to get the discussed topics sorted, I'll keep you posted. I'd also like to finally get these PR's off my back.

jf---:
hi @rjoomen this is quite an engineering effort -- though on a reading of the title of the issue its hard to fathom the engineering involved. Would you mind mincing a few words on what's involved? Helps appreciate the engineering involved even more ;)

collision/core/include/tesseract/collision/common.h
```diff
@@ -75,12 +87,14 @@ bool isContactAllowed(const std::string& name1,
...
  * @param key Link pair used as a key to look up pair specific settings
+ * @param margin The collision margin already resolved for @p key
  * @param found Specifies whether or not a collision has already been found
  * @return Pointer to the ContactResult.
  */
 ContactResult* processResult(ContactTestData& cdata,
                              ContactResult& contact,
-                             const std::pair<std::string, std::string>& key,
+                             const tesseract::common::LinkIdPair& key,
+                             double margin,
```
Levi-Armstrong, line 97: The margin should be pulled from the ContactTestData; correct?

collision/core/src/continuous_contact_manager.cpp
```diff
@@ -24,9 +24,47 @@
...
+  if (ids.size() != pose1.size() || ids.size() != pose2.size())
+    throw std::runtime_error("ContinuousContactManager, setCollisionObjectsTransform received " +
+                             std::to_string(ids.size()) + " ids but " + std::to_string(pose1.size()) +
+                             " start poses and " + std::to_string(pose2.size()) + " end poses!");
+
+  for (std::size_t i = 0; i < ids.size(); ++i)
+    setCollisionObjectsTransform(ids[i], pose1[i], pose2[i]);
+}
+
+void ContinuousContactManager::setActiveCollisionObjects(const std::vector<tesseract::common::LinkId>& ids)
```
Levi-Armstrong, line 58: Was this this added for backwards compatibility? This like it could be significant performance hit if you were accidentally using it. Maybe we should remove these and force user to update.

collision/bullet/src/bullet_discrete_bvh_manager.cpp
```diff
@@ -86,16 +86,17 @@ DiscreteContactManager::UPtr BulletDiscreteBVHManager::clone() const
 {
   auto manager = std::make_unique<BulletDiscreteBVHManager>(name_, config_info_.clone());
 
-  for (const auto& cow : link2cow_)
+  for (const auto& id : collision_objects_)
   {
-    COW::Ptr new_cow = cow.second->clone();
+    const auto& orig = link2cow_.at(id);
+    COW::Ptr new_cow = orig->clone();
```
Levi-Armstrong, line 92: It looks like this change is adding an additional lookup; should this change be made?

collision/bullet/src/bullet_discrete_simple_manager.cpp
```diff
@@ -72,16 +72,17 @@ DiscreteContactManager::UPtr BulletDiscreteSimpleManager::clone() const
 {
   auto manager = std::make_unique<BulletDiscreteSimpleManager>(name_, config_info_.clone());
 
-  for (const auto& cow : link2cow_)
+  for (const auto& id : collision_objects_)
   {
-    COW::Ptr new_cow = cow.second->clone();
+    const auto& orig = link2cow_.at(id);
+    COW::Ptr new_cow = orig->clone();
```
Levi-Armstrong, line 78: Same here.

collision/bullet/src/bullet_utils.cpp
```diff
@@ -1064,7 +917,9 @@ btScalar addCastSingleResult(btManifoldPoint& cp,
   contact.distance = static_cast<double>(cp.m_distance1);
   contact.normal = convertBtToEigen(-1 * cp.m_normalWorldOnB);
 
-  ContactResult* col = processResult(collisions, contact, key, found);
+  const double security_margin = collisions.collision_margin_data.getCollisionMargin(key);
+
+  ContactResult* col = processResult(collisions, contact, key, security_margin, found);
```
Levi-Armstrong, line 1082: Is there a benefit to doing this outside of the processResult function where previous this happened inside because it had access to the collisions object?

collision/bullet/src/bullet_utils.cpp
Levi-Armstrong, line 1149: I am just going to flag this, so I dig deeper later.

collision/fcl/include/tesseract/collision/fcl/fcl_discrete_managers.h
```diff
@@ -61,43 +66,57 @@ class FCLDiscreteBVHManager : public DiscreteContactManager
...
+  Eigen::Isometry3d getCollisionObjectsTransform(const tesseract::common::LinkId& id) const override final;
 
-  void setCollisionObjectsTransform(const tesseract::common::TransformMap& transforms) override final;
+  void setCollisionObjectsTransform(const tesseract::common::LinkIdTransformMap& transforms) override final;
+
+  void setCollisionObjectsTransform(const std::vector<tesseract::common::LinkId>& ids,
+                                    const tesseract::common::VectorIsometry3d& poses) override final;
 
-  const std::vector<std::string>& getCollisionObjects() const override final;
+  const std::vector<tesseract::common::LinkId>& getCollisionObjects() const override final;
```
Levi-Armstrong, line 115: Is there a reason to keep both getCollisionObjects and getActiveCollisionObjectIds?

collision/fcl/src/fcl_discrete_managers.cpp
```diff
@@ -61,8 +63,8 @@ DiscreteContactManager::UPtr FCLDiscreteBVHManager::clone() const
 {
   auto manager = std::make_unique<FCLDiscreteBVHManager>();
 
-  for (const auto& cow : link2cow_)
-    manager->addCollisionObject(cow.second->clone());
+  for (const auto& id : collision_objects_)
+    manager->addCollisionObject(link2cow_.at(id)->clone());
```
Levi-Armstrong, line 67: This change seems to add an extra lookup; should this change be reverted?

collision/fcl/src/fcl_discrete_managers.cpp
```diff
@@ -135,16 +138,17 @@ bool FCLDiscreteBVHManager::removeCollisionObject(const std::string& name)
         dynamic_manager_->unregisterObject(co.get());
     }
 
-    collision_objects_.erase(std::find(collision_objects_.begin(), collision_objects_.end(), name));
-    link2cow_.erase(name);
+    collision_objects_.erase(std::find(collision_objects_.begin(), collision_objects_.end(), id));
+    link2cow_.erase(it);
+    active_.erase(id);
```
Levi-Armstrong, line 143: Good catch!

collision/bullet/src/bullet_cast_bvh_manager.cpp
```diff
@@ -90,16 +90,17 @@ ContinuousContactManager::UPtr BulletCastBVHManager::clone() const
 {
   auto manager = std::make_unique<BulletCastBVHManager>(name_, config_info_.clone());
 
-  for (const auto& cow : link2cow_)
+  for (const auto& id : collision_objects_)
   {
-    COW::Ptr new_cow = cow.second->clone();
+    const auto& orig = link2cow_.at(id);
+    COW::Ptr new_cow = orig->clone();
```
Levi-Armstrong, line 96: Same here; why not still loop over link2cow_?

collision/bullet/src/bullet_cast_bvh_manager.cpp
```diff
@@ -134,47 +135,52 @@ bool BulletCastBVHManager::addCollisionObject(const std::string& name,
...
+    collision_objects_.erase(std::find(collision_objects_.begin(), collision_objects_.end(), id));
     removeCollisionObjectFromBroadphase(cow1, broadphase_, dispatcher_);
-    link2cow_.erase(name);
+    link2cow_.erase(it);
 
-    COW::Ptr& cow2 = link2castcow_[name];
-    removeCollisionObjectFromBroadphase(cow2, broadphase_, dispatcher_);
-    link2castcow_.erase(name);
+    auto cast_it = link2castcow_.find(id);
+    if (cast_it != link2castcow_.end())
```
Levi-Armstrong: I would add an assert or throw and exception because I do not think this should ever happen

collision/test_suite/include/tesseract/collision/test_suite/collision_box_capsule_unit.hpp
```diff
@@ -140,7 +147,7 @@ inline void runTest(DiscreteContactManager& checker)
   EXPECT_NEAR(result_vector[0].nearest_points[0][2], result_vector[0].nearest_points[1][2], 0.001);
 
   std::vector<int> idx = { 0, 1, 1 };
-  if (result_vector[0].link_names[0] != "box_link")
+  if (result_vector[0].link_ids[0] != "box_link")
```
Levi-Armstrong, line 150: If these are still strings; should this be link_names instead of link_ids?

common/cmake/tesseract_macros.cmake
```diff
@@ -72,7 +72,11 @@ macro(tesseract_variables)
     set(TESSERACT_ENABLE_RUN_BENCHMARKING OFF)
   endif()
 
-  set(TESSERACT_COMPILE_DEFINITIONS "")
+  set(TESSERACT_COMPILE_DEFINITIONS_PUBLIC "")
+  # Definitions applied PRIVATE to Tesseract's own production libraries so they affect only the libraries' own
+  # translation units, not their consumers (tests, examples, downstream users). Currently forces explicit string ->
+  # NameId construction inside the libraries while leaving implicit conversion available everywhere else.
+  set(TESSERACT_COMPILE_DEFINITIONS_PRIVATE "TESSERACT_NAMEID_NO_IMPLICIT")
```
Levi-Armstrong, line 79: Is this currently used?

common/include/tesseract/common/joint_state.h
```diff
@@ -32,17 +32,22 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
 
-  /** @brief The joint corresponding to the position vector. */
-  std::vector<std::string> joint_names;
+  JointState(std::vector<JointId> joint_ids, const Eigen::Ref<const Eigen::VectorXd>& position);
+
+  /** @brief The joint IDs corresponding to the position vector. */
+  std::vector<JointId> joint_ids;
+
+  /** @brief Get the joint IDs */
+  const std::vector<JointId>& getJointIds() const;
```
Levi-Armstrong: Is this method needed?

common/include/tesseract/common/types.h
```diff
@@ -26,47 +27,290 @@
...
+// which is convenient for tests and downstream users. Tesseract's own
+// production libraries set TESSERACT_NAMEID_NO_IMPLICIT (via target_compile_-
+// definitions PRIVATE) to make the constructors explicit in their own TUs,
+// forcing internal call sites to either avoid string-keyed paths entirely or
+// be explicit about when they construct an id from a string. Templates are
+// instantiated per-TU, so the check is compile-time only. Formally the
+// differing token sequences across TUs are an ODR violation (ill-formed, no
+// diagnostic required) — a deliberate, benign one: explicit affects overload
+// resolution only, not layout or codegen. C++20 modules would reject it.
+// ---------------------------------------------------------------------------
```
Levi-Armstrong, line 52: Why would we not require this to be explicit all of the time?

environment/include/tesseract/environment/environment_monitor_interface.h
```diff
@@ -116,36 +117,60 @@ class EnvironmentMonitorInterface
...
+                                   const std::vector<tesseract::common::JointId>& joint_ids,
+                                   const std::vector<double>& joint_values,
+                                   const tesseract::common::JointIdTransformMap& floating_joints = {}) const
+  {
+    return setEnvironmentState(
+        monitor_namespace,
+        joint_ids,
+        Eigen::Map<const Eigen::VectorXd>(joint_values.data(), static_cast<Eigen::Index>(joint_values.size())),
+        floating_joints);
+  }
```
Levi-Armstrong: I would move implementation to cpp

environment/include/tesseract/environment/environment_monitor_interface.h
```diff
@@ -116,36 +117,60 @@ class EnvironmentMonitorInterface
...
+  virtual std::vector<std::string>
+  setEnvironmentState(const std::vector<tesseract::common::JointId>& joint_ids,
+                      const std::vector<double>& joint_values,
+                      const tesseract::common::JointIdTransformMap& floating_joints = {}) const
+  {
+    return setEnvironmentState(
+        joint_ids,
+        Eigen::Map<const Eigen::VectorXd>(joint_values.data(), static_cast<Eigen::Index>(joint_values.size())),
+        floating_joints);
+  }
```
Levi-Armstrong: I would move the implementation to the cpp

environment/src/environment.cpp
```diff
@@ -751,14 +738,14 @@ std::vector<std::string> Environment::Implementation::getGroupJointNames(const s
     tesseract::scene_graph::ShortestPath path =
         scene_graph->getShortestPath(chain_it->second.begin()->first, chain_it->second.begin()->second);
 
-    group_joint_names_cache[group_name] = path.active_joints;
-    return path.active_joints;
+    group_joint_ids_cache[group_name] = std::move(path.active_joints);
+    return group_joint_ids_cache[group_name];
```
Levi-Armstrong: Is there any reason to no just return path.active_joints because this change requires a lookup?

visualization/include/tesseract/visualization/ignition/entity_manager.h
```diff
@@ -30,11 +30,14 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
 #include <unordered_map>
 TESSERACT_COMMON_IGNORE_WARNINGS_POP
 
+#include <tesseract/common/types.h>
+
 namespace tesseract::visualization
 {
 static const int NULL_ENTITY_ID = -1;
 using EntityID = int;
 using EntityMap = std::unordered_map<std::string, EntityID>;
```
Levi-Armstrong, line 39: Is this needed anymore?

visualization/include/tesseract/visualization/markers/axis_marker.h
```diff
@@ -12,7 +14,7 @@ class AxisMarker : public Marker
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
   AxisMarker() = default;
-  AxisMarker(const Eigen::Isometry3d& axis) : axis(axis) {}
+  AxisMarker(Eigen::Isometry3d axis) : axis(std::move(axis)) {}
```
Levi-Armstrong: Eigen types must be passed by reference

Levi-Armstrong (commented):
@rjoomen I would ignore these comments for now. Most were me flagging things with questions which I will pull the code and go through them. I will have feedback this week, but I did not see anything major.

rjoomen:
> @rjoomen I would ignore these comments for now. Most were me flagging things with questions which I will pull the code and go through them. I will have feedback this week but so I did not see anything major.

Great that you're looking at it! I'm really a the final stage of this, final huge review with Claude, tuning odds and ends. I was hoping to mark all of it ready for review this week. With some updated benchmarks, updated design document, migration doc. And I've found a way to deal with the collision risk, it's a non-issue now, just like it is not an issue for hash maps. So our whole 128-bit hash discussion is not relevant anymore.

I'll check your comments nonetheless, some I might be able to easily answer, helping you along with the review process.

Levi-Armstrong:
@rjoomen Should I wait till you push your final changes before pulling down and building to do my final review?

collision/core/include/tesseract/collision/common.h
rjoomen: It could be, but the reason it's a parameter is the other backends rather than the Bullet hunk you're looking at here.

All three FCL call sites have already computed the margin in order to configure the query itself — `fcl_utils.cpp:288` looks the margin up and `:291` sets `config.contact_margin = security_margin`, before calling `processResult` at `:324`; `:422` looks it up and `:423` uses it for an early-out (`if (d > security_margin)`) before calling at `:457`. Previously those paths did the pair lookup a second time inside `processResult` for every contact.

The Coal backend is the clearest case: `coal_utils.cpp:720` looks the margin up once per link pair and uses it at `:781`/`:782` to set the request's `security_margin` and `distance_upper_bound`, then passes it to `processResult` at `:850` inside the per-contact loop. Pulling the lookup back inside would put it in that innermost loop. More generally, the parameter is what lets a backend hand over a margin it had to compute anyway.

Bullet (`:937`, `:1079`) does have to compute it, so there it's a wash. The one regression is that when neither `calculate_distance` nor `calculate_penetration` is set, Bullet now does a lookup the guarded version skipped — but both default to `true`, so that's not a path anyone hits in practice.

collision/core/src/continuous_contact_manager.cpp
rjoomen: Not backwards compatibility — these take `std::vector<LinkId>`, not strings, so there's no implicit-conversion trap and no silent slow path to fall into.

Upstream these were pure virtual on the base and every manager implemented them with the identical loop; `BulletCastBVHManager::setCollisionObjectsTransform(names, poses)` at `bullet_cast_bvh_manager.cpp:258-264` was byte-for-byte the same as the others. All this does is hoist that duplicated loop into the base class and upgrade the size-mismatch `assert` to a `throw`. Same work, same cost as before.

They do need to stay virtual: `FCLDiscreteBVHManager` overrides the discrete one (`fcl_discrete_managers.cpp:245`) so it can batch the BVH rebalance instead of rebalancing per object.

collision/bullet/src/bullet_discrete_bvh_manager.cpp
rjoomen: It is an extra lookup. The reason isn't visible in this hunk — the container type changed underneath: `Link2Cow` went from `std::map<std::string, COW::Ptr>` to `std::unordered_map<LinkId, COW::Ptr>`.

On the container change itself: these went to `unordered_map` for the lookup performance — a `LinkId` carries its hash, so lookups are O(1) rather than O(log n). Nothing was given up in the trade: `NameId::operator<` compares the hash rather than the name, so a `std::map<LinkId, …>` would only have implied an ordering that doesn't actually exist. These are lookup tables with no ordering consumer.

That's what makes the `clone()` change necessary. Upstream's `std::map` walk was name-ordered, so `clone()` rebuilt `collision_objects_` deterministically. Iterating `link2cow_` now would hand the clone a `collision_objects_` in a different order from its source — and that vector is returned publicly by `getCollisionObjects()`. It also changes broadphase insertion order, so BVH shape and contact tie-break order would differ between a manager and its clone.

Iterating `collision_objects_` and paying one `.at()` per object makes clone == source again. One hash lookup per object per `clone()`, not on the per-check path. It doesn't restore upstream's alphabetical order — nothing could, now — but it restores determinism and source/clone equality.

collision/fcl/include/tesseract/collision/fcl/fcl_discrete_managers.h
rjoomen: Both were in master already and they mean different things, so both should stay — `getCollisionObjects()` is every registered object, `getActiveCollisionObjects()` is the movable subset. Nothing was added here.

You're right that renaming the second one was inconsistent, though, so I've put the name back. It also now matches its own setter, `setActiveCollisionObjects`, which I'd left alone.

What did change is the container: `const std::vector<std::string>&` became `const std::unordered_set<LinkId>&`. The active set is a pure membership test — `isLinkActive()` was a linear `std::find` over the vector for every collision object and is now an O(1) `count` — and it can't hold duplicates by construction, which the vector could.

The only thing that changes for callers is iteration order, which is now unspecified. `operator[]`, `at()` and `.data()` are all compile errors on the new type, so order-dependent iteration is the one way to get bitten silently; it's called out in the migration guide.

collision/bullet/src/bullet_cast_bvh_manager.cpp
rjoomen: Added the assert. The invariant does hold — `addCollisionObject(const COW::Ptr&)` writes `link2cow_[lid]` and `link2castcow_[lid]` unconditionally in the same function — so I replaced the `if` with it rather than keeping both:

```cpp
auto cast_it = link2castcow_.find(id);
assert(cast_it != link2castcow_.end());
removeCollisionObjectFromBroadphase(cast_it->second, broadphase_, dispatcher_);
link2castcow_.erase(cast_it);
```

The guard was mine, not upstream's — master uses a bare `link2castcow_[name]` here. I left the other three `operator[]` sites in this file as-is, since the same invariant covers them and they match master.

collision/test_suite/include/tesseract/collision/test_suite/collision_box_capsule_unit.hpp
rjoomen: They aren't strings — `ContactResult::link_ids` is `std::array<LinkId, 2>`. It was `std::array<std::string, 2> link_names` in master, and got renamed along with the type change, since `_names` would be misleading for a member holding ids.

What's confusing here is the right-hand side: `!= "box_link"` still compiles because `NameId` has an implicit constructor from `const char*` in test translation units. Tests and examples get only `TESSERACT_COMPILE_DEFINITIONS_PUBLIC`; the production libraries add `TESSERACT_NAMEID_NO_IMPLICIT` privately, so the same line inside a library would be a compile error. That's what keeps over a thousand lines in this repo alone building unchanged — it's the subject of your `types.h` comment as well.

common/cmake/tesseract_macros.cmake
rjoomen: Yes — 28 `target_compile_definitions(... PRIVATE ...)` calls across 16 CMakeLists, i.e. every production library. It currently carries a single define, `TESSERACT_NAMEID_NO_IMPLICIT`.

It follows the pattern master already uses for compile *options*: `TESSERACT_COMPILE_OPTIONS_PUBLIC` / `..._PRIVATE`, where the strict warnings (`-Wall -Wextra -Wconversion`, `/bigobj`) go PRIVATE so they bind our own translation units without being inflicted on consumers. Compile *definitions* were a single undifferentiated variable in master; this splits them the same way — the platform defines keep their reach as `..._PUBLIC`, and `..._PRIVATE` is the new half.

Because it's PRIVATE it doesn't propagate: tests, examples, benchmarks **and downstream users** all keep implicit `NameId` conversion, while the libraries themselves have to be explicit. That's the mechanism behind the `link_ids[0] != "box_link"` line you flagged — strictness we impose on ourselves, not on our users.

common/include/tesseract/common/joint_state.h
rjoomen: Agreed, removed. It was never upstream API — it came in paired with a `getJointNames()` compat shim, and when that shim went, its twin was left with nothing to be symmetric with. `joint_ids` is a public member, so the accessor added nothing.

Every call site read the member before the accessor existed, including the ones in tesseract_planning and tesseract_qt, so they're back to `.joint_ids` and match upstream again. `JointStateByJointIdConstructorUnit` went with it — with the string ctor gone, `JointStateTest` *is* that test.

common/include/tesseract/common/types.h
rjoomen: Short version: I'd keep it for now and revisit once the id migration PRs land.

The implicit constructors outside our own libraries are why string call sites in tests, examples and downstream code still compile untouched — a large part of why this diff is reviewable, and it keeps development on the branch from being constant mechanical churn. It's also a real convenience at the public boundary: `env.getLinkTransform("base")` still just works.

The cost, to be straight about it: the toggle gives `NameId` different definitions across TUs, which is formally an ODR violation — IFNDR. Benign in that `explicit` only affects overload resolution and changes no layout, mangling or codegen, but it is non-conforming, and C++20 modules would reject it. It's also not a shortcut over something tidier — converting constructors are members, so "explicit inside, implicit outside" can't be expressed with a single class definition at all.

Making them implicit everywhere would be conforming, but that's a bad idea: inside the library the explicit form is what stops us silently reintroducing string-keyed lookups on hot paths, which is the cost this migration exists to remove.

So the real choice is between the toggle and going explicit everywhere. The latter is large but purely mechanical, and much cheaper once the id migration PRs have landed than while they're in flight — happy to do it then if that's where you land.

One direction worth exploring when we get there: make only the `const char*` constructor implicit and keep the `std::string` one explicit. That's a single class definition in every TU, so the ODR issue goes away outright rather than being mitigated. A literal would still convert on its own, so `env.getLinkTransform("base")` keeps working everywhere, while code holding a name in a `std::string` variable would have to write `LinkId(name)` — and that's the case actually worth flagging, since a name in a variable is where string-keyed lookups creep back in.

environment/include/tesseract/environment/environment_monitor_interface.h
rjoomen: Moved both to `environment_monitor_interface.cpp` — they were only in the header as a leftover of an intermediate migration state, back when they were string→id delegating shims.

environment/src/environment.cpp
rjoomen: Good catch, though the trade wasn't quite what it looked like — the return is by value, so it swapped a vector copy for a hash lookup rather than adding one. Neither form was the good one; both branches now use:

```cpp
return group_joint_ids_cache.emplace(group_name, std::move(path.active_joints)).first->second;
```

One lookup, one move, one copy. The `joint_groups` branch below got the same treatment — it was caching a copy and returning a copy of the source.

visualization/include/tesseract/visualization/ignition/entity_manager.h
rjoomen: Yes — `EntityMap` and `model_id_map_` / `visual_id_map_` / `sensor_id_map_` are unchanged from master; only `link_id_map_` moved, to the new `EntityLinkMap`. Model, visual and sensor names are arbitrary entity names rather than link or joint names, so there's no `NameId` for them to become.

visualization/include/tesseract/visualization/markers/axis_marker.h
rjoomen: Fixed — `AxisMarker` is back to `const Eigen::Isometry3d&`.

It came from a `modernize-pass-by-value` autofix, and `KinGroupIKInput` in `kinematic_group.h` had the same thing — there the sweep deleted master's existing `// NOLINTNEXTLINE(modernize-pass-by-value)` and converted the signature. Both are reverted and carry the NOLINT again so it can't recur.

collision/bullet/src/bullet_utils.cpp
rjoomen: Same answer as on `common.h` — the margin is a parameter so the FCL and Coal paths can hand over a value they already computed for the query itself, instead of repeating the pair lookup per contact: https://github.com/tesseract-robotics/tesseract/pull/1274#discussion_r3724562108

collision/bullet/src/bullet_discrete_simple_manager.cpp
rjoomen: Same as the BVH manager — the container under this is now `std::unordered_map<LinkId, COW::Ptr>`, so iterating it would give the clone a different object order from its source: https://github.com/tesseract-robotics/tesseract/pull/1274#discussion_r3724562582

collision/fcl/src/fcl_discrete_managers.cpp
rjoomen: Same reason as the Bullet managers — FCL's `Link2COW` is also an `unordered_map<LinkId, …>` now, so `clone()` iterates `collision_objects_` to keep clone == source: https://github.com/tesseract-robotics/tesseract/pull/1274#discussion_r3724562582

collision/bullet/src/bullet_cast_bvh_manager.cpp
rjoomen: Same as the discrete managers: looping `link2cow_` would rebuild `collision_objects_` in hash order rather than the source's order: https://github.com/tesseract-robotics/tesseract/pull/1274#discussion_r3724562582

environment/include/tesseract/environment/environment_monitor_interface.h
rjoomen: Moved this one to the .cpp as well: https://github.com/tesseract-robotics/tesseract/pull/1274#discussion_r3724564197

rjoomen:
> @rjoomen Should I wait till you push your final changes before pulling down and building to do my final review?
> 
> 

Only some small fixes to the hashing and the serialization pending, so feel free to start reviewing. I responded to the review comments you already posted. A few nice catches, even after all the reviewing I did myself. Thanks for that.

visualization/include/tesseract/visualization/markers/axis_marker.h
Levi-Armstrong: Yea, Clang-tidy sometime does not recommend the correct for Eigen Types. I hope this will be fixed in newer versions.

rjoomen:
@Levi-Armstrong Just committed the last work. You can go ahead with the review.

Levi-Armstrong:
@rjoomen Do you have time to address the conflicts?

rjoomen:
> @rjoomen Do you have time to address the conflicts? 

Not until Monday, unfortunately, but I'll do them first thing then.

And if you want to discuss anything about the PRs in a call, I'm also available anytime next week.

Levi-Armstrong:
@rjoomen I rebased without seeing any conflicts and pushed but it is still saying there are conflicts which I have never seen before.

rjoomen:
> @rjoomen I rebased without seeing any conflicts and pushed but it is still saying there are conflicts which I have never seen before.

It probably says so for the 'rebase and merge' option, not for the 'squash merge' one. Reason is the PR contains a merge commit from master somewhere.

Levi-Armstrong:
Do you want yo keep the commit history or are you okay with if I squash merge it?

rjoomen:
> Do you want yo keep the commit history or are you okay with if I squash merge it?

I'm fine with squash merging. The ~250 commits are sometimes a wandering path with a lot of fixups and cleanups, i.e. not really a nice linear development history worth keeping.

rjoomen:
> hi @rjoomen this is quite an engineering effort -- though on a reading of the title of the issue its hard to fathom the engineering involved. Would you mind mincing a few words on what's involved? Helps appreciate the engineering involved even more ;)

@jf--- Sorry for never getting back to you. Please refer to the [main issue tracking this](https://github.com/tesseract-robotics/tesseract/issues/1275). It has some benchmark results and links to the design doc with a full explanation.

rjoomen:
> @rjoomen Based on the last developer meeting I believe we want to get this into the next release 0.36.0 but we wanted to get the following items updated. @johnwason let me know if I missed anything.
> 
>     * [ ]  Update so has is 128bit to reduce the chance of collisions, but we would only use 64bit for the key.
> 
>     * [ ]  Validate that the hash function is the right one for our use case.
> 
>     * [ ]  I vaguely remember discussing adding or moving where we check for collisions to catch issues.

Just for future reference, for anyone seeing unchecked boxes in this PR: Hashes stay at size_t (64 bit usually), as this is the hash map default for C++, and hash collisions are dealt with using a `hybrid equality` scheme that completely solves the collision problem, analogous to how hash maps do this internally. For more details, see the design doc IDENTITY_DESIGN.md

PR #1279, opened 2026-04-20
Fix sign-conversion warnings in geometry/conversions.cpp

johnwason:
`tesseract_geometry` library was missing `TESSERACT_COMPILE_OPTIONS_PRIVATE` so sign conversion warnings were not detected. This PR enables the missing compile options and fixes sign conversion errors.

### DIFF
--- a/geometry/src/conversions.cpp
+++ b/geometry/src/conversions.cpp
@@ -30,6 +30,7 @@
 #include <tesseract/geometry/impl/cylinder.h>
 #include <tesseract/geometry/impl/cone.h>
 #include <tesseract/geometry/impl/capsule.h>
+#include <boost/numeric/conversion/cast.hpp>
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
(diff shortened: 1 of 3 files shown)


geometry/src/conversions.cpp
```diff
@@ -30,6 +30,7 @@
 #include <tesseract/geometry/impl/cylinder.h>
 #include <tesseract/geometry/impl/cone.h>
 #include <tesseract/geometry/impl/capsule.h>
+#include <boost/numeric/conversion/cast.hpp>
```
Levi-Armstrong: Can we use static_cast to avoid boost?

johnwason: boost::numeric_cast does bounds checking while static_cast does not.

johnwason: I added a simple implementation of `numeric_cast` to `common/utils.h` and removed the boost header file. We should be using something like `numeric_cast` to cast between numeric types. Unchecked numeric casting is a significant security issue.

Levi-Armstrong: What are your thoughts on using the gsl library or should we start using c++20?

johnwason: My concern right now is getting things running asap. We can replace the implementation of numeric_cast in the future if there is a better option.

johnwason: Also c++20 support is still not great so I wouldn't recommend requiring it.

Levi-Armstrong (approved):

Issue #1277, opened 2026-04-20
Use stdint in property_tree


johnwason:
Currently property tree is using integer types like `int`, `unsigned int`, `long`, and `unsigned long`. However, these are actually not standard and differ on platforms. `long` on Linux is 64 bit, but only 32 bit on Windows (and possibly Mac OS depending on the compiler). Use `stdint` types like `int32_t` and `uint64_t` so we have guaranteed integer sizes on different platforms.

Levi-Armstrong:
This has been fixed by #1282

PR #1273, opened 2026-04-08
Windows build fixes

johnwason:
I have started working on fixing the Windows build. I am seeing lots of these errors from the tests:

```
unknown file: error: C++ exception with description "Trying to save an unregistered polymorphic type (class tesseract::common::AnyWrapper<class tesseract::common::JointState>).
Make sure your type is registered with CEREAL_REGISTER_TYPE and that the archive you are using was included (and registered with CEREAL_REGISTER_ARCHIVE) prior to calling CEREAL_REGISTER_TYPE.
If your type is already registered and you still see this error, you may need to use CEREAL_REGISTER_DYNAMIC_INIT." thrown in the test body.
```

PR #1282, opened 2026-04-22
Use stdint types for property tree

johnwason:


johnwason:
Initial work on using stdint types in property tree as discussed in #1273

marip8:
Is this a problem only in the property tree class, or in all of `tesseract` generally?

johnwason:
> Is this a problem only in the property tree class, or in all of tesseract generally?

@marip8 Not sure. On Lindows `long` is 32 bits wide while on Linux it is 64 bits wide. You need to use `long long` for 64 bit wide integers. I think `long` might be 32 bits wide on 32 bit systems on Linux. The simple solution is to use stdint types like `int64_t` and `int32_t` instead. C and C++ do not guarantee the width of numeric types so it is best practice to not use `short`,  `int`, and `long` directly.

common/include/tesseract/common/property_tree.h
```diff
@@ -375,10 +375,10 @@ class PropertyTreeBuilder
...
-  PropertyTreeBuilder& integer(std::string_view name);
-  PropertyTreeBuilder& unsignedInt(std::string_view name);
-  PropertyTreeBuilder& longInt(std::string_view name);
-  PropertyTreeBuilder& longUnsignedInt(std::string_view name);
+  PropertyTreeBuilder& int32(std::string_view name);
+  PropertyTreeBuilder& uint32(std::string_view name);
+  PropertyTreeBuilder& int64(std::string_view name);
+  PropertyTreeBuilder& uint64(std::string_view name);
   PropertyTreeBuilder& floatNum(std::string_view name);
   PropertyTreeBuilder& doubleNum(std::string_view name);
```
Levi-Armstrong: I like how this looks. Is there similar thing we could do for float/double?

johnwason: The options are `double`, `single` or `float32`, `float64`. I use `double`, `single` in my code to match MATLAB but `float32`, `float64` are more precise.

johnwason: I change property tree to use `float32`, `float64` naming conventions. These type names don't exist in C++ so we still use `float` and `double` variable types.

Levi-Armstrong:
I like the changes! It looks cleaner and consistent.

johnwason:
Fixes #1277

Levi-Armstrong (approved):

PR #1291, opened 2026-05-01
Fix three Octree bugs: pruneNode lifetime, clone field-drop, parseOctree flag

rjoomen:
## Summary

Three independent bugs in the Octree code path:

- **`Octree::pruneNode` assertion on octomap ≥ 1.9.** The custom impl deleted children individually via `deleteNodeChild()` but never freed the `children` array, tripping `assert(children == NULL)` in `~OcTreeDataNode` on destruction (under debug asserts; silent leak under `NDEBUG`). Fixed by equalizing child values so octomap's own `pruneNode` accepts them, then delegating to it — octomap has friend access to free the array.
- **`Octree::clone` dropped `pruned_` and `binary_octree_`.** Only forwarded `octree_` and `sub_type_` to the copy constructor. Clones miscompared against their originals (`operator==` checks `pruned_`) and serialized the wrong values (cereal emits both, and `binary_octree_` controls `writeBinary` vs `write`).
- **`parseOctree` dropped the `prune` flag.** When XML had `prune="true"`, the tree was pruned but the `Octree` geometry was constructed without the flag (defaulted to `false`). This caused `writeOctomap` to emit `prune="false"`, cereal to serialize the wrong value, and round-trip `operator==` to miscompare.

Split into two commits to keep the lifetime fix separate from the field-forwarding fixes.

## Test plan

TDD: tests added first, confirmed RED on master, then fixes applied and confirmed GREEN.

- [x] `TesseractURDFUnit.parse_octree` — extended existing `prune=\"true\"` blocks with `EXPECT_TRUE(geom->getPruned())`.
- [x] `TesseractGeometryUnit.OctreeClonePreservesPrunedFlag` — constructs with `pruned=true`, clones, asserts `getPruned()`.
- [x] `TesseractGeometrySerializeUnit.OctreeCloneBinaryFlag` — `binary_octree_` has no public getter and is not in `operator==`, so serialize original and clone to cereal-XML and assert blob equality. Format diverges if `binary_octree_` is dropped (different NVP value + different OcTree payload bytes from `writeBinary` vs `write`).
- [x] All 52 `Octree|octree|octomap|Geometry` tests pass.

No test added for the `pruneNode` lifetime fix: the symptom is `assert(children == NULL)`, which is compiled out under `NDEBUG` (the default RelWithDebInfo build), and the leaked children array can otherwise only be detected with valgrind/ASan — too heavy for the standard suite.

### DIFF
--- a/dependencies.repos
+++ b/dependencies.repos
@@ -11,10 +11,6 @@
     uri: https://github.com/Jmeyer1292/opw_kinematics.git
     version: 0.5.3
 - git:
-    local-name: octomap
-    uri: https://github.com/OctoMap/octomap.git
-    version: v1.9.8
-- git:
     local-name: cereal
     uri: https://github.com/Levi-Armstrong/cereal.git
     version: catkin
--- a/dependencies_focal.repos
+++ b/dependencies_focal.repos
@@ -15,10 +15,6 @@
     uri: https://github.com/flexible-collision-library/fcl.git
     version: 0.6.1
 - git:
-    local-name: octomap
-    uri: https://github.com/OctoMap/octomap.git
-    version: v1.9.8
-- git:
     local-name: cereal
     uri: https://github.com/Levi-Armstrong/cereal.git
     version: catkin
--- a/geometry/src/geometries/octree.cpp
+++ b/geometry/src/geometries/octree.cpp
@@ -45,7 +45,7 @@ OctreeSubType Octree::getSubType() const { return sub_type_; }
 
 bool Octree::getPruned() const { return pruned_; }
 
-Geometry::Ptr Octree::clone() const { return std::make_shared<Octree>(octree_, sub_type_); }
+Geometry::Ptr Octree::clone() const { return std::make_shared<Octree>(octree_, sub_type_, pruned_, binary_octree_); }
 
 long Octree::calcNumSubShapes() const
 {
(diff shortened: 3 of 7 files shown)


Levi-Armstrong (approved):

Levi-Armstrong:
With this change can we remove it from the rosinstall file?

Levi-Armstrong:
@rjoomen Should this PR also remove the octomap form the rosinstall file?

rjoomen:
I got this test failure before:

```
 13/17 Test #598: TesseractURDFUnit.parse_octree ....................................................Subprocess aborted***Exception:   0.24 sec
     Note: Google Test filter = TesseractURDFUnit.parse_octree
     [==========] Running 1 test from 1 test suite.
     [----------] Global test environment set-up.
     [----------] 1 test from TesseractURDFUnit
     [ RUN      ] TesseractURDFUnit.parse_octree
     Reading binary octree type OcTree
     Reading binary octree type OcTree
     Reading binary octree type OcTree
     tesseract_urdf_unit: ./octomap/include/octomap/OcTreeDataNode.hxx:69: octomap::OcTreeDataNode<T>::~OcTreeDataNode() [with T = float]: Assertion `children == NULL' failed.
```

With octomap 1.9.7. Afterwards, the failure is gone.

I checked this discussion and the issue referenced therein and I believe this is indeed exactly what this PR fixes. I'll update the various `dependencies.repos` files.

PR #1293, opened 2026-05-01
Fix kdl_sub_tree_builder crash on non-empty floating joints

rjoomen:
## Summary

`kdl_sub_tree_builder`'s constructor copied caller-supplied floating-joint transforms into a freshly-default-constructed `KDLTreeData::floating_joint_values` using `.at(key)`, which requires the key to already exist. The destination map is always empty at that point, so any non-empty `floating_joint_values` argument threw `std::out_of_range`. Swap to `operator[]` (insert-or-assign), matching the sibling full-tree builder a few lines up.

Production impact: `JointGroup` construction threw on any scene with a floating joint. No existing test covered the path — the two existing `parseSceneGraph` sub-tree calls in `tesseract_scene_graph_unit.cpp` pass an empty `TransformMap()`, so the constructor's loop body never ran.

## Test plan

Added `KDLParserSubTreeFloatingJointUnit` (TDD-verified):

- [x] Pre-fix: test fails with `unordered_map::at` thrown from `parseSceneGraph`.
- [x] Post-fix: test passes — `data.floating_joint_values["joint_3"]` matches the supplied transform and the sub-tree contains the expected joint count.
- [x] Full `tesseract_scene_graph_unit` suite (56 tests): all pass.

### DIFF
--- a/scene_graph/src/kdl_parser.cpp
+++ b/scene_graph/src/kdl_parser.cpp
@@ -306,7 +306,7 @@ struct kdl_sub_tree_builder : public boost::dfs_visitor<>
     : data_(data), joint_names_(joint_names), joint_values_(joint_values), floating_joint_values_(floating_joint_values)
   {
     for (const auto& floating_joint_value : floating_joint_values)
-      data_.floating_joint_values.at(floating_joint_value.first) = floating_joint_value.second;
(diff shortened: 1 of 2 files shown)


rjoomen:
CI run requires #1296

Levi-Armstrong (approved):

PR #560, opened 2021-03-03
Add ability to construct ROP and REP kinematic solver with different solver names

Levi-Armstrong:
The ROP and REP rely on sampling of the positioner axes so it it nice to have the ability to create to different solvers for the sample manipulator with different sampling resolutions. This enables the ability to construct the ROP and REP kinematics with a provided solver name. Also updates the SRDF to support providing a solver name which is optional.

Edit:
Also include a bug fix Environment applyCommands not correctly locking the mutex.

PR #1299, opened 2026-05-06
Add optional tolerances on calcJacobianTransformErrorDiff

rjoomen:
## Summary

  - Add optional `lower_tolerance` / `upper_tolerance` args (defaulted empty) to both `tesseract::common::calcJacobianTransformErrorDiff` overloads. When supplied, `err` and `perturbed_err` are clamped to a per-component dead-band
  before subtraction, keeping the FD Jacobian consistent with toleranced error functions.
  - Expose `tesseract::common::applyTolerances` as a public in-place dead-band helper.
  - Empty defaults preserve raw FD behavior — existing callers are unaffected.

  ## Context

  `calcJacobianTransformErrorDiff` is used to build the analytical Jacobian of toleranced Cartesian-pose error functions. Pre-fix, the FD Jacobian subtracted `err` and `perturbed_err` raw — without applying the tolerance clamp the error
   function itself uses — so whenever both fell inside the band the row collapsed to zero, and SQP solvers absorbed it as wasted band slack.

  The clamp is applied after the paired ±π / sign correction so both errors see the same angle-axis branch. When one error is inside and one outside the band, the per-component clamp preserves a non-zero sub-gradient — the descent
  direction a solver needs to cross the band edge.

  ## Test plan

  The new `TesseractCommonUnit.calcJacobianTransformErrorDiff_Toleranced` test covers in-band (diff row collapses to zero), out-of-band (constant offset cancels — result equals raw FD), band-edge (non-zero sub-gradient row), and
  size-mismatch (throws). The existing `calcJacobianTransformErrorDiff` test continues to verify the no-tolerance path. Full `tesseract_common_unit` suite passes locally.

### DIFF
--- a/common/src/utils.cpp
+++ b/common/src/utils.cpp
@@ -175,6 +175,13 @@ void applyTolerances(Eigen::Ref<Eigen::VectorXd> v,
 
 Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
                                                const Eigen::Isometry3d& source,
+                                               const Eigen::Isometry3d& source_perturbed)
+{
+  return calcJacobianTransformErrorDiff(target, source, source_perturbed, Eigen::VectorXd{}, Eigen::VectorXd{});
+}
+
+Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
+                                               const Eigen::Isometry3d& source,
                                                const Eigen::Isometry3d& source_perturbed,
                                                const Eigen::VectorXd& lower_tolerance,
                                                const Eigen::VectorXd& upper_tolerance)
@@ -211,6 +218,15 @@ Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
   applyTolerances(perturbed_err, lower_tolerance, upper_tolerance);
 
   return (perturbed_err - err);
+}
+
+Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
+                                               const Eigen::Isometry3d& target_perturbed,
+                                               const Eigen::Isometry3d& source,
+                                               const Eigen::Isometry3d& source_perturbed)
+{
+  return calcJacobianTransformErrorDiff(
+      target, target_perturbed, source, source_perturbed, Eigen::VectorXd{}, Eigen::VectorXd{});
 }
 
 Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
--- a/common/src/utils.cpp
+++ b/common/src/utils.cpp
@@ -159,9 +159,23 @@ Eigen::VectorXd calcTransformError(const Eigen::Isometry3d& t1, const Eigen::Iso
   return concat(pose_err.translation(), calcRotationalError(pose_err.rotation()));
 }
 
+void applyTolerances(Eigen::VectorXd& v, const Eigen::VectorXd& lower_tolerance, const Eigen::VectorXd& upper_tolerance)
+{
+  if (lower_tolerance.size() == 0 && upper_tolerance.size() == 0)
+    return;
+  if (lower_tolerance.size() != upper_tolerance.size())
+    throw std::runtime_error("applyTolerances: lower_tolerance and upper_tolerance must have the same size");
+  if (lower_tolerance.size() != v.size())
+    throw std::runtime_error("applyTolerances: tolerance size does not match value size (expected " +
+                             std::to_string(v.size()) + ", got " + std::to_string(lower_tolerance.size()) + ")");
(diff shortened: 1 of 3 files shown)


common/src/utils.cpp
```diff
@@ -159,9 +159,32 @@ Eigen::VectorXd calcTransformError(const Eigen::Isometry3d& t1, const Eigen::Iso
...
+                     const Eigen::Ref<const Eigen::VectorXd>& lower_tolerance,
+                     const Eigen::Ref<const Eigen::VectorXd>& upper_tolerance)
+{
+  if (lower_tolerance.size() == 0 && upper_tolerance.size() == 0)
+    return;
+  if (lower_tolerance.size() != upper_tolerance.size())
+    throw std::runtime_error("applyTolerances: lower_tolerance and upper_tolerance must have the same size");
+  if (lower_tolerance.size() != v.size())
+    throw std::runtime_error("applyTolerances: tolerance size does not match value size (expected " +
+                             std::to_string(v.size()) + ", got " + std::to_string(lower_tolerance.size()) + ")");
```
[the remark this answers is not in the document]
rjoomen: Agreed the inverted-band case is silent and worth flagging — but the validation belongs in the constructors of the calculators that own the tolerance vectors, not in `applyTolerances`. The tolerance vectors are invariant per calculator, while `applyTolerances` is on the FD-Jacobian hot path (called twice per perturbed DOF inside `calcJacobianTransformErrorDiff`). Validating there would re-check the same constant inputs O(iterations × DOF × timesteps) times per solve.

Trajopt's `CartPoseErrCalculator`, `CartPoseJacCalculator`, `DynamicCartPoseErrCalculator`, and `DynamicCartPoseJacCalculator` already validate `lower.size() == upper.size()` once at construction; the per-component `lower(i) <= upper(i)` check fits there. Added in trajopt PR #560 commit 974a6cc7 alongside an inverted-band regression test that asserts each ctor throws.

common/src/utils.cpp
```diff
@@ -213,13 +255,7 @@ Eigen::VectorXd calcJacobianTransformErrorDiff(const Eigen::Isometry3d& target,
     perturbed_pose_rotation_err.second *= -1;
   }
 
-  // They should always pointing in the same direction
-#ifndef NDEBUG
-  if (std::abs(pose_rotation_err.second) > 0.01 && perturbed_pose_rotation_err.first.dot(pose_rotation_err.first) < 0)
-    throw std::runtime_error("calcJacobianTransformErrorDiff, angle axes are pointing in oposite directions!");
-#endif
```
Levi-Armstrong, line 220: Do we not want to keep this?

Levi-Armstrong (approved):

common/src/utils.cpp
rjoomen: The block sits after the immediately-preceding sign-flip correction, which already negates perturbed_pose_rotation_err.first/second whenever perturbed.first.dot(pose.first) < 0. After that:

- pre-flip dot ≥ 0 → no flip → post-flip dot ≥ 0
- pre-flip dot < 0 → flip → post-flip dot = -original > 0

So post-flip dot < 0 can't be true on real data, and NaN < 0 is false, so NaN inputs don't fire it either: the throw was unreachable. The > 0.01 guard was there to skip the assertion near zero rotation (where the axis is ill-defined), but with the conjunction unreachable it's irrelevant.

Coverage of the flip itself isn't lost: the existing dynamic-target sweep plus the new calcJacobianTransformErrorDiff_AxisFlipPath test check that path.

But I'm fine with restoring if you'd like to keep an additional guard, please let me know.

jf---:
hi @rjoomen this PR looks like hardcore engineering and of considerable impact.
from a glance of the code, its hard to gauge the precise impact of the PR. 
the test gives some insight, but would be share a few words of wisdom regarding the impact of the PR?
again, this has all the aura of some stellar engineering!

rjoomen:
> hi @rjoomen this PR looks like hardcore engineering and of considerable impact.
> 
> from a glance of the code, its hard to gauge the precise impact of the PR. 
> 
> the test gives some insight, but would be share a few words of wisdom regarding the impact of the PR?
> 
> again, this has all the aura of some stellar engineering!

For details, see the Trajopt PR that fixes an issue with toleranced waypoints: https://github.com/tesseract-robotics/trajopt/pull/560, for which this is required.

And @marrts wrote the original PR introducing toleranced waypoints.

So basically, I only fixed what was already there (or rather, I guided Claude to fix it).

jf---:
that's incredible helpful @rjoomen thanks a lot, I hope I'm not being overly dramatic suggesting that landing this PR seems a perfectly reasonable place to tag 0.35? 

Keeping the delta's between releases reasonable would allow us on the tesseract-nanobind end to keep things in sync with the upstream releases.
this PR is a Big Deal in my book so perhaps I can whisper some words of encouragement  in @Levi-Armstrong  ear ;)

Congratulations again, I'm super stoked about this feature...

PR #1305, opened 2026-06-01
Restore Cereal template registration in compiled cpp files instead of…

rjoomen:
… via _impl headers for non-Windows builds.

PR #1273 caused a massive regression in build memory usage and build times. Reason was the heavy Cereal template instantiation whenever these headers were included. This PR reverts the template registration to cpp files for non-Windows builds.

This also adds [CEREAL_FORCE_DYNAMIC_INIT](https://uscilab.github.io/cereal/assets/doxygen/polymorphic_8hpp.html#a8e0d5df9830c0ed7c60451cf2f873ff5) in the header (non-Windows) to match CEREAL_REGISTER_DYNAMIC_INIT in the cpp (fixes conda-mac CI).

Windows builds are not affected, the fix of the mentioned PR is still in place for that platform.

### DIFF
--- a/.github/workflows/conda/recipe/build.sh
+++ b/.github/workflows/conda/recipe/build.sh
@@ -2,6 +2,13 @@ set -e
 
 # ln -s $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gcc $BUILD_PREFIX/bin/gcc
 
+# conda's clang adds -fvisibility-inlines-hidden, hiding Cereal's visibility("default")
+# registration singleton; with Mach-O's two-level namespace each dylib gets its own registry
+# and consumers throw "unregistered polymorphic type". Strip it so the registry is shared.
+if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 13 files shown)


Levi-Armstrong:
@rjoomen Should the conda-mac CI be passing?

johnwason:
conda on mac uses its own version of clang instead of the system compiler. I see that the cereal tests are having the same polymorphic error we saw on Windows.

rjoomen:
> @rjoomen Should the conda-mac CI be passing?

Yes, looking into it

johnwason:
I think the problem might be symbol visibility. I had to add this to the build.sh script for the Python build:

```bash
if [[ "$(uname)" == "Darwin" ]]; then
  # 1. Prevent hard-linking to a specific libpython binary
  # 2. Allow missing symbols to resolve dynamically when Python imports it
  export LDFLAGS="${LDFLAGS} -Wl,-undefined,dynamic_lookup -Wl,-flat_namespace"
fi
```

This does not work when building for colcon on osx.

rjoomen:
Fixed

Levi-Armstrong (approved):

PR #1316, opened 2026-06-15
feat(geometries): add support for Signed Distance Fields

Joelkang:
- Use case: from a dense pointcloud, im trying to plan a weld around a circular arc. 
- There are pros and cons for speed and trajopt convergability for all of superfine octrees, VHACD and a distance-grid `Mesh`, but claude seems to believe SDF gets you both
- Includes utils to build an SDF by sampling from a tricubic grid (e.g. nvblox). 
- `SDFMesh` doesn't seem to actually be SDF related, but lmk if i should merge/replace


jf---:
Thrilling idea!
FWIW, curious if you've come across [`pytorch_volumetric`](https://github.com/UM-ARM-Lab/pytorch_volumetric)?

Levi-Armstrong:
When I first started writing the collision checker the SDFMesh was a place holder for what was supported by Bullets SDF but I never had the time to revisit. I would propose replacing it with what makes sense to support this functionality in most of the supported collision checkers.

Joelkang:
Oooh thanks @jf--- . @Levi-Armstrong, with reference to pytorch_volumetric's `MeshSDF` as prior art, i kept our existing `SDFMesh` and made it backed by the SDF, which should keep it backwards compatible.

My understanding via discussions with Claude, is also that once the nanobind bindings land, I should be able to use pytorch_volumetric (or nvblox) to create an SDF from my pointcloud on the GPU and pass it to tesseract for Super Fast™ planning

jf---:
> to create an SDF from my pointcloud on the GPU and pass it to tesseract for Super Fast™ planning

Exciting work @Joelkang -- fun follow your PR along.
And true: tesseract + PyTorch is highly potent cocktail: the paper on [configuration space distance field](https://sites.google.com/view/cdfmp/home) strikes me as such an example. It merges `SDF` + `joint spaces` which makes a ton of sense: fusing cartesian space with manipubility strikes as a brilliant intuition (working in voxel spaces makes stacking / layering concerns so straight fwd) . follow the gradient the geodesic, solve your motion-plan. It's a highly intuitive perspective, super elegant, and probably GPU friendly.

just amazing how the SOTA is [made publicly available](https://github.com/yimingli1998/cdf/blob/main/frankaemika/nn_cdf.py)

Joelkang:
@jf--- ooh this is very interesting -- CDFs alone are very compelling, and it does seem we're at the right point in time to  exploit their neural counterparts.

@Levi-Armstrong does this PR align with how you were thinking tesseract should support SDFs, or are there any changes you'd make?

Levi-Armstrong:
@Joelkang Thank you for the PR. I do believe it does align and I plan on reviewing this soon.

geometry/include/tesseract/geometry/impl/sdf_mesh.h
```diff
@@ -31,9 +31,27 @@ TESSERACT_COMMON_IGNORE_WARNINGS_PUSH
...
+ * broadphase, and by backends that cannot consume a distance field (FCL and the Bullet
+ * continuous/cast managers). If a @ref SignedDistanceField is attached via
+ * @ref setSignedDistanceField, the Bullet discrete backend uses it instead of the triangle soup,
+ * giving true volumetric concave contact and signed-distance/gradient queries.
+ *
+ * @note The attached field is concave, so it is only consumed by the Bullet discrete backend. With
+ * no field attached the geometry behaves as a plain triangle mesh on every backend (this is also
+ * the case for meshes loaded from a @c tesseract:sdf_mesh URDF element, which carry no field).
+ */
 class SDFMesh : public PolygonMesh
```
Levi-Armstrong: This looks like it just consumes the new geometry type SignedDistanceField geometry type. Do you still think this geometry type is needed?

geometry/src/geometries/sdf_mesh.cpp
```diff
@@ -72,15 +72,25 @@ SDFMesh::SDFMesh(std::shared_ptr<const tesseract::common::VectorVector3d> vertic
...
+
+std::shared_ptr<const SignedDistanceField> SDFMesh::getSignedDistanceField() const { return sdf_; }
+
 Geometry::Ptr SDFMesh::clone() const
 {
-  return std::make_shared<SDFMesh>(getVertices(), getFaces(), getFaceCount(), getResource(), getScale());
+  auto cloned = std::make_shared<SDFMesh>(getVertices(), getFaces(), getFaceCount(), getResource(), getScale());
+  if (sdf_ != nullptr)
+    cloned->setSignedDistanceField(std::static_pointer_cast<SignedDistanceField>(sdf_->clone()));
+  return cloned;
```
Levi-Armstrong: Are there any tools for creating the Signed Distance Field from a mesh?

geometry/include/tesseract/geometry/impl/signed_distance_field.h
```diff
@@ -0,0 +1,116 @@
...
+  /** @brief Get the collision margin */
+  double getMargin() const;
+
+  Geometry::Ptr clone() const override final;
+
+  bool operator==(const SignedDistanceField& rhs) const;
+  bool operator!=(const SignedDistanceField& rhs) const;
+
+private:
+  std::vector<std::uint8_t> sdf_data_;
```
Levi-Armstrong: I believe you mentioned this is what Bullet expects; how does this related to other SDF representation like VDB? I just want to see if there is anything we need to change to provide broader support for other SDF libraries like VDB.

Levi-Armstrong (commented):
@Joelkang Thank you for the contribution and I am excited to get this merged. I had a few questions and comments.

1. How is this leverage from a URDF; because I did not see any changes in the urdf section for parsing in an SDF?
2. What tools can be used to generate this SDF?

geometry/include/tesseract/geometry/impl/sdf_mesh.h
Joelkang: Indeed it does -- since FCL doesn't support a SDFs, and we already had this construction, I wasn't sure if we wanted to get rid of it completely (are people using it?) If indeed there's such a use case, this at least lets users tie the Mesh to its source/generated SDF.

I can't really think of a use case though, happy to remove completely if this breaking change is acceptable

geometry/src/geometries/sdf_mesh.cpp
Joelkang: As @jf--- suggested above, im actually using https://github.com/UM-ARM-Lab/pytorch_volumetric for doing that. But *that* SDF is not used for motion planning.

geometry/include/tesseract/geometry/impl/signed_distance_field.h
Joelkang: I think this is actually a larger architectural choice in Tesseract -- in the same way as we have our own `Pointcloud` geometry class (vs using Open3D's `Pointcloud`), is it perhaps the consumer's responsibility for converting the SDF data from the library of their choice into the Tesseract SDF?

~As mentioned above, `pytorch_volumetric` lets you generate an analytical SDF from a Mesh, but for motion planning (and other purposes), im actually using https://github.com/openvdb/fvdb-core to create a TSDF/ESDF from our depth-maps. I figured that it should then be my job to, I suppose, copy over the data into a Tesseract SDF object for collision purposes specifically.~

ETA: Actually let me dig into this now, i think i understand better what you are asking

Joelkang:
@Levi-Armstrong updated as per your comments:
1. SDF is now natively backed by a `SignedDistanceFunction sampler`, and bullet builds its own `btMiniSDF`
1. Added URDF loading support, including a way to discretize the grid for serialization

geometry/include/tesseract/geometry/impl/sdf_mesh.h
Levi-Armstrong: I am good with removing it if you are.

Joelkang:
SDFMesh removal rendered

jf---:
@Joelkang that's quite an push the other day, could you describe it / summarise it, the diff is pretty massive?

Joelkang:
Sure thing, the main difference is how the SDF geometry object stores the underlying data. Previously it was storing the data as a vector specifically for btMiniSdf, which unnecessarily couples the pure geom class to the contact manager. 

Now it stores it in a more neutral format, so any SDF producer and consumer can use it without having to do bullet-specific transformations. That all now happens as a bullet util, so the separation of concerns is clearer. 

As per Levis suggestion, I've also added support for loading SDFs from a URDF, which necessitates descretising the field, and helpers for de/serialising the field. 

Other than that, also added a few more tests. 

The 2nd commit then completely removes the SDFMesh from the code based. Happy to split this into a separate PR if that makes things easier.

jf---:
thanks for providing some context, that's very insightful

geometry/include/tesseract/geometry/cereal_serialization.h
```diff
@@ -121,6 +121,23 @@ void serialize(Archive& ar, Octree& obj)
...
+  ar(cereal::base_class<Geometry>(&obj));
+  ar(cereal::make_nvp("domain_min", obj.domain_.min()));
+  ar(cereal::make_nvp("domain_max", obj.domain_.max()));
+  ar(cereal::make_nvp("dim_x", obj.dimensions_.x()));
+  ar(cereal::make_nvp("dim_y", obj.dimensions_.y()));
+  ar(cereal::make_nvp("dim_z", obj.dimensions_.z()));
+  ar(cereal::make_nvp("distances", obj.distances_));
+  ar(cereal::make_nvp("scale", obj.scale_));
+  ar(cereal::make_nvp("margin", obj.margin_));
+}
```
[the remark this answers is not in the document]
Joelkang: added a validation method, which deserialisation will call

geometry/include/tesseract/geometry/geometry.h
```diff
@@ -48,15 +48,24 @@ enum class GeometryType : std::uint8_t
...
+                                                              "CAPSULE",
+                                                              "CONE",
+                                                              "BOX",
+                                                              "PLANE",
+                                                              "MESH",
+                                                              "CONVEX_MESH",
+                                                              "OCTREE",
+                                                              "POLYGON_MESH",
+                                                              "COMPOUND_MESH",
+                                                              "SIGNED_DISTANCE_FIELD" };
```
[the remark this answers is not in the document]
Joelkang: updated the SDFMesh removal commit so that we use the same enum index and simply replace "SDFMesh" -> "SIGNED_DISTANCE_FIELD"

Joelkang:
Just pushed a fix for the failing positive distance test: Stock Bullet routes convex-vs-SDF pairs to `btConvexConcaveCollisionAlgorithm`, whose [SDF branch](https://github.com/bulletphysics/bullet3/blob/63c4d67e337017f9d8b298c900e9aabdb69296e7/src/BulletCollision/CollisionDispatch/btConvexConcaveCollisionAlgorithm.cpp#L231C1-L290C4) only emits a contact when the query point is within the probe's surface (`dist <= radius`).

It ignores `btManifoldResult::m_closestPointDistanceThreshold`, so contacts at positive distance inside the collision margin band are never generated. Positive-distance contacts are the load-bearing case for motion planning: 
- TrajOpt's collision costs consume contacts reported at `0 < distance < margin`
to obtain a gradient that pushes links apart BEFORE they collide
- OMPL-style validity gates admit states by the same margin band. 

A backend that only reports penetration still passes every boolean collision check while silently breaking all margin-based consumers since they are the only tests that distinguish "collision detection" from the distance-to-collision contract tesseract requires.

This fix adds `TesseractConvexSdfAlgorithm`, replicating Bullet's SDF vertex query but widening the acceptance band by the pair's contact distance threshold, registered for convex-vs-SDF pairs via `TesseractCollisionConfiguration`.

Additions along the way:
- Capsule probes are now supported (absent in stock Bullet) by sampling the field along the axis segment with a radius offset; samples are spaced at most one radius apart.
- Unsupported probes (cylinder/cone) now warn once per pair instead of silently missing every contact.

Levi-Armstrong:
@Joelkang Please ping me when this is ready for another look.

Joelkang:
@Levi-Armstrong yea this is pretty stable now to have another look -- sorry for the PR growing as big as it did. All the failing checks seem to be dependency related things that were not changed in this PR.

Levi-Armstrong:
@Joelkang Do you have a tesseract_qt and tesseract_ros2 branch updated that works with these changes?

Joelkang:
@Levi-Armstrong created PRs in these repos

- [X] https://github.com/tesseract-robotics/tesseract_ros2/pull/196 - Tests round tripping messages pass
- [X] https://github.com/tesseract-robotics/tesseract_qt/pull/214 - Compiles neatly
- [X] https://github.com/tesseract-robotics/tesseract_ros/pull/287 - Round trip test succeeds

Im currently building a container to test,  will update them accordinglty

Joelkang:
@Levi-Armstrong i see now that my mistake is creating a custom SDF serialisation format, whereas we should just support `.vdb` and/or `.nvdb`. That obviously comes with additions to deps that probably requires a discussion.

Joelkang:
Superseded by #1322

PR #1319, opened 2026-07-08
Optimize package dependencies

Rick-v-E:
Optimize dependencies:

* Mark header-only libraries as `build_depend` (and `build_export_depend` when they are public).
* Use `-dev` dependencies for `build` and the normal onces for `exec` when they are available through `rosdep`
* Only define used `pcl` components for `exec` to avoid installing complete `pcl` when not needed

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -39,8 +39,7 @@
   <depend>yaml-cpp</depend>
   <depend>boost_plugin_loader</depend>
   <depend>liboctomap-dev</depend>
-  <build_depend>cereal</build_depend>
-  <build_export_depend>cereal</build_export_depend>
+  <depend>cereal</depend>
   <build_depend>assimp-dev</build_depend>
   <build_export_depend>assimp-dev</build_export_depend>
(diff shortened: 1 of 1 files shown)


Levi-Armstrong:
@johnwason It looks like conda CI is failing for an unrelated reason.

Levi-Armstrong (approved):

Rick-v-E:
The [`libcereal-dev`](https://index.ros.org/d/libcereal-dev/) rosdep key is also available. Since cereal is header only, we could also do:

```xml
<build_depend>libcereal-dev</build_depend>
<build_export_depend>libcereal-dev</build_export_depend>
```

instead of 

```xml
<depend>cereal</depend>
```

Then rosdep can install the cereal dependency and we can remove it from the `dependencies.repos`. The only downside is that the rosdep key is not specified for Mac-OS, so we need to do something custom for MacOS in the CI.

johnwason:
@Levi-Armstrong is the conda build back to normal? I don't see the issue.

PR #1324, opened 2026-07-15
Fix broken cereal install on jammy

Rick-v-E:
Fix broken cereal install on focal and jammy, `libcereal-dev` is broken on these distributions. With this change, we use the original `cereal` package on these distribtions while we rely on rosdep on noble.

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -34,7 +34,8 @@ jobs:
         uses: tesseract-robotics/colcon-action@v14
         with:
           ccache-prefix: benchmarks
-          vcs-file: dependencies.repos
+          vcs-file: dependencies_jammy.repos
+          rosdep-install-args: -iry --skip-keys=libcereal-dev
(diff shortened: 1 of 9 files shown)


rjoomen:
This would also apply to mac then, right? Should dependencies-jammy become something like dependencies-cereal?

Rick-v-E:
> This would also apply to mac then, right? Should dependencies-jammy become something like dependencies-cereal?

Mac is not affected by this, since it has its own dependencies [specified](https://github.com/tesseract-robotics/tesseract/blob/a9f750001bd38a943ddbf5dd5023893755f23342/.github/workflows/mac.yml#L34). 

What I also did notice is that in the `.repos` files the `ros_industrial_cmake_boilerplate` is also cloned and build. This is not needed since the same version is also available as rosdep (for both jammy and noble). Should I remove this dependency from the repos files as well?

Levi-Armstrong:
Let leave the ros_industrial_cmake_boilerplate in repos for now.

Issue #1323, opened 2026-07-13
PCL's exported `-march=native` splits the Eigen alignment ABI across tesseract libraries (heap corruption in `tesseract_urdf` consumers)


munich35:
## Summary

On Ubuntu 24.04 (and any distro whose packaged PCL exports architecture flags), the imported PCL targets `pcl_common`, `pcl_io`, `pcl_io_ply`, `pcl_octree` carry:

```
INTERFACE_COMPILE_OPTIONS = $<$<COMPILE_LANGUAGE:CXX>:-msse4.2;-mfpmath=sse;-march=native;-mavx2>
```

`tesseract_urdf` is the only tesseract library that links PCL (for the point-cloud URDF elements), so it alone inherits `-mavx2`. That raises Eigen's max alignment to 32 bytes in that one set of translation units, while every other tesseract library is built with 16-byte alignment. The result is that `sizeof(tesseract_scene_graph::Collision)` differs across libraries — we measured **192 bytes vs 176 bytes** for the identical type depending on which library's flags the TU was compiled with.

Objects constructed by the URDF parser therefore have a different layout than the code that later destroys them (e.g. `tesseract_rosutils` or downstream consumers), producing cross-library free/heap corruption.

Observed symptom at `-O2` (Release): `munmap_chunk(): invalid pointer` crashing the environment monitor within seconds of startup. At `-O0` the same binaries happen to survive on symbol-binding luck (weak-template interposition picks a single instantiation) — i.e. **the bug is optimization-independent and latent**, and is a plausible explanation for otherwise-sporadic SIGSEGVs in downstream applications. We hit this in a 6-DOF industrial arm application; the failure is fully reproducible.

This is a memory-corruption class issue, so we consider it critical.

## Environment / versions

- tesseract 0.34.1 (`e3ee23cd`)
- tesseract_planning 0.34.0 (`d7dfa425ca`), tesseract_ros2 0.34.0 (as consumers; the root cause is in tesseract's `tesseract_urdf` build)
- Ubuntu 24.04, ROS 2 Jazzy, GCC 13
- System (distro-packaged) PCL

## Reproduction

1. Build tesseract on Ubuntu 24.04 against the system PCL, Release (`-O2`).
2. Run any consumer that parses a URDF in one library and destroys the environment in another. For example, tesseract_ros2's `tesseract_monitoring_environment_node` crashes within seconds of startup with `munmap_chunk(): invalid pointer`.
[...]

PR #1325, opened 2026-07-15
Prevents PCL-exported ISA options

Levi-Armstrong:


munich35:
Thanks for the quick turnaround, @Levi-Armstrong — this matches exactly what we independently diagnosed and fixed.

For the record, from our investigation (Ubuntu 24.04 / ROS 2 Jazzy): the PCL packaged targets (`pcl_common`, `pcl_io`, `pcl_io_ply`, `pcl_octree`) export `-msse4.2 -mfpmath=sse -march=native -mavx2` via `INTERFACE_COMPILE_OPTIONS`. `tesseract_urdf` is the only PCL consumer (the `<octomap><point_cloud>` path), so it inherits `-march=native` and ends up with **Eigen `MaxAlign` 32 while every other tesseract library uses 16** — we measured `sizeof(...Collision...)` at **192 bytes inside `tesseract_urdf` vs 176 elsewhere**. Objects the URDF parser constructs then have a different layout than the code destroying them → heap corruption, which surfaced for us as ASAN `attempting free on address ...` and intermittent SIGSEGVs (including in the environment-monitor node).

Stripping `INTERFACE_COMPILE_OPTIONS` on those four PCL targets is exactly the right defense and matches our fix — we verified it ASAN-clean and through a full production run.

One purely optional belt-and-braces addition we layered on top: a baseline `target_compile_options(tesseract_urdf PRIVATE -march=x86-64 -mtune=generic -mno-avx2 -mno-avx)` on the `tesseract_urdf` target. Stripping the inherited flags is sufficient on its own; the explicit baseline just pins the ABI for any consumer built without its own arch flags. Your PR resolves the issue as-is.

Happy to test #1325 against our stack if that's useful.

Greetings from Stefan / Munich

munich35:
Re-tested the merged fix on our production stack (Ubuntu 24.04, GP25 packing cell, TrajOptIfopt pipeline) as requested — **verified, and it fully supersedes our local workaround.**

Test setup: to isolate the fix we cherry-picked 19291ea7 onto our production base (e3ee23cd) with our local workaround commits removed, rebuilt `tesseract` (Release, compile-commands exported), and ran our certification replay battery against the rebuilt libs.

Results:
1. **Flag gate:** 0/167 TUs (0/26 in the urdf component) carry `-mavx2` or `-march=native` — the PCL ISA leak from #1323 is gone.
2. **Runtime battery:** 27 TrajOptIfopt planning replays (pinned freespace + DiscreteContactCheck against a 74-link production scene) — 27/27 SUCCESS, 27/27 contact-clean, solve times in the normal band. No heap corruption, no `sizeof(Collision)` divergence symptoms.
3. **Bit-parity:** all 27 solution trajectories are bit-identical to our reference build that used our local workaround (worst per-joint delta 0.00e+00 rad) — the fix is behaviorally equivalent.

Also reviewed the diff: same mechanism we converged on locally (clearing `INTERFACE_COMPILE_OPTIONS` on `pcl_common`/`pcl_io`/`pcl_io_ply`/`pcl_octree`), and the GNU/Clang-x86 guard with the MSVC exemption makes it more portable than our version. We'll drop our local commits in favor of upstream at our next tesseract bump. Thanks for the quick turnaround, @Levi-Armstrong.

PR #1339, opened 2026-08-17
Silence PCL's malloc-convention guard in the point cloud parser

rjoomen:
`pcl/memory.h` `#error`s when Eigen's allocation convention in the consuming translation unit differs from the one PCL was built with. conda-forge's Windows PCL is built with AVX, so it uses Eigen's handmade aligned malloc, while the rest of the workspace uses plain malloc. Until now that was handled by compiling all of conda-win with `/arch:AVX` and building `orocos-kdl` from source to match — see #1325 and the recipe changes that came in with #1274.

That is more than the situation needs. The guard is a blanket include-time check with no way to see which API is called, and the hazard it protects against is not present in `parsePointCloud`:

- the only object handed to `pcl_io` is a `PCLPointCloud2`, whose storage uses `std::allocator`;
- `PCDReader::read` is a header template with no `extern template` declaration, so the `Eigen::aligned_allocator` resize of `PointCloud::points` — and its release — both happen in our translation unit;
- the two Eigen objects passed by reference, `Vector4f` and `Quaternionf`, are 16 bytes and `alignas(16)` under either configuration.

So no Eigen-allocated buffer ever crosses the DLL boundary. This defines `PCL_SILENCE_MALLOC_WARNING` in `point_cloud.cpp` only, with a comment recording the invariant and warning against breaking it, and drops the `/arch:AVX` flag, the source-built `orocos-kdl` and `conda/windows_kdl.repos` from the conda recipe.

The result is a single Eigen configuration everywhere, and downstream repositories that build tesseract from source on conda-win no longer need to replicate any of it.

To be explicit about what is *not* reverted: `urdf/CMakeLists.txt` keeps stripping PCL's exported `INTERFACE_COMPILE_OPTIONS` on every platform, and #1325's MSVC carve-out does not come back. That strip and the `/arch:AVX` flag were added together in one commit and were mutually dependent — the strip was only viable on MSVC because the recipe was supplying the alignment instead. This PR breaks that pairing and re-justifies the strip on a different basis: with the guard silenced where it does not apply, nothing needs PCL's flags any more, so they can be kept away from our code everywhere.

Restoring the carve-out would undo what #1325 set out to fix. PCL is linked `PRIVATE` into `urdf`, so its exported `/arch:AVX` applies to that whole target and nothing else — which puts `Isometry3d` members such as `Joint::parent_to_joint_origin_transform` at `alignas(32)` inside `tesseract_urdf` and `alignas(16)` in `tesseract_scene_graph`, for objects that cross between them during parsing.

The commit also carries an unrelated one-token fix: the conda recipes ignored `osqp_eigen`, which is not the workspace package name — it is `osqp_eigen_ext` (https://github.com/tesseract-robotics/trajopt/pull/575).

Note the guard cannot be exercised on Linux CI: PCL 1.14 predates `PCL_USES_EIGEN_HANDMADE_ALIGNED_MALLOC`. conda-win against PCL 1.15.1 is the real test.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -27,9 +27,6 @@ jobs:
     - name: vcs import
       run: >
         vcs import --input "${{ github.workspace }}/workspace/src/tesseract/.github/workflows/windows_dependencies.repos" workspace/src/
-    - name: vcs import kdl source
-      run: >
-        vcs import --input "${{ github.workspace }}/workspace/src/tesseract/.github/workflows/conda/windows_kdl.repos" workspace/src/
     - uses: conda-incubator/setup-miniconda@v4
       with:
         channel-priority: strict
--- a/.github/workflows/conda/recipe/build.bat
+++ b/.github/workflows/conda/recipe/build.bat
@@ -1,29 +1,7 @@
 
-rem PCL and the other conda dependencies are built with AVX, i.e. 32-byte Eigen alignment, and their headers
-rem require consumers to match (pcl/memory.h #errors otherwise). Build all of Tesseract with AVX so Eigen's static
-rem and malloc alignment are 32 in every translation unit, giving one uniform layout/ABI across the whole workspace.
-set "CXXFLAGS=%CXXFLAGS% /arch:AVX"
-
-rem conda-forge's Windows orocos-kdl is a static library compiled without AVX, so the Eigen
-rem allocation/free convention inlined into its objects (16-byte, plain malloc) is incompatible with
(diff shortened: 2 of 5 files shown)


Levi-Armstrong (approved):

rjoomen:
If this works and is merged and the container published, I'll strip the two PCL-related commits from https://github.com/tesseract-robotics/trajopt/pull/554, as conda-win CI should not need those anymore.

rjoomen:
@johnwason FYI

PR #1344, opened 2026-08-17
Guard index access in TrajectoryPlayer and TrajectoryInterpolator

rjoomen:
Two missing guards on the same index path.

**`TrajectoryInterpolator::getStateDuration` clamps from above but not from below.** `index >= s` is folded to `s - 1`, but a negative index falls straight through to `trajectory_[static_cast<std::size_t>(index)]`, so `-1` subscripts the vector at `SIZE_MAX`.

This is silent by default — the read lands in mapped memory and the caller gets a plausible-looking small double. Compiled into a standalone harness it returns `4.8270213598689787e-321` for `getStateDuration(-1)` where `0` is correct. That is also why the obvious regression assertion (`EXPECT_NEAR(getStateDuration(-1), 0, 1e-5)`) is worthless: a denormal is within `1e-5` of zero, so it passes against the unfixed code. The assertion added here instead uses a trajectory whose first state time is `5.0`, which garbage does not match.

**`TrajectoryPlayer::getByIndex` is the only public method that does not check `trajectory_`.** Its four neighbours all `throw std::runtime_error("Trajectory is empty!")`; this one dereferences a null `unique_ptr` when called before `setTrajectory`. It is also the only path that can pass an unfiltered negative index into `getStateDuration` — `setCurrentDurationByIndex` special-cases `index <= 0` and never calls it — so the two defects meet here.

The added `PlayerWithoutTrajectoryTest` segfaults against the unfixed code rather than failing; nothing in the suite constructed a player without a trajectory before.

### DIFF
--- a/visualization/src/trajectory_interpolator.cpp
+++ b/visualization/src/trajectory_interpolator.cpp
@@ -136,6 +136,8 @@ double TrajectoryInterpolator::getStateDuration(long index) const
   int s = static_cast<int>(trajectory_.size());
   if (index >= s)
     index = s - 1;
+  if (index < 0)
+    index = 0;
 
   return trajectory_[static_cast<std::size_t>(index)].time;
 }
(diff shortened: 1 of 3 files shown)


rjoomen:
Folded into #1343 — same class, same file pair, all one-line guards. Closing in favour of that PR.

rjoomen:
Added to #1343

PR #1357, opened 2026-08-26
DO NOT MERGE: test colcon-action@main

rjoomen:
Throwaway draft, opened to exercise the unreleased `colcon-action` against a real consumer before a
`v16` tag is cut. **Not for merge — I will close it once the legs report.**

Switches all seven workflows from `@v15` to `@main`. Four of them run on `pull_request` and are the
actual test: `ubuntu`, `mac`, `windows`, `code_quality`. (`benchmarking` and `package_debian` are
`push`-only, `nightly` needs a dispatch, so those three are along for the ride.)

What is being tested, from tesseract-robotics/colcon-action#32:

- **macOS** is the real change — `Install base deps` now probes each tool and installs only what is
  missing, instead of `brew update` plus a `brew install` that upgraded the runner's cmake, git and
  python3 every run. `mac.yml` is why this repository was chosen: it is the only `@v15` consumer with
  a macOS leg.
- **Linux** applies the same guard via `command -v`, plus one fewer `apt update`.
- **The ROS key fetch** now passes `curl -f`, so a throttled `raw.githubusercontent.com` fails at the
  fetch rather than writing an error page into the keyring.

The interesting failure modes are a tool the guard wrongly considers present, and the `-f` firing on
a transient 429. Windows is untouched by that PR and is here as a regression check.

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -35,7 +35,7 @@ jobs:
           apt install -y liboctomap-dev
 
       - name: Build
-        uses: tesseract-robotics/colcon-action@v15
+        uses: tesseract-robotics/colcon-action@main
         with:
           ccache-prefix: benchmarks
           vcs-file: dependencies_jammy.repos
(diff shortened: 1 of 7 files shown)


rjoomen:
Test complete: all seven jobs green against `colcon-action@main`, including the macOS leg this repository was chosen for. Closing — the run stays as the evidence for cutting a v16 tag. Not for merge.

PR #1363, opened 2026-09-09
Make KDLStateSolver thread-safe and concurrent

rjoomen:
# Make KDLStateSolver thread-safe and concurrent

`KDLStateSolver` has two concurrency defects. Both come from one cause: KDL memoises joint poses
into `mutable` members (`KDL::Joint`'s `joint_pose` and `q_previous`) of objects the solver shares
between threads. Every revolute joint tesseract builds is `KDL::Joint::RotAxis` and every prismatic
one `TransAxis` — exactly the two branches that memoise.

## The jacobian path returns wrong answers under concurrency

`getJacobian` reaches `TreeJntToJacSolver::JntToJac` on a single shared solver, taking no lock at
all. That solver owns a copy of the tree, carrying the same memoised joints, so concurrent callers
overwrite each other's cached pose and receive a jacobian for another thread's configuration.

This is live in shipped multi-threaded `trajopt_sco`, which calls `JointGroup::calcJacobian` per
contact inside its OpenMP regions. Measured against single-threaded reference jacobians, 12 threads:

| | mismatched | worst absolute error |
|---|---|---|
| before | 183 / 60000 | 2.446 |
| after | 0 / 60000 | 0 |

The first commit gives each thread its own solver rather than adding a lock, so the jacobian keeps
running concurrently. Construction copies the tree and costs about 30 us against roughly 1 us per
call, but happens once per thread.

## Forward kinematics was serialised, so it scaled backwards

`calculateTransforms` held an exclusive mutex across the whole recursive traversal, because the
traversal called KDL's memoising `Segment::pose`. `JointGroup` always builds its own
`KDLStateSolver`, and `Environment::getJointGroup(name)` is cached, so every consumer shares one
solver and one lock.

The second commit computes the joint frame directly from geometry captured once at construction —
joint type, axis, origin and tip transform — which is a pure function of the joint value and needs
no lock. The mutex is deleted.

Forward kinematics throughput through a shared `JointGroup`, calls/ms:

| threads | 1 | 2 | 4 | 8 | 12 |
|---|---|---|---|---|---|
| before | 379 | 134 | 163 | 178 | 163 |
| after | 537 | 1014 | 1920 | 3371 | 3052 |

Before, throughput falls as threads are added and never recovers. After, it rises. In the same run,
a shared `JointGroup` now measures in the same band as one private solver per thread, so sharing no
longer costs anything.

## Verification

- Full `tesseract` suite: 903 tests, 0 failures.
- `tesseract_environment_unit.cpp` compares `KDLStateSolver` against the OFKT solver in roughly
  twenty places, and `state_solver_test_suite.h` uses KDL as the reference for OFKT-named tests.
- Downstream: `trajopt` 43/43 and `trajopt_common` 20/20, built from source against this branch.
- Two new tests. The jacobian one was demonstrated failing against unmodified code (1415/4000
  mismatches) before the fix; it skips on a single-core runner. The equivalence test walks the tree
  with KDL's own `Segment::pose` and requires the solver to agree, pinning the replacement against
  any future change in KDL's convention.
- End-to-end, `BM_TRAJOPT_MULTI_THREADED_SIMPLE_COLLISION_SOLVE` over 50 iterations, interleaved:
  84.3 ms before against 77.1 ms after. The arms' ranges overlap, so treat the mean as suggestive;
  the clearer signal is that the spread narrows from 1.21x to 1.03x.

## Trade-offs, stated deliberately

- **Per-thread solvers are never evicted.** A solver used from a thread-per-request pattern retains
  one tree copy per thread that ever called `getJacobian`. Thread pools — what OpenMP and every
  in-tree caller use — reuse a fixed set of threads, so the map stays at pool size. Evicting on
  thread exit needs a thread-exit hook and buys nothing for any caller here; if an unbounded-thread
  consumer appears, that is the change to make.
- **The closed form assumes KDL's default joint scale of 1.** `Joint::pose` evaluates at
  `scale * q + offset`; the captured tip transform absorbs `origin` and `offset` exactly, but nothing
  absorbs `scale`. `kdl_parser` always uses the four-argument `KDL::Joint` constructor, which leaves
  scale at 1, so nothing tesseract builds can reach this. `KDL::Joint` exposes no accessor for scale,
  so it is documented at the capture site rather than asserted.
- **`RotX/RotY/RotZ/TransX/TransY/TransZ` are handled but untested.** They are covered for
  completeness, since the `Segment::pose` call being replaced handled them. `kdl_parser` emits only
  `None`, `RotAxis` and `TransAxis`, so no tree tesseract can construct reaches them, and a test
  would need a hand-built tree. The types that are reachable are covered by the equivalence test.

`KDLFwdKinChain` has the same serialisation and is deliberately untouched: it locks both its
forward kinematics and its jacobian, so it is correct, and it is not on the `JointGroup` path.

The two commits are independently revertible.


Levi-Armstrong:
It may be best to just bit the bullet and switch it over to OKFT version; what do you think?

rjoomen:
> It may be best to just bit the bullet and switch it over to OKFT version; what do you think?

Just what I was discussing with Claude this afternoon, and that seems to be a viable option indeed. Besides a reduction of code and complexity, we'd also get rid of this [thread safety issue](https://github.com/orocos/orocos_kinematics_dynamics/pull/399) in KDL < v1.5.2.

Shall we do that as a follow up? Having a correct KDL as a baseline is a good starting point.

I am doing some more investigation in this actual PR, btw, as I'm not too happy about the quoted regression.

Levi-Armstrong:
> Shall we do that as a follow up? Having a correct KDL as a baseline is a good starting point.

I am good with doing switching to OKFT in a follow-on PR.

rjoomen:
@rjoomen This is ready. The more investigating I mentioned before was actually about https://github.com/tesseract-robotics/trajopt/pull/588, my bad. This PR has no regression and fixes concurrency issues with Jacobians if Trajopt SCO is running multi-threaded.

rjoomen:
@Levi-Armstrong This is ready. The more investigating I mentioned before was actually about https://github.com/tesseract-robotics/trajopt/pull/588, my bad. This PR has no regression and fixes concurrency issues with Jacobians if Trajopt SCO is running multi-threaded.

Levi-Armstrong (approved):

Levi-Armstrong:
@rjoomen How did you track this down?

rjoomen:
> @rjoomen How did you track this down?

I had Claude sweep all of Tesseract and Coal for bugs and other issues. I have a list of over 1000 findings. Part of these are tests that do not really test/verify what they should, etc., but still hundreds of actual issues (again part of these are hypothetical and not currently an issue, but still latent problems). I'm going to gradually get these issues fixed, starting with the ones that are most serious and most likely to affect actual planning.

And my main focus now is to get planning more robust and faster.

